PR #28, opened 2026-04-22
[LOG-4476] feat(install): recognize `skills/` subdirectory as a tap's skill index

stevekrenzel:
## Summary
When a tap's root has neither a `SKILL.md` nor skill-containing subdirectories at the root but DOES have a `skills/` subdirectory, walk that instead. This matches the common layout where skills live under `skills/*` and the root holds READMEs, licenses, and CI config.

## Precedence (PRD §9 step 5)
1. Root `SKILL.md` → one skill (unchanged)
2. Root `skills/` exists → walk one level under `skills/`; root is **not** walked in this case — `skills/` is authoritative
3. Otherwise → walk one level under the root (unchanged)

## Changes
- `PRD.md` §9 step 5 rewritten as a three-case decision tree; new conformance criterion `C-INST-08b`.
- `src/sources/expand.ts` refactored to share the collector between the `skills/` branch and the fallback.
- New e2e test for the `skills/` case.
- New `tests/unit/expand.test.ts` with focused branch coverage (root-wins, skills/-only, skills/-empty error, fallback, skip-non-dir, skip-no-SKILL.md).

## Test plan
- [x] `bun run check` green (746 tests, 100% coverage including `src/sources/expand.ts`).

### DIFF
--- a/.github/pr-review-prompt.md
+++ b/.github/pr-review-prompt.md
@@ -35,6 +35,12 @@ Before reviewing any code, read:
   corresponding `PRD.md` update, or a PRD change not reflected in
   `src/` and `tests/`. CLAUDE.md's opening section defines what
   triggers a PRD update — apply that test.
+- **User-facing surface drift**: a change to user-visible behavior
+  (command, flag, install/tap semantics, repo-layout conventions,
+  defaults) without corresponding updates to `site/` and/or
+  `README.md` where those surfaces describe the thing. See
+  CLAUDE.md's "user-facing surfaces" section — flag prose that is
+  now wrong or misleadingly incomplete.
 - **Coverage gaps**: new code in `src/` without tests, or code
   structured in a way that can't reach 100% coverage (see the
   "coverage quirks" section in CLAUDE.md for the recurring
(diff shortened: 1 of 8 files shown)


stevekrenzel:
Addressed both nits from the review.

**Applied**
- `src/sources/expand.ts`: the `no_skills_found` error message now lists all three accepted shapes (root `SKILL.md`, subdirectories, or a `skills/` directory of either) — commit b0a5e4c.

**Not applied (intentional)**
- The C-INST-09 criterion reads "a directory source that expands to zero valid skills" — which correctly covers every case (1, 2, and 3) via a single assertion. The reviewer flagged this as a not-blocking nit and called out that C-INST-09 is implementation-agnostic. Splitting it into per-case variants would multiply the conformance surface without catching anything new (the three-case dispatch inside `expandSkills` is already exhaustively covered by the unit tests at `tests/unit/expand.test.ts`). Leaving as-is.

On the symlink observation: confirmed pre-existing behavior of `isDirectory`; not in scope for this PR.

jgarms (approved):
LGTM

PR #29, opened 2026-04-23
[LOG-4503] feat(install): one-level namespace grouping under skills/

stevekrenzel:
> **Stacked on #28.** Please merge that first; this PR's diff will automatically rebase to `main` once it does.

## Summary
Teaches `crew install` about namespace grouping inside a tap's `skills/` directory, adds 3-segment refs, and gives the user clean ways to disambiguate when a name could mean more than one thing.

### Repo layout
A **namespace** is a directory directly under `skills/` that has no `SKILL.md` of its own but contains child directories that do.

```
acme-skills/
└── skills/
    ├── marketing/
    │   ├── email-outreach/SKILL.md
    │   └── social-posts/SKILL.md
    └── engineering/
        ├── code-review/SKILL.md
        └── pr-descriptions/SKILL.md
```

### Install forms
| Ref | Behavior |
| --- | --- |
| `crew install marketing` | installs every skill in the `marketing` namespace (when the name is unique) |
| `crew install acme/marketing/email-outreach` | picks one skill — always unambiguous |
| `crew install marketing/email-outreach` | picks one skill (when the namespace exists in exactly one tap) |
| `crew install --tap acme` | forces "install the whole `acme` tap" |
| `crew install --bundle marketing` | forces "install the `marketing` namespace" |
| `crew install --skill pdf` | forces single-skill install |

When a bare name is ambiguous across taps/skills/namespaces, crew prompts on a TTY (numbered menu) and falls back to a rich `ambiguous_reference` error with a copy-pasteable install command for every candidate.

## PRD
- §8.3 rewritten for the broader reference semantics (1-, 2-, 3-segment forms; disambiguation rules; new flags).
- §8.4 grammar extended with the 3-segment form and a whole-tap install form.
- §9 step 5 case 2 gains a sub-rule for namespace descent (one extra level under `skills/`).
- §13 `ambiguous_reference` description broadened to cover the new cases.
- New conformance rows C-NS-01..10.

## Code surfaces
- `src/sources/expand.ts` — descend one more level under `skills/` into namespace dirs.
- `src/install/tap-index.ts` (new) — shallow index of a tap's layout (skills by name + namespaces with members).
- `src/install/resolve-ref/` (new directory, `index.ts` + `format.ts`) — `resolveTapRef` turns a parsed `TapSource` into a concrete `NameCandidate`; `formatCandidate` / `shortLabelFor` render copy-pasteable commands.
- `src/install/attribute-bare-name.ts` — `enumerateCandidates` now returns every interpretation; legacy `findTapForBareName` (used by `info`) delegates to it.
- `src/install/resolve.ts` — `enqueueTapRef` dispatches on resolver output (`skill` / `namespace` / short-circuit tap).
- `src/commands/install/ambiguity-prompt.ts` (new) — interactive menu + non-TTY error formatter for bare-name ambiguity.
- `src/cli/args.ts` — `--tap` / `--bundle` / `--skill` flags on `install`.
- `src/core/types.ts` — `TapSource` gained an optional `namespace` field.
- `src/refs/parsers.ts` — parses 1/2/3-segment refs.

## Tests
- New unit files: `tap-index.test.ts`, `resolve-ref.test.ts`, `attribute-bare-name.test.ts`.
- New e2e file: `namespaces.test.ts` — C-NS-02..09 against real git repos.
- Existing tests updated for the new `TapSource.namespace` shape.
- `bun run check` green: 800 tests, 100% line + function coverage.

## Docs
- `README.md` adds a "Namespaces" subsection with a layout tree and the three install forms.
- `site/components/sections/Taps.tsx` adds a short paragraph naming namespaces and showing the install forms.


stevekrenzel:
Addressed the review. Summary:

**Applied (all priority/important items)**
- **File-size cap**: split `src/install/resolve.ts` (372 lines) into a directory. `resolve/index.ts` is 160 lines of orchestration; `resolve/enqueue.ts` is 239 lines of per-reference helpers. enqueue.ts is modestly over 200 but well under the original; further splitting would break up `enqueueTapRef`'s decision tree and make it harder to read.
- **Stale doc example**: replaced `copy-review` with `email-outreach` in README.md and site/components/sections/Taps.tsx — the example tree only contains `email-outreach` and `social-posts`, so `copy-review` didn't reference anything in the tree.
- **Re-export**: dropped the `export { formatCandidate, shortLabelFor } from "./format.ts"` from `resolve-ref/index.ts`. `ambiguity-prompt.ts` now imports directly from `./format.ts`. The unit test file was already doing this correctly.
- **`--yes` behavior comment**: added a 4-line comment at the `if (ctx.flags.yes) return raw` branch explaining that `--yes` skips the prompt but does NOT pick a default here — skill/namespace ambiguity has no obvious default, so we fall through and let `ambiguous_reference` surface.
- **PRD grammar**: rewrote §8.3 and C-NS-06/07/08 to clarify that `--tap`, `--bundle`, `--skill` are presence flags; the name comes from the positional.

**Not applied (intentional)**
- **`src/core/types.ts` size**: reviewer flagged this as pre-existing, not caused by this PR, and out of scope — agreed. The `namespace` field's doc comment is 6 lines and that file is 258. Splitting the shared types file isn't worth a PR on its own; will handle as a follow-up refactor when there's a natural seam.
- **`safeIndex` broad catch**: reviewer noted this swallows more than unreachable taps, but flagged it as not a showstopper. The comment at the call site says "soft-fail on unreachable taps"; the wider swallow is deliberate since a busted tap clone (disk corruption, race with `tap remove`) shouldn't abort a resolver that's trying to service every other tap. Left as-is.

`bun run check` green: 800 tests, 100% coverage.

jgarms (dismissed):
👍

jgarms (dismissed):

jgarms (approved):

PR #31, opened 2026-04-23
[LOG-4505] feat(install): render already-installed skills as full per-agent blocks

stevekrenzel:
> **Stacked on #30 → #29 → #28.** Merge those in order first.

## Summary
When `crew install <something-already-installed>` runs, the command currently renders a one-line footnote:

```
  · landing-page-copy already installed · in agent-skills, claude-code, codex, cursor, factory, gemini-cli
    <description>
```

In a multi-skill install where some are new and some are already-installed, this cramped format looks like a footnote compared to the rich per-agent grid used for fresh installs. You have to eyeball two different shapes to see the same kind of information.

This PR renders already-installed skills in the **same block shape** as newly-installed ones:

```
  landing-page-copy  (already installed)
    A rigorous content marketer for landing pages…
    · agent-skills → ~/.agents/skills/landing-page-copy
    · codex → ~/.agents/skills/landing-page-copy
    · cursor → ~/.agents/skills/landing-page-copy
    · gemini-cli → ~/.agents/skills/landing-page-copy
    · claude-code → ~/.claude/skills/landing-page-copy
    · factory → ~/.factory/skills/landing-page-copy
```

Markers are muted (`·`) instead of green (`✓`) so the visual hierarchy still distinguishes "already installed" from "just installed."

## Test plan
- [x] C-INST-11 updated to lock in the new per-agent row.
- [x] `bun run check` green: 804 tests, 100% coverage.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #30, opened 2026-04-23
[LOG-4504] feat(search): list everything installable, mark installed with ✓

stevekrenzel:
> **Stacked on #29.** That PR is stacked on #28. Merge those in order first.

## Summary
`crew search` learns two new tricks:

1. **No-query form** — `crew search` with no argument lists every skill in every configured tap, grouped by tap. The exhaustive catalog.
2. **Installed marker** — every row is prefixed with `✓` (green, TTY-only) if the skill is already in local state, and a space otherwise. Works for both the no-query listing and the query form.

Also made search **namespace-aware** — it now walks taps via the `indexTap` helper shipped in the previous PR, so namespaced skills show up as `marketing/email-outreach`.

JSON output gains `installed: boolean` and `namespace: string | null`.

## Example
```
$ crew install marketing/email-outreach
$ crew search
  3 skills available

  acme
    ✓ marketing/email-outreach   Send personalized cold emails
      marketing/social-posts     Draft LinkedIn threads
      engineering/code-review    Structured review pass

  Install any of these with `crew install <name>`.
```

## PRD
- §16.4 rewritten to cover both the query and no-query forms and to name the `✓` marker.
- C-TAP-08 tightened (JSON shape now includes `installed`).
- New C-TAP-08b for no-query behavior.

## Test plan
- [x] Existing search tests updated for the new JSON shape — all 11 pass.
- [x] New tests: no-query lists all, installed marker works, empty-tap-list message, unreachable tap warning.
- [x] `bun run check` green: 804 tests, 100% coverage.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #32, opened 2026-04-23
[LOG-4506] feat(install): soft-fail invalid skills during multi-skill expansion

stevekrenzel:
> **Stacked on #31 → #30 → #29 → #28.** Merge those in order first.

## Summary
One bad apple used to spoil the batch. If a tap had 7 skills and one had a \`name\`-vs-directory mismatch, the whole install aborted with \`invalid_skill\` exit 4 and the other 6 valid skills got nothing.

Now:

- **Multi-skill expansions** (walking \`skills/\` or a source root) soft-skip invalid children. Each skipped skill is collected with its validation message; valid siblings proceed through the full install flow.
- **Exit code** is 1 when anything was skipped (was 4), matching the "partial success" semantics we already use for per-agent failures.
- **Single-skill sources** (root SKILL.md, or a qualified \`<tap>/<skill>\` / \`<tap>/<ns>/<skill>\`) continue to hard-fail with exit 4 — the user asked for that one skill and partial success doesn't exist.
- The renderer gets a new **"Skipped"** section at the bottom of the output, naming each path and reason, alongside a summary line that includes the skipped count.
- \`--json\` output gains a \`skipped: SkippedSkill[]\` field.

## Example output
```
Installing 6 skills

  software-architect
    …
    ✓ claude-code → ~/.claude/skills/software-architect
    …

  (… 5 more blocks …)

  30 installs · 1 skipped

Skipped (1):
  ✗ name: \`plan-design-review\` doesn't match the parent directory name \`founding-ux-designer\` — rename one so they match
```

## PRD
- §9 step 4 rewritten to name the single-skill vs. multi-skill validation policy.
- New conformance rows C-INST-20 (soft-skip + exit 1) and C-INST-21 (single-skill hard-fail).

## Test plan
- [x] C-INST-20: invalid sibling is soft-skipped, valid sibling installs, exit 1.
- [x] C-INST-20 JSON: \`skipped\` field is populated.
- [x] C-INST-21: single-skill source still hard-fails with exit 4.
- [x] \`bun run check\` green: 807 tests, 100% coverage.

### DIFF
--- a/PRD.md
+++ b/PRD.md
@@ -687,7 +687,8 @@ Given one or more skill references on the command line, `crew install` proceeds 
    - `description` is present, non-empty, length ≤ 1024 characters.
    - If `compatibility` is present, length ≤ 500 characters.
    - Every other spec rule from the Agent Skills specification.
-   Invalid skills abort with `invalid_skill` (§13) before any files are written.
+
+   **Validation policy.** For **single-skill** sources (a source whose resolved location IS one skill — case 1 in step 5 below, or a 3-segment `<tap>/<ns>/<skill>` or 2-segment `<tap>/<skill>` reference), an invalid skill aborts the whole command with `invalid_skill` (§13). For **multi-skill expansions** (cases 2 and 3 in step 5 — walking a `skills/` directory or the source root), invalid skills are **soft-skipped**: each invalid skill is recorded as a skip with its validation message, valid siblings continue to install, and the command exits `1` (not `4`) if any skills were skipped. This avoids one bad skill blocking a whole-tap install. The CLI renders skipped skills in a final "Skipped" section naming each offending path and the validation error.
 5. **Expand directories.** Three cases, checked in order:
    1. If the resolved source location has a `SKILL.md` at its root, it is one skill.
    2. Else, if the resolved source location has a `skills/` subdirectory, crew walks under `skills/` and collects skills:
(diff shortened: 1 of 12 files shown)


stevekrenzel:
Addressed every item — the bug, both file-size violations, the missing test, and all three nits.

**Applied**
- **Dead fallback in `pushSoft`**: dropped `?? "invalid_skill"`. `loadSkill` only throws `CrewError` and `CrewError.code` is a non-null `CrewErrorName`. A non-CrewError here would indicate a bug in `loadSkill` and should surface loudly; hiding it behind `invalid_skill` was a latent hazard.
- **File-size cap — `install/index.ts`**: extracted `promptForCollision` + `abortError` to `collision-prompt.ts`. Top-level command is now 155 lines.
- **File-size cap — `install/render.ts`**: this was already resolved in the #31 review pass (split into `render/index.ts` + `render/helpers.ts`). Current sizes: 175 and 77 lines. The review was reading pre-#31 line numbers.
- **Missing C-INST-21-equivalent test**: added a test for the "every skill in the multi-skill dir is invalid" edge. Asserts exit 4 and that every invalid skill is listed in the Failed section.
- **Nit: unreachable `"nothing to do"`**: removed the fallback; replaced with a comment explaining why `parts` is never empty past the early-return guard. `return parts.join(" · ")`.
- **Nit: `wrapped[0] ?? s.message`**: dropped the fallback; `wrap` always returns ≥1 element for a non-empty message. Changed to `wrapped[0]!` with a short comment.
- **Nit: `info` silently drops skipped skills**: added a docstring on `buildSkillInfos` making the design choice explicit. `info` is read-only "show me what's valid here"; `crew install` is the surface that reports validation failures (§9 step 9).

**Not applied (intentional)**
- **`src/install/resolve.ts` (387 lines) pre-existing file-size violation**: flagged by the reviewer as "not introduced by this PR, worth knowing it needs attention separately." Was split in the PR #29 review pass into `resolve/index.ts` (160 lines) + `resolve/enqueue.ts` (245 lines — still slightly over 200, but well under the starting number). Downstream in this stack, not addressable here beyond what #29 did.

`bun run check` green: 809 tests, 100% coverage.

jgarms (approved):
🚀

PR #42, opened 2026-05-06
[LOG-4705] Skip Linear ticket creation when unavailable

stevekrenzel:
## Summary
- Skip Linear ticket creation when the Linear API key is unavailable in a workflow context
- Preserve the existing early return for PR titles that already include a ticket marker

## Why this can happen
GitHub Actions does not expose repository secrets to every pull request workflow run. Untrusted actors such as Dependabot and PRs from forks can trigger the workflow, but GitHub withholds secrets like `LINEAR_API_KEY` so those runs cannot exfiltrate private credentials. In those contexts `secrets.LINEAR_API_KEY` resolves as unavailable/empty, so Linear ticket creation needs to be treated as an optional integration instead of a required check.

## Verification
- git diff --check
- bun run lint

### DIFF
--- a/.github/workflows/create-linear-ticket.yml
+++ b/.github/workflows/create-linear-ticket.yml
@@ -51,16 +51,16 @@ jobs:
             const repoOwner = process.env.REPO_OWNER;
             const repoName = process.env.REPO_NAME;
 
-            if (!apiKey) {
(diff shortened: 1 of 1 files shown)


markgolazeski (approved):
LGTM!

PR #44, opened 2026-05-07
[LOG-4708] Verify release downloads with checksums

stevekrenzel:
## Summary
- Make the hosted installer download SHA256SUMS from the selected GitHub release and verify the chosen macOS binary before installing
- Make `crew self-update` download SHA256SUMS, verify the selected binary, and fail with `self_update_unavailable` before replacement on missing or mismatched checksums
- Add SHA256SUMS to the generated latest-version.json release feed
- Document the checksum requirement in the PRD, README, site install section, and self-update command reference

## Verification
- bash -n site/public/install.sh
- bash -n scripts/release.sh
- git diff --check
- bun run lint
- cd site && bun run lint
- cd site && bun run check
- bun test tests/self-update (tests passed; partial-suite coverage threshold exits non-zero as expected)
- bun run check
- cd site && bun run check

### DIFF
--- a/PRD.md
+++ b/PRD.md
@@ -938,9 +938,10 @@ binaries, the site just advertises the pointer).
 
 Every release MUST publish a `SHA256SUMS` asset containing SHA-256 hashes
 for `crew-macos-arm64` and `crew-macos-x64`. The hosted installer
-(`https://crew.logic.inc/install.sh`) MUST download this file from the same
-release as the selected binary and verify the chosen asset before installing
-it. A checksum mismatch or missing checksum is a failed install.
+(`https://crew.logic.inc/install.sh`) and `crew self-update` MUST download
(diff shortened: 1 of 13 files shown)


markgolazeski (dismissed):
LGTM -- nit: I think I'm in agreement with the `Same-source integrity, not supply-chain authenticity.` suggestion, but non-blocking.

stevekrenzel:
@claude – we added signature verification to the `crew self-update` path as well. Can you re-review this PR? We

stevekrenzel:
@claude – we added signature verification to the `crew self-update` path as well. Can you re-review this PR? We'll handle supply-chain hardening in a fast-follow PR.

markgolazeski (approved):
LGTM!

PR #45, opened 2026-05-07
[LOG-4710] Add ergonomic command aliases

stevekrenzel:
## Summary
- add `crew skills` as a top-level alias for `crew list`
- add `crew taps` as a top-level alias for `crew tap list`
- document both aliases in the PRD, README, overview help, and structured help output

## Verification
- `bun run check`


src/cli/dispatch.ts
```diff
@@ -40,15 +40,22 @@ export const COMMAND_HANDLERS: Record<string, CommandHandler> = {
   version: versionCommand,
 };
 
+const COMMAND_ALIASES: Record<string, string> = {
+  skills: "list",
+  taps: "tap",
```
stevekrenzel: I would have expected this to be:

```suggestion
  skills: "list",
  taps: "tap list",
```

dutabado (commented):
LGTM!

dutabado (approved):

PR #46, opened 2026-05-07
[LOG-4711] Trigger Claude from PR mentions

stevekrenzel:
## Summary
- keep automatic Claude review on PR open using the existing review prompt
- add `@claude` triggers for PR top-level comments, inline review comments, and submitted review bodies
- let mention-triggered runs use the comment context directly instead of the review prompt, so Claude can answer questions or make requested changes

## Notes
- GitHub uses the workflow definition from `main` for comment-triggered workflows, so `@claude` mention handling will only become active after this PR merges.

## Verification
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/claude.yml"); puts "yaml ok"'`\n- `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/claude.yml`\n- `git diff --check`\n

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -1,13 +1,19 @@
-# Claude Code PR review — runs automatically on every opened PR.
+# Claude Code PR review + mention responder.
 #
-# Posts a review with inline comments grounded in CLAUDE.md's working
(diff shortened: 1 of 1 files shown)


dutabado (approved):
LGTM!

PR #47, opened 2026-05-07
[LOG-4712] Verify signed release checksums

stevekrenzel:
## Summary

- sign release `SHA256SUMS` with an RSA/SHA-256 release key and publish `SHA256SUMS.sig` with every future release
- verify the checksum signature in both `crew self-update` and the hosted installer before trusting binary checksums
- create releases as drafts, upload all assets, then publish so the workflow is compatible with GitHub immutable releases
- update PRD, README, site copy, and release metadata generation for the new signed-checksum contract
- make the invalid git URL test deterministic by using a missing local `file://` repo instead of DNS

## Immutable releases

This branch is compatible with immutable releases. GitHub recommends creating a draft release first, attaching all assets, and publishing only after the assets are present. The release workflow now does exactly that: build binaries, generate `SHA256SUMS`, sign it, create a draft release with all four assets, then publish the release.

## Key management

The public key is committed in `src/self-update/signing-key.ts` and embedded in `site/public/install.sh`. The matching private key has been installed as the `CREW_RELEASE_SIGNING_PRIVATE_KEY` GitHub Actions secret and the local private-key file was removed.

## Verification

- `bun run typecheck`
- `bun run lint`
- `bash -n site/public/install.sh && bash -n scripts/release.sh`
- workflow YAML parse check
- `actionlint .github/workflows/release.yml`
- `cd site && bun run check`
- `bun test tests/self-update` (90 tests passed; subset exits nonzero because global coverage is enforced per invocation)
- `bun run check` (827 tests, 100% line/function coverage)
- manual OpenSSL sign/verify smoke test with the generated release key


stevekrenzel:
@claude addressed the review feedback in `7da65a8`.

- Added a `runSelfUpdate` integration test for the legacy `v0.7.0` path where the release has `SHA256SUMS` but no `SHA256SUMS.sig`. The test uses `force: true` because the repo version is currently `0.7.0`, so without force the self-update path correctly no-ops before downloading assets.
- Replaced the misleading `url.includes("SHA256SUMS")` fixture branch with exact `CHECKSUMS_URL` / `CHECKSUMS_SIGNATURE_URL` comparisons.

Verified locally with `bun run check`: 828 pass, 0 fail, 100% function/line coverage.

dutabado (approved):
LGTM!

PR #48, opened 2026-05-07
[LOG-4712] Add known tap registry foundation

stevekrenzel:
## Summary
- Specify a bundled known-tap registry snapshot in the PRD, with explicit no-first-run-cloning semantics
- Add typed known-tap registry/search helpers for future search/install integrations
- Add unit coverage for registry search, lookup, labels, and tap-config conversion

## Notes
This PR intentionally does not change CLI output or install behavior yet. The registry is inert until later PRs explicitly wire it into search/install flows.

On the discovery design: we should ship a precomputed registry snapshot rather than clone many repos on first run. Git sparse/partial clone can reduce object transfer for a selected repo, but it still requires one network operation per repo and does not avoid needing an index source. The bundled snapshot keeps first-run behavior offline/fast, and normal git cloning only happens after the user chooses to tap a known repo.

## Verification
- bun run check

### DIFF
--- a/PRD.md
+++ b/PRD.md
@@ -1310,9 +1310,10 @@ populate the registry.
 Each known tap entry has:
 
 - `name` — the tap name crew would use if the user chooses to add it.
-- `url` — git clone URL.
+- `url` — git clone URL. All known taps are git taps; path-kind taps are local
+  user state and are never shipped in the known-tap registry.
 - `subpath` — POSIX path inside the repo that forms the tap root; empty for repo
-  root.
+  root, represented as the empty string `""`.
 - `description` — short human-readable tap description.
 - `trust` — either `official` (published by the owner of the tool/service the tap
   represents) or `curated` (reviewed and selected by the Homecrew maintainers).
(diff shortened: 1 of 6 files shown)


stevekrenzel:
Added a compact JSON example to PRD §16.2.1 showing the known-tap registry shape, including `skills` as a list of precomputed skill summaries. I used a placeholder URL so the example documents shape without implying that specific repo exists yet.

stevekrenzel:
@claude addressed the review feedback in `4b16eb4`.

- Fixed known-tap name and skill-label matching to lowercase the source before comparing against the normalized query.
- Added coverage for tap-name matches (`supa`) and mixed-case tap/namespace/skill labels (`OpenAI`, `Evals/Prompt-Eval`).
- Moved `tapConfigFromKnownTap` out of `types.ts` into `known-taps/convert.ts` so the types file stays type-only.

Verification: `bun run check` passed locally on the second full run: 825 pass, 0 fail, 100% function/line coverage. The first full run hit an unrelated timeout in an existing git e2e test, then passed cleanly on rerun.

stevekrenzel:
Addressed the review feedback in the latest push:

- Documented that known taps are git-only registry entries and that root `subpath` is represented as `""`.
- Documented the tap-level search behavior: matching a tap name/description includes every skill in that tap.
- Made exact known-tap lookups case-insensitive while keeping them exact, not substring/fuzzy.
- Added comments clarifying fuzzy search vs exact lookup and that `tapConfigFromKnownTap` is for user opt-in tap registration, not discovery rendering.
- Added/renamed tests for empty registry lookup, case-insensitive lookup, and separate tap-name vs namespaced-label search matching.

Validation: `bun run check` passes: 827 tests, 100% line/function coverage.

dutabado (approved):
LGTM!

PR #49, opened 2026-05-07
[LOG-4713] Fix ambiguous tap install test prompt

stevekrenzel:
## Summary
- Make the ambiguous bare-name tap install test inject an abort prompt instead of inheriting the caller's TTY stdin.
- Assert the ambiguous-reference error text so the test proves it hit the intended failure path.

## Test plan
- bun test tests/e2e/tap-install.test.ts (test file passes; single-file run exits non-zero because global coverage is intentionally incomplete)
- bun run check

### DIFF
--- a/tests/e2e/tap-install.test.ts
+++ b/tests/e2e/tap-install.test.ts
@@ -127,7 +127,7 @@ describe("tap source install", () => {
     const code = runCli(["install", "alpha"], {
       home,
       streams: c.streams,
-      prompt: () => "abort",
+      promptChoice: () => "abort",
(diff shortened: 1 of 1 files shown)


stevekrenzel:
Fixed. The ambiguous bare-name test now injects `promptChoice: () => "abort"`, which is the seam used by `promptBareNameAmbiguity` / the numbered-menu path.

I reran `bun run check` successfully: 818 tests, 100% line/function coverage.

dutabado (approved):
LGTM!

PR #50, opened 2026-05-07
[LOG-4714] Make install refs case-insensitive

stevekrenzel:
## Summary
- Canonicalize tap-source identifiers to lowercase during parsing while leaving path sources and git URLs unchanged.
- Make bare, qualified, and namespaced install refs work regardless of user-entered case.
- Document the behavior in the PRD and add C-REF-21 coverage.

## Test plan
- bun test tests/refs/parse.test.ts
- bun test tests/e2e/tap-install.test.ts tests/e2e/namespaces.test.ts
- bun run check

### DIFF
--- a/PRD.md
+++ b/PRD.md
@@ -648,6 +648,13 @@ tap-ref     := any non-empty string not containing "/" or whitespace
 subpath     := any POSIX relative path not starting with "/"
 ```
 
+Tap-source identifiers are matched case-insensitively and canonicalized to
+lowercase before lookup. For example, `crew install Core/Python-Testing`
+resolves identically to `crew install core/python-testing`. This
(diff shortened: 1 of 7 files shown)


stevekrenzel:
Addressed the review nits in the latest push:

- Removed the duplicate `TAP_REF_NAME_PATTERN` regex and now normalize each segment before validating it with the existing `NAME_PATTERN`.
- Clarified invalid tap/name error messages so they describe invalid characters rather than case behavior.
- Renamed the two `C-REF-21` tests so the reporter output distinguishes bare vs 3-segment canonicalization.

Validation: `bun run check` passes: 821 tests, 100% line/function coverage.

dutabado (approved):
LGTM!

PR #51, opened 2026-05-07
[LOG-4718] Suggest known taps in search results

stevekrenzel:
## Summary
- show known-but-untapped search suggestions for non-empty `crew search` queries after configured tap results
- omit known taps that are already configured by name or source, and keep no-query search configured-only
- update PRD/help/README/site copy and split the search/homepage command reference files to keep them under the file-size cap

## Tests
- `bun run check`
- `cd site && bun run check`


stevekrenzel:
Addressed Claude's feedback and pushed `150e12a`.

- Added C-TAP-23 coverage for a known tap already configured under a different name but the same `(url, subpath)`, so the URL+subpath dedupe branch is covered.
- Restored rendered `<code>` styling for the `crew skills` / `crew taps` alias descriptions without putting `Commands.data.tsx` over the file-size cap.
- Added the note that subpath comparison is intentionally case-sensitive because subpaths are filesystem paths.

Checks:
- `bun run check`
- `cd site && bun run check`

src/commands/search.ts
```diff
@@ -1,165 +0,0 @@
-/**
```
stevekrenzel, line 1: I *love* that our 200-line max rule forced us to refactor this file into a directory of a bunch of smaller files.

stevekrenzel:
Addressed the additional `REVIEW.md` notes and pushed `16e5ff7`.

- Clarified in §16.6 that known-tap dedupe compares tap names and URLs case-insensitively, while subpaths compare exactly.
- Passed the normalized query into both configured-tap and known-tap collectors.
- Replaced the repeated `tapHits[0]!` reads in the known-tap renderer with a named `firstHit` and a comment explaining why the non-null assertion is safe.

I left the `Homecrew` wording in `crew help search` as-is because the rest of the help content already uses `Homecrew` for product-level behavior and `crew` for command invocations.

Checks:
- `bun run check`
- `cd site && bun run check`

dutabado (approved):
🚀

PR #52, opened 2026-05-07
[LOG-4720] Suggest known taps for install misses

stevekrenzel:
## Summary
- enrich `crew install` invalid_ref misses with exact known-tap suggestions and JSON details
- document C-TAP-24 and update README/help/site copy
- add e2e coverage for bare, qualified, JSON, ref-preserving, and configured-source suppression cases

## Tests
- `bun run check`
- `cd site && bun run check`

### DIFF
--- a/PRD.md
+++ b/PRD.md
@@ -689,6 +689,7 @@ Given one or more skill references on the command line, `crew install` proceeds 
    - Git URL or shorthand (`@org/repo`, `gh:org/repo`, etc.): if any configured tap already points at the same URL+subpath, use it; otherwise create an auto tap (§16.5) and use it. The new tap is cloned and refreshed.
    - Path reference (`./foo`, `/abs/foo`): if any configured tap already points at the same path, use it; otherwise create a path-kind auto tap and use it.
    In all cases, the resolved `state.installations[i].source` is `{ tap: <tap-name>, path: <skill-relative-path-inside-tap> }`. The URL/path of the tap itself lives in `config.yaml`.
+   - If a tap-source reference cannot be resolved from configured taps, Homecrew consults the known-tap registry (§16.2.1) before returning `invalid_ref`. Exact known-tap matches are suggestions only: Homecrew MUST NOT clone, fetch, add a tap to config, or install anything from a known tap until the user explicitly runs `crew tap add <url>[//<subpath>] <name>` (or a future interactive flow confirms that same action). The error MUST include the tap-add command and the follow-up `crew install <tap>/<skill>` or `crew install <tap>` command.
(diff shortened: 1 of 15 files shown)


stevekrenzel:
@claude addressed the review feedback.

- Added a C-TAP-24 e2e case for multi-ref install misses, confirming the known-tap suggestion block is intentionally suppressed when `refs.length !== 1`.
- Added a short comment explaining why multi-ref `invalid_ref` errors are not enriched.
- Narrowed `sameNullableText` so the second argument is a non-null string and removed the dead `b === null` branch.
- Kept the flattened `known-tap-fallback/` split and all touched files under the 200-line cap.

Validation:
- `bun run check` passed with 100% coverage after the code/test changes.
- `bun run lint` passed after the final test-title cleanup.

stevekrenzel:
Addressed the `REVIEW.md` follow-up notes.

Changes pushed:
- Hoisted known-tap shared helpers into `src/known-taps/`:
  - `configured.ts` for `knownTapIsConfigured`
  - `format.ts` for URL/subpath source formatting
  - `text.ts` for case-insensitive text comparisons
- Updated both `crew search` known-hit rendering and `crew install` known-tap fallback to use the shared helpers.
- Removed the install-specific `known-tap-fallback/text.ts` duplicate.
- Tightened `KnownInstallSuggestionJson.trust` to `KnownTapTrust`.
- Changed bare known-tap matching to prefer a same-named skill over a same-named tap suggestion, avoiding duplicate tap-add guidance within one tap. Added a C-TAP-24 e2e regression test for that case.
- Tightened PRD §9 from optional wording to `consults` / `MUST include` so it matches the C-TAP-24 contract and tests.
- Added the small test setup comment/named stream from the review note.

Validation:
- `bun run check` passed with 859 tests and 100% coverage.
- File-size cap checked; touched files are all under 200 lines.

stevekrenzel:
Addressed the new `REVIEW.md` findings. Both findings were valid.

- Forced-kind flags: passed `kindHint` into the known-tap fallback and now filter suggestions so `--tap` only permits tap suggestions, `--skill` only permits skill suggestions, and `--bundle` suppresses suggestions until we model namespace suggestions explicitly. Added C-TAP-24 e2e coverage for `--tap`, `--skill`, and `--bundle` so we do not emit contradictory guidance.
- Multi-ref misses: removed the single-ref suppression. The fallback now scans tap-source refs in the command, skips refs that do not parse as tap refs, dedupes suggestions, and includes exact known-tap matches even when another positional caused the `invalid_ref`. Updated the multi-ref C-TAP-24 test to assert that behavior.

Validation:
- `bun run check` passed: 862 tests, 100% line/function coverage.
- File-size cap checked; touched files are under 200 lines.

stevekrenzel:
@claude - can you review again?

dutabado (approved):
🚀

PR #53, opened 2026-05-08
[LOG-4729] Build known tap registry artifact

stevekrenzel:
## Summary
- Add a pinned known-tap source manifest and generated registry snapshot.
- Add a unified `bun run known-taps <command>` maintenance CLI with `build`, `check`, `update`, and `add` subcommands.
- Keep compatibility package scripts for `build-known-taps`, `check-known-taps`, and `update-known-tap-pins` as wrappers around the unified command.
- Wire `known-taps check` into `bun run check` / `bun run build`, and regenerate the snapshot from reviewed pins during `bun run release`.
- Add `known-taps add <name> <url>` to resolve the initial pin, validate/index the tap, write a sorted manifest entry, and regenerate the snapshot.
- Document the operating model for adding taps, refreshing pins, and cutting releases.

The manifest is intentionally empty in this PR; this adds the release-safe pipeline first. Adding real curated sources should be a separate content PR once we decide which taps to bless.

## Design Notes
- The bundled snapshot is generated from pinned commit SHAs, not upstream default branches. `trackingRef` is only refresh metadata.
- `bun run release` regenerates the snapshot from already-reviewed pins, but does not advance those pins.
- Pin advancement goes through `bun run known-taps update <tap-name>` or `bun run known-taps update --all`, producing a normal reviewable diff.
- New taps go through `bun run known-taps add <name> <url>`, which resolves the initial pin before writing `known-taps/manifest.json`.
- Manifest writes for `add` / `update` happen only after the generated registry can be rebuilt, so failed indexing does not leave a stale manifest/snapshot pair.
- The pin is build-time provenance only: users still add normal git taps, and those taps follow the existing `crew tap add` / update rules.

## Review Feedback Addressed
- Added malformed `description` manifest coverage and root-skill `namespace: null` coverage.
- Changed docs and generated headers to the unified `bun run known-taps build` / `check` flow.
- Added duplicate known-tap name validation.
- Made `add` / `update` build the generated registry before writing the manifest.
- Added a safer disposable `--work-dir` guard and removed the mutable cleanup-dir script pattern.
- Moved routine update/add output to stdout, kept errors on stderr.
- Deduplicated SHA validation and made `ls-remote` parsing match the tab-separated git output format.
- Improved stale generated-registry errors with the first mismatched line number.

## Verification
- `bun run known-taps build`
- `bun run known-taps check`
- `bun run build-known-taps`
- `bun run update-known-tap-pins --all`
- `bun run check`

Linear: LOG-4729


stevekrenzel:
@claude addressed the review feedback:

- Added malformed `description` manifest coverage.
- Asserted root-skill entries render with `namespace: null`.
- Updated the registry docs to recommend the full `bun run check` gate after regenerating the snapshot.

stevekrenzel:
Addressed the review notes from `REVIEW.md` / `REVIEW-2.md` and folded in the unified known-taps command:

- Replaced the separate build/update script entrypoints with `bun run known-taps build|check|update|add` while keeping package-script compatibility wrappers.
- Added `known-taps add <name> <url>` to resolve the initial pin, validate/index, sort the manifest, and regenerate the snapshot.
- Made add/update atomic with respect to manifest writes: the registry is built first, then manifest + generated snapshot are written.
- Added duplicate tap-name validation, shared SHA validation, tab-specific `ls-remote` parsing, safer work-dir cleanup, stdout for routine script output, and a clearer stale snapshot error.
- Left shallow/fetch-by-SHA cloning as a follow-up because it changes the git acquisition strategy and is performance-oriented, not correctness-blocking.

`bun run check` is green locally: 879 tests, 100% coverage.

dutabado (dismissed):

stevekrenzel:
Rebased this onto latest `origin/main` after #56/#57 merged and fixed the known-tap builder test fixture that became stale after relaxed skill-name/source-directory semantics. CI is green now, but the force-push dismissed the prior approval, so this needs a fresh approval before it can merge.

dutabado (approved):

PR #57, opened 2026-05-11
[LOG-4731] Relax skill name validation

stevekrenzel:
As we started adding known-taps, it turns out in practice a lot of people deviate from the spec. Like the directory and name in SKILL.md should be the same, but they aren't always. And the skill should start with an alpha-char, but they don't always.

This fixes that.

## Summary
- Treat `SKILL.md` frontmatter `name` as the authoritative skill name for validation, tap indexing, search, installs, dependencies, and collision checks.
- Allow skill and tap-source identifiers to begin with a digit so names like `3-statement-model` can be installed.
- Update PRD, README, help text, and homepage tap copy to document that source directory names do not need to match skill names.

## Tests
- `bun run check`
- `cd site && bun run check`


stevekrenzel:
Addressed the review notes from REVIEW-1.md / REVIEW-2.md in 434acec.

Changes made:
- `crew tap add` now uses the shared reference name pattern, so explicit tap names can start with digits too.
- Tap re-expansion now preserves the actual tap-relative source path when a new child directory differs from the declared `SKILL.md` name, and source-gone detection compares `(source.path, name)` instead of name only.
- Clarified the optimistic tap-index behavior and root-skill tap comment.
- Tightened invalid-name wording and cleaned up the test name called out in review.

Validation: `bun run check` passes locally: 873 tests, 100% coverage.

stevekrenzel:
Addressed the second REVIEW-1.md / REVIEW-2.md pass in 02e040f.

Fixes included:
- `topoSort` dependency extraction now uses the shared name grammar, so digit-leading dependencies install before dependents.
- `crew info` now resolves uninstalled tap refs through the tap index / resolver, so declared `SKILL.md` names work for bare, qualified, and namespace info lookups.
- Duplicate declared names from different tap-relative source paths now fail as `conflicting_dependencies`, and the PRD/conformance text reflects that.
- Auto tap-name derivation preserves leading digits.
- Tap re-expansion now treats declared name as the upstream identity and updates `source.path` when the directory moves instead of reporting `source_gone`.
- Cleaned up the stale invalid-skill test comment.

Validation: `bun run check` passes locally: 880 tests, 100% line/function coverage.

stevekrenzel:
Fixed the `bun run check` flake in d1bce54.

Root cause: a few coverage-closeout tests were accidentally doing real network work during `crew update` because the default `core` tap remained configured, and one SSH-style tap-add test depended on real DNS/SSH timeout behavior.

Fix: remove the default `core` tap in those local update coverage tests and stub the git runner for the unreachable SSH tap-add test.

Validation: `bun run check` now passes locally: 880 tests, 100% line/function coverage.

stevekrenzel:
Addressed the latest REVIEW-1/REVIEW-2 pass in `3e7065d`:

- duplicate declared names during tap re-expansion now produce `conflicting_dependencies`, leave existing state paths unchanged, and make update exit non-zero
- split `crew update` selection helpers and tap child discovery into smaller files to keep command/update logic focused
- removed the dead `findTapForBareName` helper and its unit test
- removed the unsafe `crew info` candidate cast by adding an internal non-tap resolver path, with missing/ambiguous info coverage
- clarified PRD/help wording around source directory names and duplicate declared names

Validation: `bun run check` passes: 878 tests, 100% function/line coverage.

stevekrenzel:
Addressed the last REVIEW-1/REVIEW-2 pass in `37dd6d9`:

- indexed root-skill taps by declared `SKILL.md` name, with search/install/info coverage
- insulated the new `crew info` tests from the default core tap
- fixed sibling dependency attribution when the sibling lives outside the parent tap root by giving that sibling its own auto tap, so update resolves from a valid source path
- made duplicate declared sibling dependencies fail deterministically with `conflicting_dependencies`
- split dependency enqueueing out of `resolve/enqueue.ts` to keep touched files under the 200-line cap
- handled the remaining PRD/comment/test-name nits, including the `conflicting_dependencies` §13 wording and root tap child test

Validation: `bun run check` passes: 885 tests, 100% function/line coverage.

stevekrenzel:
@claude – we've made a few changes since the last review. Can you give it a once over?

stevekrenzel:
Addressed Claude’s three followups in `b343f69`:

- `loadSkillName` now checks for missing `SKILL.md` and throws the same `invalid_skill` `CrewError` shape as `loadSkill`
- `countSkills` now validates root-skill tap names before counting them, so invalid root skills count as zero like indexed children
- added the happy-path unit test for `findSiblingDep` returning the declared-name sibling and tap-relative path

Also cleaned up the stale validation-test comment and split frontmatter/load-name tests out so touched test files stay under the file-size cap.

Validation: `bun run check` passes: 888 tests, 100% function/line coverage.

dutabado (approved):

PR #56, opened 2026-05-10
[LOG-4724] Add Logic attribution

stevekrenzel:
## Summary
- identify Homecrew in the README as an open-source project by Logic, Inc.
- add restrained Logic attribution in the site footer, FAQ, and metadata description
- keep the copyright attribution linked to logic.inc

## Tests
- bun run check
- (site) bun run check

PR #58, opened 2026-05-11
[LOG-4732] Add recursive tap discovery opt-in

stevekrenzel:
Some repos have weird deeply nested skills directories. This allows us to opt-in to that behavior.

If you tap a tap that doesn't meet one of our expected patterns, you can pass `--recursive` and see if that discovers anything.

(Note: This could become default behavior one day, but I think the implications need to be thought out a bit more)

## Summary
- add an explicit recursive discovery mode for trusted non-standard tap layouts via `crew tap add --recursive` and direct `crew install --recursive <git-or-path>`
- persist recursive discovery in config/markers so search, install, info, doctor repair, and whole-tap update re-expansion use the same mode
- document the new behavior in the PRD, README, help text, and site copy
- preserve `crew info <tap-name>` precedence when a tap name also matches a skill elsewhere
- note: tap re-expansion now treats only `no_skills_found` as an empty tap; filesystem errors bubble instead of being converted to source-gone rows

## Review notes
- `discovery: standard` is now documented as a no-op spelling equivalent to omitting `discovery`
- marker `schema_version` now explicitly allows backward-compatible optional additive fields without a bump
- recursive tap downgrade semantics are documented: config edits affect live resolution, while existing markers retain install-time discovery until rewritten
- `--recursive` dependency non-propagation and auto-tap promotion/upgrade behavior are covered in docs/tests

## Verification
- `bun run check`


dutabado (dismissed):

stevekrenzel:
Rebased this onto latest `origin/main` after #56/#57 merged and resolved conflicts with the selector/name-validation work. Local `bun run check` and GitHub CI are green, but the force-push dismissed the prior approval, so this needs a fresh approval before it can merge.

dutabado (approved):

PR #60, opened 2026-05-11
[LOG-4734] Accept qualified installed skill names

stevekrenzel:
## Summary

- add installed-skill selector resolution so `crew uninstall anthropic/pdf` works for a skill installed as `pdf`
- extend the same state-oriented behavior to `crew update <tap>/<skill>` and `crew info <tap>/<skill>`
- update the PRD conformance language and CLI help examples

## Product language note

The PRD uses `selector` as precise spec terminology, but user-facing help intentionally uses `name`. To users, both `pdf` and `anthropic/pdf` are skill names; `selector` is colder and more technical.

## Verification

- `bun run check`

### DIFF
--- a/PRD.md
+++ b/PRD.md
@@ -780,6 +780,22 @@ Given one or more skill references on the command line, `crew install` proceeds 
 
 Exit code: 0 if every skill succeeded in at least one agent; 1 if any skill failed in every agent; 2 if nothing was attempted (empty install set after expansion when the user explicitly asked for something). Other exit codes per §15.
 
+### 9.1 `crew info`
+
+`crew info` accepts any install reference and, for installed skills, the same
+installed-skill selectors defined in §5.1. It reads `state.json` without taking
+the state lock. If the argument matches one or more installed state entries,
+`crew info` renders the installed view first: source, resolved version, installed
+agents, and local description when available. A tap-qualified selector such as
+`<tap>/<skill>` or `<tap>/<namespace>/<skill>` narrows the installed entries by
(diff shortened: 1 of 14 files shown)


stevekrenzel:
Addressed the review feedback in `39cd01d`:

- Added a PRD `crew info` section plus `C-INFO-01`, and renamed the info e2e test accordingly.
- Fixed installed-state selector parsing so tap refs with `@ref` tails do not silently match installed state. `info` falls through to reference preview; `update`/`uninstall` report the normal non-installed errors. Added unit and e2e coverage for all three commands.
- Re-resolve uninstall args against the current state on each iteration, avoiding stale `StateSubject.entries` for overlapping inputs.
- Removed the unused `state` parameter from `tapsToRefreshFor` and added a comment clarifying resolved names vs raw inputs.
- Added comments for the `StateSubject.raw`/`.name` invariant and namespace path assumption.
- Collapsed update dependency selection back into a single pass.
- Added README/site copy that keeps human-facing “name” language while documenting bare and tap-qualified names.

Verification: `bun run check` passes: 877 tests, 100% coverage.

dutabado (dismissed):

stevekrenzel:
Rebased this onto latest `origin/main` after #56/#57 merged and resolved conflicts in the shared update-selection helper. Local `bun run check` is green; GitHub CI is green as well, but the force-push dismissed the prior approval, so this needs a fresh approval before it can merge.

dutabado (approved):

PR #61, opened 2026-05-11
[LOG-4723] Polish CLI error output

stevekrenzel:
## Summary
- Render human errors as readable blocks with an explicit Next step instead of dense one-line messages
- Reword common tap, command, install-miss, and search empty-state copy so the offending value and remedy are easier to scan
- Tighten README and homepage positioning around Homecrew as a Logic open-source project for sharing agent skills through git
- Keep PRD, README, site copy, CLI help, and search suggestion tests aligned with the polished public wording

## Notes
This branch is rebased on the latest main after the known-taps, catalog, recursive discovery, and qualified-name work merged. LOG-4738 (`crew untap`) and LOG-4740 (source-aware search install markers) are still separate PRs; this pass avoids duplicating those implementations while keeping adjacent wording ready for them.

## Test plan
- bun run format
- bun run check


stevekrenzel:
Addressed the duplicate guidance feedback in `6e140bc`.

Changes made:
- Added an optional `CrewError.remedy` override so error call sites can provide one context-specific next step, suppress it with `null`, or fall back to the shared remedy table.
- Moved embedded `\n\n` guidance out of the error message body for the affected install/tap/unknown-command paths, so human output renders a single `Next step` block.
- Suppressed the generic `invalid_ref` remedy for known-tap install suggestions, since that message already includes the suggested commands.
- Made `writeError` require an explicit styler instead of carrying an untested default parameter.

Verification: `bun run check` passes locally: 863 tests, 100% line/function coverage.

dutabado (approved):

PR #62, opened 2026-05-11
[LOG-4738] Add untap alias

stevekrenzel:
## Summary
- Add `crew untap <name>` as a top-level alias for `crew tap remove <name>`
- Add dedicated `crew help untap` content and include the alias in the help overview
- Keep PRD, README, and homepage command reference in sync

## Test plan
- bun run check

Linear: https://linear.app/logic-app/issue/LOG-4738/add-untap-alias-for-tap-remove


stevekrenzel:
Addressed the review feedback in de666bd:

- Rendered `crew tap remove` as inline code in the site command table, matching the `crew taps` alias entry.
- Added `install` to the `untap` help `see also` list.
- Fixed the site formatting failure and verified both site checks (`bun run build && bun run typecheck && bun run lint`) and root `bun run check` locally.

dutabado (approved):

PR #63, opened 2026-05-11
[LOG-4740] Make search installed markers source-aware

stevekrenzel:
## Summary
- make `crew search` compute installed status from state source tap/path instead of skill name alone
- add `same_name_installed` to configured search hits so same-name conflicts can be distinguished without showing a green installed marker
- update PRD §16.6 and add the LOG-4740 regression test for removed/replaced taps

## Verification
- `bun run check`


stevekrenzel:
Addressed Claude's minor notes in 0b27b5a:

- Kept `SearchInstallStatus` private since callers do not import it directly.
- Changed the same-name human warning to lead with a shorter `same name installed elsewhere` note before the skill description, so the important warning survives better on narrow terminals.

Verified locally with `bun run typecheck && bun run lint` and the full `bun run check` gate.

dutabado (approved):
Looks like a small test failure that should be easy to fix, but LGTM otherwise

PR #66, opened 2026-05-12
[LOG-4746] Release v0.8.1

stevekrenzel:
### Added

- **Compound Engineering tap** — EveryInc's [Compound Engineering Plugin](https://github.com/EveryInc/compound-engineering-plugin) is now a curated known tap, adding 37 skills for spec writing, implementation planning, code review, debugging, and product workflows. Because the skills live in a non-standard subdirectory, the tap requires the `--recursive`-style subpath form — crew handles this automatically when you use the known-tap suggestion:
  ```sh
  crew tap add https://github.com/EveryInc/compound-engineering-plugin//plugins/compound-engineering/skills compound-engineering
  ```
  The tap also appears in `crew search` results and the skill catalog on the crew website.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [0.8.1] — 2026-05-11
+
+### Added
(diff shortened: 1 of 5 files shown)


jgarms (approved):
👍

PR #69, opened 2026-05-12
[LOG-4758] Use latest version in install snippet

stevekrenzel:
image

Fixing the hardcoded version string in the install example.

image

## Summary

- Replace the hardcoded homepage install snippet version with the latest release tag from `site/public/latest-version.json`.
- Strip the leading `v` for the terminal example so it matches `crew version` output.

## Test plan

- `cd site && bun run check`


dutabado (approved):
🤘

PR #70, opened 2026-05-18
[LOG-4857] Add Vercel analytics to site

stevekrenzel:
## Summary
The Crew site now loads Vercel Web Analytics from the root Next.js layout, so page visits are captured across the site once deployed on Vercel.

Linear: https://linear.app/logic-app/issue/LOG-4857/add-vercel-analytics-to-crew-site

## Test Plan
- `cd site && bun run check`

### DIFF
--- a/site/app/layout.tsx
+++ b/site/app/layout.tsx
@@ -1,3 +1,4 @@
+import { Analytics } from "@vercel/analytics/next";
 import type { Metadata } from "next";
 import { Inter, JetBrains_Mono } from "next/font/google";
 import type { ReactNode } from "react";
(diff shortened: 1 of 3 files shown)


site/app/layout.tsx
```diff
@@ -40,7 +41,10 @@ export const metadata: Metadata = {
 export default function RootLayout({ children }: { children: ReactNode }) {
   return (
     <html lang="en" className={`${inter.variable} ${jetbrainsMono.variable}`}>
-      <body>{children}</body>
+      <body>
+        {children}
+        <Analytics />
```
dutabado, line 46: does this handle any opt-out stuff we might need natively? or all analytics of the kind that that isn't a concern?

dutabado (approved):
one question but LGTM!

site/app/layout.tsx
stevekrenzel: Good question. @claude do you know how Vercel analyitcs works in this regard?

PR #71, opened 2026-05-20
[LOG-4892] fix: persist local update content hashes

stevekrenzel:
## Summary

Local path-sourced skills now stop reporting as updated after the first real content change. `crew update` persists the newly staged content hash back into state, so the next run compares against current source bytes instead of stale state.

The JSON update row shape stays unchanged: the new hash is carried only through internal update plumbing, while public update output continues to expose the same fields as before.

## Tests

- `bun run typecheck`
- `bun run lint`
- `bun test` — 947 pass, 100% line/function coverage

---

GPT-5


jgarms (approved):
LGTM

PR #73, opened 2026-05-22
[LOG-4923] Surface auth errors in release.sh instead of failing silently

losephjambert:
## Summary

`bun run release <bump>` exited after `==> Asking Claude to summarize changes…` with only `error: script "release" exited with code 1` — no indication of what went wrong. Root cause: `claude --bare` (used here for deterministic output) explicitly ignores OAuth/keychain auth and requires `ANTHROPIC_API_KEY` in the env, and the claude call had `2>/dev/null`, so the "Not logged in" diagnostic was swallowed.

## Changes (`scripts/release.sh` only)

- **Upfront `ANTHROPIC_API_KEY` check** in the safety-check block, with a message that names `--bare`'s auth contract so the next person doesn't have to rediscover it.
- **Drop `2>/dev/null`** on the `claude -p` invocation so future stderr surfaces normally.
- **Detect `{"is_error": true, "result": "…"}`** in the captured response and print `result` before the downstream `jq` would fail with a cryptic parse error. `claude` returns this envelope (with exit 0) for auth/budget/refusal failures.

No PRD, src, or site changes — script-only, no user-facing behavior shift beyond clearer errors.

## Test plan

- [ ] Unset `ANTHROPIC_API_KEY`, run `bun run release minor --dry-run`; expect the new upfront error.
- [ ] With a valid `ANTHROPIC_API_KEY`, run `bun run release minor --dry-run`; expect the existing PR-body/CHANGELOG preview to render.
- [ ] (Manual) Confirm `bash -n scripts/release.sh` passes (it does locally).

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -71,6 +71,18 @@ for cmd in claude gh jq bun; do
   fi
 done
 
+# `claude --bare` (used below for deterministic output) explicitly
+# bypasses OAuth and the keychain — it only accepts ANTHROPIC_API_KEY
+# or an apiKeyHelper via --settings. Catch this here so the failure
+# isn't opaque at the summary step.
+if [ -z "${ANTHROPIC_API_KEY:-}" ]; then
+  echo "error: ANTHROPIC_API_KEY is not set" >&2
+  echo "  This script invokes 'claude --bare' to summarize changes," >&2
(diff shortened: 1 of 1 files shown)


stevekrenzel (approved):
LGTM - good find

PR #75, opened 2026-05-26
[LOG-4941] Release v0.9.0

losephjambert:
### Added

- **Antigravity CLI is now a supported agent.** `crew` detects Antigravity CLI automatically and installs skills to `~/.gemini/antigravity-cli/skills/` for user-level installs, or the shared `<project>/.agents/skills/` path for project-level installs.

### Fixed

- **`crew update` no longer re-reports local skills as changed on every run.** After applying an update to a local path-sourced skill, `crew` now persists the new content hash so subsequent runs correctly see the skill as up-to-date.
- **The install snippet on the crew website always shows the current release version.** The homepage example was previously hardcoded; it now reads the latest release tag dynamically so copy-pasted commands stay correct after every release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## [0.9.0] — 2026-05-26
+
+### Added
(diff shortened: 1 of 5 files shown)


stevekrenzel (approved):
LGTM!

PR #76, opened 2026-05-29
[LOG-4981] style(site): refresh visual system and grid fidget

stevekrenzel:
Video comparing current site to new site:

https://github.com/user-attachments/assets/f6d1b988-0f00-40b0-b963-04220aa9490f

## Summary

The site now has a sharper, more distinctive visual system and a quieter white-first palette, with rounded treatment preserved where it reads as intentional copy or brand chrome.

The background grid is now an interactive fidget layer: hover tracks the current cell, background clicks send a soft wavy square ripple across the viewport, and the canvas handles its own grid drawing so the effect stays aligned while scrolling. Foreground content remains above the fidget layer, and real controls/media/code still do not trigger accidental waves.

Inline command text in the hero now reads as a compact code token instead of oversized monospace copy.

## Test Plan

- `bun run typecheck` in `site/`
- `bun run lint` in `site/`
- `bun run build` in `site/`

---

GPT-5

### DIFF
--- a/site/app/globals.css
+++ b/site/app/globals.css
@@ -11,17 +11,20 @@ body {
 }
 
 body {
-  background:
-    linear-gradient(color-mix(in oklab, var(--saffron) 7%, transparent) 1px, transparent 1px),
-    linear-gradient(90deg, color-mix(in oklab, var(--saffron) 6%, transparent) 1px, transparent 1px),
-    var(--bg);
-  background-size: 18px 18px;
+  background: var(--bg);
   color: var(--ink);
   font-family: var(--font-sans);
   font-size: 16px;
   line-height: 1.55;
   -webkit-font-smoothing: antialiased;
   font-feature-settings: "ss01", "cv11";
+}
+
+body > nav,
+body > main,
+body > footer {
+  position: relative;
+  z-index: 2;
 }
 
 a {
--- a/site/app/globals.css
+++ b/site/app/globals.css
@@ -23,6 +23,7 @@ body {
 body > nav,
 body > main,
 body > footer {
+  /* Keep page content above the visual-only GridFidget canvas. */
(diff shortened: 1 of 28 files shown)


stevekrenzel:
Addressed the review feedback in `74d46f0`.

Changes made:
- Moved shared grid helpers into `grid-fidget/util.ts` so `CELL_SIZE`, `snapToGrid`, and `colorWithAlpha` have one source of truth.
- Made `colorWithAlpha` parse hex and rgb/rgba values into `rgba(...)` instead of appending hex alpha bytes to arbitrary CSS values.
- Hid the decorative canvas from assistive tech via an `aria-hidden` wrapper. Biome rejects `aria-hidden` directly on `<canvas>`, so the wrapper preserves the intent without disabling the a11y rule.
- Removed the nonexistent `§16.6` docstring references and marked the fidget files as site-only/no CLI PRD section.
- Batched the baseline grid drawing into one path per axis instead of one `stroke()` per line.
- Added `[data-no-fidget]` as a future opt-out hook for overlays/foreground regions.
- Derived the ring fade span from `RING_OFFSETS` instead of a magic divisor.
- Added a comment documenting that page content intentionally stacks above the GridFidget canvas.
- Kept the permissive click behavior for plain text/whitespace intentionally: the requested UX was that anything that visually reads as background should fidget, while real controls/media/code remain excluded.
- Adjusted reduced-motion handling so the static grid still renders, while hover/wave motion is disabled.

Validated with:
- `bun run typecheck` in `site/`
- `bun run lint` in `site/`
- `bun run build` in `site/`

There are no unresolved review threads on the PR right now, so I’m replying here rather than resolving a thread.

jgarms (approved):
Love the grid fidget!

PR #80, opened 2026-06-14
[LOG-5085] feat: add Linux support

### LINKED ISSUE #79: Timeline for Linux support
Great tool idea, super useful! 
However, half my work is on a remote (linux) workstation - do you already have a timeline for linux support?

stevekrenzel:
## Summary

Homecrew can now run as a first-class Linux install: release assets, the public installer, self-update, and autoupdate all understand Linux arm64/x64 alongside macOS. Linux autoupdate uses a systemd user timer behind the same platform-neutral scheduler contract that launchd uses on macOS, so status, doctor, logging, and error handling stay consistent across platforms.

The public surface has been brought along with the behavior: the PRD, README, install script, release workflows, and website now describe Linux support and the new platform scheduler model. The site also includes the presentation refresh and grid background work already on this branch.

## Design Notes

| Area | What changed |
|---|---|
| Linux autoupdate | Adds a systemd user backend, shared scheduler dispatch, shared autoupdate log parsing, and platform-neutral `autoupdate_failure` handling. |
| Release/install | Builds and publishes Linux binaries, teaches the installer to select Linux assets, and supports `sha256sum` when `shasum` is unavailable. |
| Self-update | Selects Linux assets, verifies signed checksums before swapping binaries, caps downloaded asset sizes, and cleans failed temp downloads. |
| Status/doctor | Reports scheduler state instead of macOS-only agent state, while keeping `agent_loaded` as a deprecated JSON alias for compatibility. |
| Site/docs | Updates user-facing Linux copy and refreshes the site visual system, including the interactive grid background. |

## Test Plan

- `bun run check`
- 985 tests passed, 0 failed
- 100% line/function coverage

---

GPT-5

Closes #79

### DIFF
--- a/src/cli/main.ts
+++ b/src/cli/main.ts
@@ -11,6 +11,7 @@ import { CrewError } from "../core/errors.ts";
 import { crewHome } from "../core/paths.ts";
 import { maybeEmitUpdateNotice } from "../self-update/notice.ts";
 import { colorEnabled, makeStyler, type Styler, terminalWidth } from "../util/term.ts";
+import { nowIso } from "../util/time.ts";
 import { parseArgs } from "./args.ts";
 import { dispatch } from "./dispatch.ts";
 import { defaultStreams, type OutputStreams, writeError, writeSuccess } from "./output.ts";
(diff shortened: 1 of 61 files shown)


src/cli/main.ts
```diff
@@ -105,6 +118,11 @@ export function runCli(argv: readonly string[], options: RunCliOptions = {}): nu
       );
       exitCode = 4;
     }
+    // §10.2: scheduled updates append stderr to the autoupdate log. This
+    // line lives in the CLI catch path so crashes still update status.
+    if (process.env["CREW_AUTOUPDATE_LOG"] === "1") {
```
markgolazeski: Nit: Do we want this to also log in non-error scenarios?

markgolazeski (approved):
LGTM - One logging question but non-blocking.

src/cli/main.ts
stevekrenzel: Good call. Normal scheduled `crew update --quiet` completions were already logging from the update command, while crashes logged from the CLI catch path, but that split made the invariant harder to see. I centralized the autoupdate status-line emission in `runCli`, gated to `crew update` + `CREW_AUTOUPDATE_LOG=1`, so success and failure both log exactly once at the CLI boundary.

PR #82, opened 2026-06-15
[LOG-5092] Release v0.10.0

stevekrenzel:
## Summary

Release v0.10.0 publishes the Linux support work merged since v0.9.0. Homecrew now has Linux x64 and ARM64 release assets, Linux installer/self-update support, and systemd user timers for background autoupdates.

## Changed

- Add Linux as a supported Homecrew platform alongside macOS.
- Point the hosted latest-version metadata at the v0.10.0 release assets for macOS, Linux, and signed checksums.
- Bump package, implementation, and PRD versions to 0.10.0.
- Add the v0.10.0 changelog entry.
- Fix `scripts/release.sh` to handle Claude CLI JSON output when the result envelope is wrapped in an event array.

## Verification

- `bun run known-taps build && bun run known-taps check`
- `bun run check` (985 tests, 100% line/function coverage)
- `bash -n scripts/release.sh`
- Sampled the release helper jq parser against both direct-object and array-wrapped Claude responses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## [0.10.0] — 2026-06-15
+
+### Added
+- Linux is now a supported platform. Homecrew ships Linux x64 and ARM64 release assets, installs through the hosted installer, performs signed self-updates, and runs background autoupdates through systemd user timers.
+
+### Changed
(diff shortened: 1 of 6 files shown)


jgarms (approved):
🚢 🐧

PR #107, opened 2026-09-11
[LOG-5751] fix(uninstall): honor --dry-run instead of removing skills

stevekrenzel:
## What

`crew uninstall --dry-run <name>` was doing the full removal: agent directories deleted, markers rewritten, `state.json` written, auto taps garbage-collected. The flag was parsed and never consulted.

A dry run now runs every selector, `--agent` filter, `--prune` pass, and safety check exactly as a real run would (`not_installed_here`, `untracked_directory`, `inconsistent_marker` still abort), but writes nothing:

- the per-agent step reports instead of removing (no `rmrf`, no marker rewrite);
- the command skips `writeState` and auto-tap GC;
- the state lock is **not** taken — acquiring it creates `state.json` as a side effect, so a dry run against a fresh `CREW_HOME` used to leave a state file behind;
- output is tagged `(dry run)`, with `would remove from N agents` and a `Would prune N dependencies` heading;
- `--json` carries `dry_run: true|false`, matching install's shape.

Selector walking lives in one `planUninstall` function used by both paths, so the preview and the real removal can't disagree about what happens.

## PRD

- §7.4 gains a `--dry-run` paragraph. It states removed, retained, and aborted as three separate output obligations, and records that the state lock is not taken.
- §18.3 gains C-UNINST-19 and C-UNINST-19a (fresh home is not created).

## Help / surfaces

- `crew help uninstall` lists `--dry-run` and gets a `--dry-run --prune` example.
- README already says every command accepts `--dry-run`; no change needed. Site command reference doesn't enumerate global flags.

## Tests

`tests/e2e/uninstall-dry-run.test.ts` (9 tests, all e2e through `runCli` against real tmp dirs):

- install dirs, markers, and `state.json` compared as **raw bytes** before and after a dry run;
- a dry run against a home with no `state.json` creates neither the file, the lock, nor the directory;
- `--prune --dry-run` lists the orphan under `Would prune` and leaves it;
- `--json` `dry_run` true on a dry run, false on a real run;
- `--agent --dry-run` on a path-shared dest leaves the marker's agent list intact;
- an untracked dest aborts with exit 1;
- `--force --dry-run` leaves an untracked dest (directory, sentinel file, and SKILL.md all present) and, separately, leaves a mismatched marker byte-identical;
- the backing auto tap is not garbage-collected.

Both new guards fail against the pre-fix code.

`bun run check` green: 994 pass, 100% line + function coverage.


stevekrenzel:
## `/review`

Verdict: **not ready — 0 blockers, 2 majors, 5 minors, 1 nit**

### Major findings

1. **Dry-run can still write `state.json`** — `src/commands/uninstall/index.ts:56`  
   The dry-run path enters `withStateLock` before returning. `acquireStateLock` creates the state directory and touches a missing `state.json`, so a dry run against a fresh `CREW_HOME` violates the new “writes nothing / state is not written” contract and the repo convention that read-only commands do not lock. This was independently flagged by the clarity, error-handling, and architecture lenses.  
   **Fix:** run the shared uninstall planning path without `withStateLock` for dry runs, reserving the lock and persistence/GC for real uninstalls. Add coverage with an absent state file.

2. **The forced untracked-directory dry-run branch is not tested** — `tests/e2e/uninstall-dry-run.test.ts:137`  
   The test first checks an untracked directory without `--force`, then writes a mismatched marker before the forced call. The forced call therefore exercises `inconsistentMarker`, never `untrackedInstall`. A regression that deletes an untracked directory during `--force --dry-run` would pass.  
   **Fix:** leave `.crew.json` absent, run `uninstall --force --dry-run foo`, and assert the directory plus a sentinel file remain.

### Minor findings

- `src/agents/uninstall.ts:25`: make `dryRun` required. All CLI call sites provide a boolean, while omission currently silently selects the destructive path.
- `src/agents/uninstall.ts:30-38`: the outcome docs say `removed`/`detached` and the function says it removes ownership even when dry-run only predicts those actions. Document the projected meaning or model planned actions separately.
- `src/commands/uninstall/render.ts:46`: `Pruned 1 dependency (dry run)` uses past tense for an action that did not happen. Prefer `Would prune 1 dependency` and update the assertion.
- `tests/e2e/uninstall-dry-run.test.ts:143`: capture the unforced safety-check result and assert exit code 1; checking only the text does not verify that it aborts.
- `tests/e2e/uninstall-dry-run.test.ts:120,131`: use the canonical `Marker` type rather than the mutable ad hoc `{ agents: string[] }`; compare a copied agents array before sorting.

### Nit

- `PRD.md:534`: the sentence puts agents that `--agent` would leave in place inside “what would have been removed.” State removed and retained agents as separate output obligations.

The PR description also says the test proves directories, markers, and state are byte-identical, but the test parses/reserializes marker and state and checks only `SKILL.md` existence. Either compare raw bytes/full contents or narrow that claim.

### Reviewer coverage

- clarity: 4 findings
- naming: 1 finding
- type safety: 1 finding
- database: no findings
- testing: 2 findings
- security: no findings
- error handling: 1 finding
- observability: no findings
- performance: no findings
- concurrency: no findings
- architecture/conventions: 1 finding

Deduplicated overlaps are counted once in the verdict. CI is green, but the majors above prevent approval.

stevekrenzel:
All 2 majors, 5 minors, the nit, and the description issue are addressed in `1a1f4f8`. Nothing pushed back on.

### Majors

**1. Dry run could still write `state.json`.** Confirmed before fixing: against a `CREW_HOME` that didn't exist yet, a dry run created both the directory and `state.json`, because `acquireStateLock` calls `ensureDir` + `touch` when the file is absent. Selector walking moved into a `planUninstall` function in `src/commands/uninstall/index.ts` that takes no lock; a dry run calls it directly, a real run calls it inside `withStateLock` and then does `writeState` + auto-tap GC. Both paths share the one decision function, so the preview and the real removal can't drift. New test `C-UNINST-19a` asserts that after a dry run against an untouched home, `state.json`, the `.lock`, and the home directory itself all still don't exist. It fails on the previous commit.

**2. The forced untracked-directory branch was never exercised.** Correct — the mismatched marker written on line 149 meant the forced call landed in `inconsistentMarker`, so `untrackedInstall`'s `rmrf` was never covered by a dry-run test. Split into three tests in `tests/e2e/uninstall-dry-run.test.ts`: an untracked dest aborting without `--force`, `--force --dry-run` over an untracked dest (no marker, plus a `sentinel.txt` that must survive), and `--force --dry-run` over a mismatched marker. The untracked case now asserts the directory, the sentinel, and `SKILL.md` are all still present.

### Minors

- **`dryRun` now required** (`src/agents/uninstall.ts:25`). Worth noting: making it required surfaced seven existing call sites in `tests/misc/` that were relying on the destructive default, which is exactly the hazard you flagged. All now pass `dryRun: false` explicitly.
- **Outcome docs** (`src/agents/uninstall.ts:30-38`) now say that under `dryRun` the variants describe what *would* happen — checks have run and the decision is final, but no bytes moved — and that rendering the tense is the caller's job. I kept one shape rather than splitting planned from actual outcomes, since the decision is identical and a second type would duplicate the discriminants.
- **`Pruned` → `Would prune`** (`src/commands/uninstall/render.ts:46`). The verb now carries the tense, so I dropped the now-redundant `(dry run)` tag from that one heading. `tests/e2e/uninstall-prune.test.ts` still asserts `Pruned` for real runs and is unaffected.
- **Abort exit code asserted** — the untracked-without-force test now captures the code and asserts `1`, not just the message text.
- **Canonical `Marker` type** used for both marker reads, with `[...(marker.agents ?? [])].sort()` so the assertion doesn't sort the underlying array in place.

### Nit

**PRD §534** now states three separate output obligations — what would be removed (including `--prune` orphans), what would be retained (where `--agent` leaves the skill installed elsewhere), and what would abort — instead of folding retained agents into the removed list. The same paragraph records the no-lock guarantee, and C-UNINST-19a covers it in §18.3.

### PR description

You were right that it overclaimed. The first test parsed and reserialized before comparing, so a write that changed key order or formatting would have slipped through. It now compares raw `Buffer`s for `state.json`, the marker, and `SKILL.md`. Description updated to match the actual test set.

`bun run check` green: 994 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean.

stevekrenzel:
## Updated `/review`

Verdict: **not approved — 1 blocker, 1 major, 1 minor**

### Blocker

- **Dry-run still omits the agents retained by a partial uninstall** (`PRD.md:541`, `src/commands/uninstall/core.ts:19`, `src/commands/uninstall/render.ts:64`). The updated PRD requires preview output to name the agents that `--agent` leaves installed. Records retain only `partial: true`, human output says only “kept elsewhere,” and JSON cannot report those agents. Add `remainingAgents` to `UninstallRecord`, render/serialize it, and assert the exact retained agents in the partial dry-run test.

### Major

- **The new test file violates the repository’s hard file-size cap** (`tests/e2e/uninstall-dry-run.test.ts:245`). It is 245 lines; split the scenarios into focused sibling files or extract shared setup so every file remains below 200 lines.

### Minor

- **The byte-integrity test covers only one physical target** (`tests/e2e/uninstall-dry-run.test.ts:58`). It snapshots Claude’s destination and state, but not Codex’s destination/marker or config and the backing auto-tap clone. Snapshot both targets and config; in the auto-tap scenario also assert the clone remains present.

All eleven lenses reran against updated head `1a1f4f8`; CI is green. I did not approve because a blocker and major remain.

stevekrenzel:
All three findings addressed. Nothing declined.

### Blocker — dry-run omitted the agents a partial uninstall retains

Reproduced first. `crew uninstall --dry-run --agent codex foo`, with the skill installed in both agents, printed:

```
  foo (kept elsewhere)
    [ok] codex

would remove from 1 agent
```

and `--json` carried `"partial": true` with no agent names anywhere. `claude-code` — where the skill actually survives — appeared in neither. `removeOne` computed `remainingAgents` locally (`src/commands/uninstall/core.ts`) and threw the names away, keeping only the boolean.

`UninstallRecord` now carries `remainingAgents`, populated as the deduplicated union across entries (a skill can hold entries at several scopes, each retaining a different subset). `src/commands/uninstall/render.ts` lists each retained agent on its own row, tensed by mode — `(would keep)` on a preview, `(kept)` on a real run — and `--json` serializes the field. Output now reads:

```
  foo (kept elsewhere)
    [ok] codex
    - claude-code (would keep)
```

The real run reports the same field the same way, so a preview can't disagree with what follows. `tests/e2e/uninstall-dry-run/reporting.test.ts` asserts the exact retained list in both human and JSON output, and asserts preview and real runs agree (C-UNINST-19b).

PRD §7.4's retained-agents bullet now says they must be *named, not merely counted*, records the `remainingAgents` JSON field, and states the preview/real parity requirement. New criterion C-UNINST-19b in §18.3.

### Major — test file over the 200-line cap

It was 245 lines. Split into `tests/e2e/uninstall-dry-run/`, grouped by what each file proves:

| File | Lines | Covers |
|---|---|---|
| `helpers.ts` | 63 | adapter redirection, shared fixture |
| `integrity.test.ts` | 105 | writes-nothing, fresh-home, auto-tap GC |
| `reporting.test.ts` | 119 | prune preview, `--json`, retained agents |
| `safety.test.ts` | 78 | abort + `--force` branches |

### Minor — byte-integrity covered only one target

Correct; it snapshotted Claude Code's destination and `state.json` only. `snapshot()` in `integrity.test.ts` now captures raw bytes for **both** agents' `SKILL.md` and `.crew.json`, plus `state.json` and `config.yaml`, and compares every one after the run.

The auto-tap test previously asserted only that the config *row* survived, which would pass even if the clone directory had been deleted. It now resolves the auto tap's clone path, asserts the directory still exists, and compares `config.yaml` bytes.

### Pre-fix verification

Restored `core.ts` and `render.ts` to the previous commit, keeping the new tests: **both C-UNINST-19b tests fail**, the other two in that file pass. Restored afterwards and confirmed green.

The integrity and safety tests pass either way — they were already covering their branches, and the minor widened what they inspect rather than catching a live bug. Saying so explicitly so the count isn't read as four caught regressions.

### On the gate

`typecheck` and `lint` are clean. `bun test` is green for everything this PR touches: the 10 tests in `tests/e2e/uninstall-dry-run/`, and 201 pass / 0 fail across every uninstall suite (`uninstall-dry-run/`, `uninstall-prune`, `uninstall-scope`, `tests/misc/`).

Full-suite runs on this machine are currently unreliable and I don't want to report a number I can't stand behind. Four consecutive runs gave 2, 5, 8, then 11 failures, with wall time climbing 111s → 161s → 192s → 704s. The failing set differed every run, never included anything under `uninstall-`, and every one I checked passed in isolation (`tap-install` + `install-known-tap`: 24 pass / 0 fail). One run failed before reaching the tests at all, in the `known-taps check` pre-step:

```
couldn't clone `https://github.com/anthropics/knowledge-work-plugins.git`
error: RPC failed; curl 56 Recv failure: Connection reset by peer
```

That's contention from other agents sharing this machine, not a regression here — and the clone in that pre-step is the same network dependency #131 removes from the suite. Worth a clean CI run to confirm, which this PR will get once #134 lands and stacked PRs actually run tests.

stevekrenzel:

**Verdict:** 0 blockers, 1 major, 3 minors, 1 nit. **Not approved.**

Reviewed `72e6acf..4a0b678` through all 11 fresh `/review` lenses. CI is green at this head.

### Major

1. **A safety failure during `--prune` can spin forever on the same orphan** — `src/commands/uninstall/index.ts:141-145`  
   `removeOne` now correctly retains an orphan whose uninstall fails a safety check. That leaves `current` unchanged, so `findOrphan(current)` selects the same orphan again on every iteration. `crew uninstall --dry-run --prune ...` hangs; a real run also holds the state lock indefinitely. This was independently flagged by the database/state and error-handling lenses.  
   **Fix:** track orphans already attempted (or otherwise stop when removal makes no progress), retain the failure record, continue to any other orphan, and propagate the prune failure into exit code 1. Add a regression case with an orphan whose marker is missing or inconsistent.

### Minors

- `tests/e2e/uninstall-dry-run/integrity.test.ts:3-6`: the suite docstring says every artifact, including the backing auto-tap clone, is compared as raw bytes. The clone test checks only that the directory survives. Narrow the claim or snapshot the clone contents too.
- The PR description points to the removed `tests/e2e/uninstall-dry-run.test.ts` and says there are nine tests; the current suite is split under `tests/e2e/uninstall-dry-run/` and contains twelve tests. Update the description.
- The PR description says the site command reference does not enumerate global flags, but `site/components/sections/Commands.tsx` explicitly lists `--dry-run`. Correct that explanation for the site decision.

### Nit

- `src/commands/uninstall/index.ts:75,87`: `UninstallPlan` / `planUninstall` sound side-effect-free even though the real-run path performs removals inside the function. An execution-neutral name such as `UninstallResult` / `runUninstall` would make the write boundary clearer.

### Reviewer coverage

- clarity: 3 findings
- naming: 1 finding
- type safety: no findings
- database/state: 1 finding
- testing: no findings; CI `check` passed at `4a0b678`
- security: no findings
- error handling: 1 finding (deduplicated with database/state)
- observability: no findings
- performance: no findings
- concurrency: no findings

No approval attempted because the major is a real command hang and lock-liveness failure.

stevekrenzel:
Round three addressed in `a15bb80`. Nothing declined — both majors reproduced first.

**Major 1 — the update check writes during a dry run.** Reproduced with a stubbed release fetcher on a fresh home: `uninstall --dry-run --force ghost` with `stderrIsTty: true` made **1 fetch attempt** and left `version-check.json` on disk. Every existing write-free test uses captured streams, which are never a TTY, so `isSuppressed` returned early and the stale branch — the one that fetches and calls `writeVersionCheck` — was never reached by any of them.

Fixed centrally rather than at this command's call site. `NoticeContext` gains `dryRun`, and `isSuppressed` treats it like the other suppression conditions, so the 24h check and its write are skipped for **any** dry run. Fixing it in `maybeEmitUpdateNotice` means the sibling dry-run PRs inherit it rather than each needing the same patch: **#109** (`update --dry-run`), **#112** (agents/autoupdate/cache/doctor), **#119** (tap), **#122** (`outdated`), **#130**. Whichever merges first carries the fix for all of them; none should add its own.

PRD §10.4's suppression list now includes the condition, with the reason, so the spec and code agree.

**Major 2 — a safety-failed agent was omitted from `remainingAgents`.** Reproduced: with codex's destination untracked and `--dry-run --agent codex foo`, the JSON reported `failures: [{agent: "codex", error: {code: "untracked_directory"}}]` alongside `remainingAgents: ["claude-code"]`. Codex's bytes stay on disk, so the preview claimed the skill was gone from an agent that still has it.

`removeFromAgents` now returns the set of agents whose ownership actually came off, and retention is computed from that outcome rather than from the requested filter. Post-fix the same run reports `["claude-code", "codex"]`. In human output an aborted agent appears under its failure line, which already carries the reason, instead of being listed twice.

§7.4 and C-UNINST-19b now state that retention is determined by outcome: an agent is retained when a filter excludes it **or** when its removal aborts, since its bytes remain either way.

**Minor 1 — inconsistent partial records.** `partial` and `remainingAgents` are now one co-dependent pair via a discriminated union, so a record cannot assert a partial removal without naming who kept the skill. Verified the union bites: a record with `partial: true` and no agent list fails to compile (`TS2322`). Both fields stay top-level in JSON, since `remainingAgents` is the field name §7.4 and C-UNINST-19b pin for the wire format — changing that would be a breaking contract change this review did not ask for.

**Minor 2 — duplicated test type.** `UninstallJson` now derives its record shape from `UninstallRecord` with `Pick`, so schema drift is a compile error rather than a test asserting a shape the command no longer emits.

**Pre-fix verification.** Stashed `src/` only and restored the tests over the old source: exactly the two new tests fail, one per major, and the ten pre-existing tests still pass.

```
(fail) C-UNINST-19a a dry run on a TTY does not write the version-check cache
(fail) C-UNINST-19b an agent whose removal aborts is still reported as retained
 10 pass, 2 fail
```

**Loop termination**, per the standing directive after the earlier runaway in this area: the only loops I touched are bounded `for...of` passes over fixed collections (`groups.values()`, `group`, `remainingAgents`, `failures`) with no mutation of the iterated sequence. `pruneOrphans` and its `attempted` guard are untouched. I re-read each before running anything against it.

**Results.** Targeted suites only, given the known non-determinism on branches without #131: `uninstall-dry-run/` + `uninstall-prune` + `self-update/` + `commands/` — **125 pass, 0 fail**. Typecheck and Biome clean; every touched file under the 200-line cap.

stevekrenzel:
Rebased onto `main` after the cap-compliance refactors (#135–#143) merged. This PR's only edits to the files that were split are the new required `dryRun: false` on option literals; each was re-placed where the call now lives, driven by `tsc` so none could be missed: five in `tests/misc/closeout/adapters.test.ts` and two in `tests/misc/closeout/uninstall-edges.test.ts` (formerly `gaps` / `coverage-closeout` / `final-gaps`), and one in the notice harness, now `tests/self-update/notice/helpers.ts`. No other changes. Full suite 1104 pass, 0 fail, 100% coverage; lint incl. the file-size gate and typecheck clean.

PR #108, opened 2026-09-11
[LOG-5752] feat(cli): make --verbose emit progress details to stderr

stevekrenzel:
## What

`--verbose` was accepted on every command (PRD §5.2) but nothing read it. This PR makes it emit one progress line per subprocess / I/O step on stderr:

- every `git …` invocation (with cwd), plus `launchctl` / `systemctl --user` calls
- tap refresh (`refreshing tap <name> from <url>`)
- store staging or reuse (`staging <name>@<sha8> into …`)
- per-agent install / uninstall with the destination path

Lines are prefixed `crew: ` and go to **stderr only**, so `--verbose --json` still leaves a clean JSON payload on stdout.

## How

- New `src/util/progress.ts`: `setProgressSink()` / `progress()`. Low-level modules call `progress(line)`; with no sink installed it is a no-op, so commands stay pure functions.
- `src/cli/main.ts` installs a sink for the run when `--verbose` is set and clears it after dispatch on every path (success or error), so it can never leak into a later `runCli` call.
- Calls added in `src/git/exec.ts`, `src/commands/tap/refresh.ts`, `src/sources/store.ts`, `src/agents/install.ts`, `src/agents/uninstall.ts`, `src/autoupdate/launchd.ts`, `src/autoupdate/systemd.ts`.

## PRD / docs

- §5.2: `--verbose` description now says what is emitted and that it is stderr-only.
- §18.3: new criterion **C-CLI-06a**.
- Help: overview gains a short GLOBAL FLAGS block; `install` and `update` pages list `--verbose`.
- README already described the flag correctly; no change needed.

## Tests

`tests/e2e/verbose/` (real `file://` repos, redirected adapter) — 14 tests
across two files:

`progress.test.ts` (5) — what reaches which stream:
- install `--verbose` -> stderr has clone / staging / install lines, stdout has none
- `--verbose --json` -> stdout parses as JSON, stderr has progress
- uninstall / reinstall / update -> removal, store-reuse, tap-refresh and fetch lines
- a nested `runCli` neither borrows the outer sink nor clears it. The inner
  command is an install: one that emits no progress could not fail the
  assertion however the sink behaved.
- a failed `--verbose` run followed by a non-verbose run -> nothing emitted

`safety.test.ts` (9) — what must never reach a stream intact:
- credentials in a direct install URL, a configured tap URL, `--json` details,
  and a `?token=` query parameter, in both output modes
- a credential interpolated into a **human** error message
- an unlisted secret parameter (`client_secret`), covering the inverted allow-list
- control characters in a path source and in a configured tap name
- a newline in a tap name cannot forge a line inside a multi-line error

`bun run check` green: 989 tests, 100% line + function coverage, Biome and tsc clean.


stevekrenzel:
## `/review`

Verdict: **not ready — 1 blocker, 1 major, 3 minors, 1 nit**

### Blocker

**Verbose output leaks credentials from supported Git URLs** — `src/git/exec.ts:40`, `src/commands/tap/refresh.ts:46`

Both new progress paths write raw Git arguments or `tap.url` to stderr. Crew accepts HTTPS remotes containing userinfo, such as `https://oauth2:<token>@host/org/repo.git`, so `install --verbose`, `update --verbose`, and tap refresh can expose a usable token in terminal captures or CI logs. Raw user-controlled arguments can also inject terminal control characters or forged progress lines. Security, observability, and error-handling independently flagged this.

**Fix:** add one shared safe renderer for verbose command/URL output. Redact URL userinfo and any supported sensitive query parameters, escape control characters, and use it for both Git arguments and tap URLs. Add direct-install and configured-tap tests with dummy credentials and assert the secret never appears.

### Major

**The module-global progress sink is not invocation-scoped** — `src/cli/main.ts:101,129`, `src/util/progress.ts:20`

A nested `runCli` call can inherit or replace the outer call’s sink, and every inner call unconditionally clears it. For example, a custom prompt or stream callback can invoke non-verbose run B during verbose run A: B writes progress to A’s stderr, clears A’s sink, and A then loses later progress.

**Fix:** make sink installation return the previous sink and restore it in a `finally` block, installing an explicit null sink for non-verbose invocations; passing progress through `CommandContext` would also remove the shared-state hazard. Add a nested-invocation regression test.

### Minors

- `src/agents/uninstall.ts:40`: `removing …` is emitted before existence and marker validation. Missing, untracked, or inconsistent installs report work that never starts. Emit it immediately before the mutation, including force-removal branches.
- `src/autoupdate/launchd.ts:133` and `src/autoupdate/systemd.ts:179`: the new scheduler progress paths have no verbose-mode assertions. Add CLI tests using the existing platform and runner seams for both rendered commands.
- `tests/e2e/verbose.test.ts:69`: assert success for setup install, uninstall, reinstall, and update. Their progress lines can be emitted before later failure, so the current text-only assertions do not prove successful execution.

### Nit

- `crew help uninstall` does not list `--verbose`, although uninstall now emits a per-agent progress line and install/update help list the flag.

### Reviewer coverage

- clarity: no findings
- naming: no findings
- type safety: no findings
- database: no findings
- testing: 2 findings
- security: 1 finding
- observability: 3 findings
- performance: no findings
- concurrency: 1 finding

Deduplicated overlaps are counted once in the verdict. CI is green, but the blocker and major prevent approval.

stevekrenzel:
All eight findings addressed in `c858351`. Nothing pushed back on — every finding reproduced.

## Blocker — credential leak in verbose output

Confirmed before fixing: `parseRef("https://oauth2:ghp_SECRET@github.com/acme/skills.git")` is accepted and the userinfo is preserved verbatim in `source.url`, so `runGit`'s `args.join(" ")` put a usable token on stderr.

New `src/util/redact.ts` is the single safe renderer, used by every verbose path (`src/git/exec.ts`, `src/commands/tap/refresh.ts`, `src/sources/store.ts`, `src/agents/install.ts`, `src/agents/uninstall.ts`, `src/autoupdate/launchd.ts`, `src/autoupdate/systemd.ts`):

- **URL passwords** → `https://oauth2:***@host/...`
- **Lone https userinfo tokens** (the `https://<token>@host` shape) → `https://***@host/...`
- **Sensitive query parameters** (`token`, `access_token`, `private_token`, `api_key`, `apikey`, `password`) → `***`, leaving other parameters like `ref=main` intact
- **Control characters** → escaped to a visible `\xNN`

One deliberate exception: **SSH userinfo is not redacted.** `git@host` and `ssh://git@host/…` carry a username, not a secret; masking it would cost readability and protect nothing. Documented in the module docstring and PRD §5.2.

**Adjacent fix, worth a look since it widens the diff.** The same leak existed *outside* verbose mode: `cloneRepo` interpolated the raw URL into its `source_unreachable` message, so a failed clone printed the token on every run. I fixed that too — leaving it would have made the blocker fix hollow. Relatedly, while testing control characters I found that a hostile path reached `no_skills_found` with a raw ESC, so escaping is now applied centrally in `writeMessageBlock` (`src/cli/output.ts`), which covers every human error message including git's passthrough stderr.

## Major — module-global progress sink

`setProgressSink` now returns the sink it replaced, and `src/cli/main.ts` restores it in a `finally`, installing an explicit null sink for non-verbose runs. I chose save/restore over threading progress through `CommandContext`: the sink is consumed by low-level modules (`git/exec.ts`, `sources/store.ts`) that have no access to a command context, so threading it would have meant passing it through every intermediate call, whereas save/restore fixes the hazard at the one place that owns invocation lifetime.

## Minors

- **`src/agents/uninstall.ts`** — the `removing …` line moved out of the entry point and into a `progressRemoving` helper called immediately before each of the four real mutations (untracked force-remove, inconsistent-marker force-remove, last-owner removal, and marker rewrite). Missing, untracked, and inconsistent installs no longer announce work that never starts.
- **Scheduler coverage** — new `tests/autoupdate/verbose.test.ts` drives `autoupdate enable --verbose` through both platform seams and asserts the rendered `launchctl` and `systemctl --user` lines.
- **`tests/e2e/verbose.test.ts:69`** — the setup install, uninstall, reinstall, and update steps now assert exit code `0` rather than matching text alone.

## Nit

`crew help uninstall` lists `--verbose`.

## Do the new tests actually fail pre-fix?

I verified by stashing the source changes and restoring only the tests:

| Test | Pre-fix |
|---|---|
| credential in a direct install url | **fails** |
| credential in a configured tap url | **fails** |
| control characters in a path source | **fails** |
| nested `runCli` sink isolation | **fails** |

The nesting test needed a correction to be meaningful. My first version triggered the nested call from a **stdout** write, which happens *after* all of install's progress lines — so it passed even against the buggy sink. It now nests on the **first progress line on stderr**, which is the only window where a stolen or cleared sink is observable. Against the old sink code it fails with the outer run's stderr truncated to just the `git clone` line, exactly the described bug.

## Gate

`bun run check` green: **1008 pass, 0 fail, 100% line and function coverage**, typecheck and Biome clean. All `src/` files under the 200-line cap.

PRD §5.2 documents the redaction rule; new criterion **C-CLI-06b**.

stevekrenzel:
## Updated `/review`

Verdict: **not approved — 1 blocker, 3 majors**

### Blocker

- **Failed-clone errors still leak credential-bearing URLs** (`src/git/repo.ts:35`, `src/cli/output.ts:64`). The displayed URL is redacted, but `details.url` remains raw in JSON, and Git stderr can repeat the raw URL including query parameters. Redact URL substrings in diagnostics and redact or omit credential-bearing structured fields. Add failed `--verbose --json` cases for password and `?token=` URLs.

### Majors

- **Configured tap names can inject terminal controls into verbose output** (`src/commands/tap/refresh.ts:47`, `src/config/taps.ts:18`). Config accepts any nonempty string, which is interpolated directly into a progress line. Sanitize at the progress sink boundary or render names with the shared safe renderer; test CR/LF/ESC values.
- **Message output splits untrusted newlines before sanitizing** (`src/cli/output.ts:95`). An embedded user value can become an actual additional terminal line instead of escaped `\x0a`, violating the new §5.2 guarantee. Sanitize interpolated values before structural layout, or separate trusted layout newlines from untrusted text.
- **Two changed files exceed the hard file-size cap**: `src/git/repo.ts` and `tests/e2e/verbose.test.ts` are each 203/213 lines. Split them below 200 lines as required by AGENTS.md.

All eleven lenses reran against updated head `c858351`; CI is green. I did not approve because a blocker and majors remain.

stevekrenzel:
All four round-two findings addressed in `b9165c6`. Nothing declined — every one reproduced first.

## Blocker — failed-clone errors still leaked credentials

Reproduced before fixing, and it was worse than one channel. Against `c858351`, installing from `https://oauth2:<secret>@127.0.0.1:1/a/b.git`:

| Channel | Leaked |
|---|---|
| `details.url` in `--json` | **yes** — raw, both secret shapes |
| git's stderr, quoted into the message | **yes** — `fatal: unable to access 'https://…?token=<secret>/'` |
| our own rendered URL | no — round one fixed this |

The `?token=` case is the instructive one: the URL *we* render was already redacted, but git echoes the remote back in its own diagnostic, and we pass that through verbatim. It leaked in **human mode too**, not just `--json`.

Three changes in `src/util/redact.ts` and its call sites:

- **`redactText`** scans free text for URL-shaped substrings. `redactUrl` only handled a string that is entirely one URL; git's stderr is prose with a URL embedded in it, so it needed a scanner rather than a parser. Applied wherever git stderr is quoted — `cloneRepo`, `fetchAndCheckout`, and `checkoutSha` (the last wasn't reported, but it quotes stderr the same way, and leaving one sibling unguarded makes the fix hollow).
- **`redactDetails`** masks values in the structured payload. §13 makes `details` a stable machine contract, so keys are preserved and only values are redacted.
- **`writeError`** applies both at the JSON boundary.

New tests assert the secret appears in **no** stream, in both output modes, for a password URL and a `?token=` URL.

## Major — tap names could inject terminal controls

Reproduced: a tap named `evil\x1b[2Kcrew: forged\rname` put a raw ESC and CR on stderr, with the text `crew: forged` inside the progress line. A tap name is any non-empty string per §6.1, carries no credential, and so received none of the URL-specific handling — while being interpolated straight into the line.

Fixed at the **progress sink boundary** (`src/util/progress.ts`), your preferred option: `progress()` sanitizes its own argument, so every present and future call site is covered rather than each one remembering to.

**Interaction with #128, as requested:** that PR deliberately does *not* enforce the full slug pattern on tap names at config load, because auto-tap names derive from directory basenames and rejecting them would break existing configs on upgrade. This fix doesn't contradict that — nothing is rejected at load; the escaping happens at render.

## Major — untrusted newlines split before sanitizing

Reproduced: a path containing `\n` produced a real second terminal line beginning `Error: forged line here`, reading exactly like a second crew error.

The fix has to happen where the distinction between *layout* and *untrusted content* is still known, which is not the renderer — by then both are just `\n`. So `CrewError`'s constructor escapes its message, and the small number of genuinely multi-line errors opt in with a new `multiline: true` parameter. Exactly one caller needed it (`ambiguityError`, which composes from trusted literals); its output is unchanged and covered by the existing namespace and known-tap suites.

I first tried restructuring `writeMessageBlock` and stopped when I realized it was the same behavior rearranged — it would have looked like a fix while changing nothing.

One existing assertion encoded the old behavior:

```
// The embedded newline survives as block structure…
```

That is the bug, so the assertion is now that the newline is escaped to `\x0a` and no line begins with the forged text.

## Major — file-size cap

- `src/git/repo.ts` 206 → **143**, extracting `resolveRef` / `classifyRef` into `src/git/refs.ts`.
- `tests/e2e/verbose.test.ts` 213 → `tests/e2e/verbose/`: `helpers.ts` (39), `progress.test.ts` (119), `safety.test.ts` (158).

**Expected merge with #129:** that PR splits `repo.ts` the same way and to the same boundary — I checked its `refs.ts` and matched the file name, the exported functions, and the docstring rationale, so the two should reconcile as a duplicate rather than a conflict. #129 additionally extracts `export.ts` and `tree-safety.ts`, which are disjoint from this change.

## Do the new tests fail pre-fix?

Stashed `src/` only, kept the tests: **4 of 11 fail**, one per finding.

```
(fail) a failed clone keeps a password out of --json details
(fail) a failed clone keeps a token query parameter out of every stream
(fail) control characters in a path source are escaped, not echoed raw
(fail) a configured tap name cannot forge a progress line
```

The other seven are pre-existing and pass either way — they are the round-one coverage, not evidence of anything caught here.

## Gate

`bun run check` green on this run: **1011 pass, 0 fail**, 100% line and function coverage, typecheck and Biome clean.

Worth flagging on test reliability, since it affects how you read any full-suite number in this stack: the suite is **non-deterministic without #131**. Measured on a quiet machine, `origin/main` itself fails 3/985, and branches here fail 13–18 with a *different set each run*, all clustered in tests that clone from remotes — the default `core` tap URL is `github.com`, so any test not explicitly discarding it attempts a live clone. `tests/e2e/tap.test.ts` passes 41/41 alone and fails in a full run. On `fix/no-network-in-tests` the suite is 992 pass / 0 fail in 50s versus 310s+ here. My run above happened to come up clean; I would not treat that as proof on its own, which is why the pre-fix/post-fix comparison above is per-test rather than a headline count.

## PRD

§5.2 now states the redaction guarantee covers every channel a source URL can reach — rendered message, diagnostics quoted from `git`, and `--json` `details` — and that newline escaping applies to interpolated values while composed layout breaks are preserved. C-CLI-06b updated to match.

stevekrenzel:

**Verdict:** 0 blockers, 1 major, 3 minors, 1 nit. **Not approved.**

Reviewed `72e6acf..e2d900a` through all 11 `/review` lenses. The head remained unchanged during review, and required CI checks are green.

### Major

1. **Malformed credential-bearing URLs fail open in the redactor** — `src/util/redact.ts:90`

   `redactStandardUrl` returns the original token when `new URL()` throws. Config parsing accepts any nonempty git URL, so a value such as `https://user:SECRET@127.0.0.1:99999/org/repo.git` reaches verbose Git/tap progress unchanged. The same fallback is used by Git argument and error rendering. This violates §5.2’s credential guarantee and can publish a usable secret in terminal or CI output.

   **Fix:** fail closed when parsing a scheme-bearing URL fails: conservatively redact userinfo and query values, or render a fully redacted placeholder. Add a malformed-URL regression covering verbose output and failed-clone human/JSON errors.

### Minors

1. **The interactive ambiguity menu still writes raw configured/user values** — `src/commands/install/ambiguity-prompt.ts:62-65`, `src/install/resolve-ref/format.ts:40-47`

   The noninteractive ambiguity error escapes tap, namespace, skill, and argv fragments, but `shortLabelFor` and the prompt heading interpolate them raw. A crafted name containing newline/ESC can forge or reposition the menu. Escape each fragment before composing the prompt, and cover the interactive path.

2. **The human-error credential test does not reach the claimed error** — `tests/e2e/verbose/control-chars.test.ts:80-92`

   Its unreachable git tap fails during indexing, so the run exits through `invalid_ref` before `acquireTap` can produce the URL-bearing `no_skills_found` message. Seed a valid local clone with a credential-bearing configured URL and request a missing subpath; assert both absence of the secret and presence of the redacted URL.

3. **The multiline ambiguity test is vacuous** — `tests/e2e/verbose/control-chars.test.ts:107-143`

   Both configured taps are unreachable, so no duplicate candidates exist and `ambiguityError` is never exercised. The conditional assertion can pass without seeing the hostile name. Use indexed fixtures containing the same skill, force the ambiguity/abort path, and assert `ambiguous_reference`, the escaped hostile name, and no forged line.

### Nit

- `src/git/exec.ts:42-43` eagerly formats and URL-parses every Git argv even when verbose output is disabled, then sanitizes it again at the progress boundary. A lazy progress renderer would avoid the extra work, though subprocess startup dominates the practical cost.

### Reviewer coverage

- clarity: 1 finding
- naming: no findings
- type safety: no findings
- database/state: no findings
- testing: 2 findings
- security: no additional findings
- error handling: 1 finding
- observability: no findings
- performance: 1 finding
- concurrency: no findings

Deduplicated overlaps are counted once. The credential leak is release-holding, so I did not approve.

stevekrenzel:
Round three addressed in `25babae`. All six reproduced first; nothing declined.

**Major — human error output leaked credentials.** Reproduced before changing anything: the same error rendered both ways gave `mode=human leaks=true`, `mode=json leaks=false`. Round two redacted at the JSON boundary only, while the human path wrote `err.message` verbatim, and `acquireTap`'s `no_skills_found` interpolates `tap.url` — enough on its own. Fixed in `writeMessageBlock` (`src/cli/output.ts`), the single funnel every human message *and* remedy hint passes through, rather than at the message sites, so a future error inherits it. Reused the existing `redactText`; it scans prose for URL-shaped substrings, which a whole-string URL parser cannot do when the URL is embedded in a sentence.

**Sinks now covered:** verbose progress (`progress()` sanitizes its own argument), quoted stderr in `cloneRepo`/`fetchAndCheckout`/`checkoutSha`, the `--json` payload (message and `details`), and now the human error path including remedy hints. I found no remaining unredacted sink.

**Minors 1 and 2 — solved together, since they are one problem.** Sanitization ran over the *whole composed message*, where layout and content are both just `\n`. So `multiline: true` spared every newline including one inside a tap name, and the single-line default escaped the trusted breaks a genuinely multi-line error had composed. Both reproduced: a hostile tap name survived as a real forged line, and a two-line message rendered as literal `\x0a`.

Fragments are now escaped **as they are interpolated** — `formatCandidate`, and `ambiguityError`'s `name`/`reason` — which is the only point where trusted structure is still distinguishable from untrusted data. `ambiguity-prompt.ts` had a hand-rolled copy of the same message; it now raises `ambiguityError` too, so the layout is composed in exactly one place and the abort path and the non-interactive path read identically.

**Minor 3 — inverted to an allow-list.** The blocklist failed open, which I measured: of `token`, `client_secret`, `secret`, `auth`, `sig`, only `token` was masked. For a security control the failure mode should be a needlessly masked identifier, not a published credential, so unknown parameters are now redacted and only git's own transport parameters are displayable.

**Minor 4 — the vacuous test, and it was worth more than the finding suggested.** You were right that `list` emits no progress, so the "inner run must not borrow the outer sink" assertion could never fail. Rewritten to nest an install. **It fails against the pre-fix sink behaviour where the old one passed** — and writing it surfaced a second problem: both runs share the redirected adapter root, so two skills named `demo` collide and the outer run reports "already installed" instead of emitting its install line. The inner repo now uses a distinct name, with the reason recorded in `makeRepo`.

**Two things beyond the findings**, both found by the new tests rather than by reading:

1. A **remedy hint is not sanitized the way a message is** — it bypasses `CrewError` entirely. The `Searched: <taps>` hint joined raw tap names, so a newline in a configured name forged what read as a second crew error. Fixed in `resolve-ref/index.ts`. I confirmed the new test is load-bearing: reverting just that change fails it.
2. The docstring on `sanitizeBlock` claimed interpolated values "reach here already escaped". That was not true when written; it is now.

**Verification.** Targeted only, per the resource directive. `tests/e2e/verbose/` + `tests/unit/redact.test.ts`: 26 pass. `install-known-tap` + `namespaces` (the ambiguity paths): 20 pass. `error-paths` + `tests/commands/`: 12 pass. Typecheck and Biome clean.

Two failures I did **not** attribute to flakiness by assumption: `uninstall and reinstall report removal and store reuse` and `upstream repo gone -> tap-refresh warns` both fail intermittently at 5–77s timeouts. I stashed my work and reproduced the second on the completely unmodified base, and the first runs `crew update` against a home that keeps the GitHub-backed default tap. That is the non-determinism #131 removes, not this change.

**Note for the other dry-run PRs:** #107's round-three fix puts dry-run suppression in `isSuppressed`, so this branch and #109/#112/#119/#122/#130 inherit it — none should add its own copy.

stevekrenzel:
Rebased onto `main` after the cap-compliance refactors (#135–#143) merged. No behavior change from the reconcile; the four commits are unchanged in intent.

What moved, and where it landed:

- **`src/git/repo.ts` → `src/git/repo/{index,refs}.ts`.** This branch had independently split the same file (adding its own `src/git/refs.ts`) to get under the cap. `main`'s split won and this branch's duplicate was dropped, with each redaction edit re-placed into whichever half now owns the function: `cloneRepo` / `fetchAndCheckout` (`safeUrl`, `redactText`, redacted `details.url`) stayed in `repo/index.ts`; `checkoutSha`'s `redactText(...)` moved into `repo/refs.ts`, whose docstring now carries the same §5.2 note. Note the boundaries differed — this branch kept `checkoutSha`/`initRepo` in `repo.ts`, `main` moved them to `repo/refs.ts`.
- **Import specifiers** merged without duplication in `src/commands/tap/refresh.ts` (`ensureRepo` from `repo/index.ts`) and `src/sources/acquire/index.ts` (`ensureClone` from `repo/index.ts`, `resolveRef` from `repo/refs.ts`).
- **`tests/misc/{coverage-closeout,coverage}.test.ts`** are gone on `main` (now `tests/misc/closeout/`). Both edits here were only path-rewrites for this branch's superseded split, so they were dropped rather than resurrected — `tests/misc/closeout/git.test.ts` already covers `classifyRef`/`resolveRef`/`initRepo` through the new paths. The one piece of real intent, fixing the stale `describe` title, was applied there: `repo.ts:91-92` → `repo/refs.ts`.

Two files would have breached the newly-enforced 200-line cap, so each was split and folded into the commit that grew it (no separate fix-up commits):

- `tests/e2e/verbose/safety.test.ts` (223) → credentials stay in `safety.test.ts` (85); the control-character `describe` moved to `tests/e2e/verbose/control-chars.test.ts` (144). The shared `tapWithUrl` helper moved up into `tests/e2e/verbose/helpers.ts`. All five `C-CLI-06b` tests are preserved.
- `src/install/resolve-ref/index.ts` (201) → `lookupInTap` / `safeIndex` extracted to `src/install/resolve-ref/lookup.ts` (181 / 46). `NonTapNameCandidate` stayed exported from `index.ts` so no importer changed.

`bun run check` is clean: typecheck passes, lint passes including the file-size gate, 1121 pass / 0 fail, coverage 100.00 / 100.00.

PR #109, opened 2026-09-11
[LOG-5753] fix(update): honor --dry-run instead of writing updates

stevekrenzel:
## What

`crew update --dry-run` was not a preview. It wrote the new version of every moved skill into every agent and installed newly added tap children, exactly like a real run. PRD §10.1.1 already promised a preview and criterion C-UPD-18 named it, but no test exercised the flag.

This PR makes it a real preview:

- Taps are still fetched (that's how crew learns what moved; clones aren't user-visible install state).
- Per-skill moves report `would_update` (with the `new_sha` a real run would install). The skill is still validated so a broken upstream still shows as `failed`.
- Newly added tap children report `would_add` instead of being installed.
- Nothing is staged into the store, no agent directory or marker is touched, `state.json` is not written, and the store is not garbage-collected.
- Human output tags the header `(dry run)` and says "would update" / "would add". `--json` carries `dry_run: true`.

## Why

A user (or autoupdate operator) asking "what would change?" must not have their skills rewritten. This is the same class of bug as the pending `uninstall --dry-run` fix.

## PRD

- §10.1.1: replaced the one-line dry-run promise with the precise semantics above.
- §18.3 C-UPD-18: made precise (fetch, `would_update`/`would_add`, byte-identical disk, `dry_run: true`).

## Also

- The path-tap up-to-date check now uses `hashDirectory` on the source instead of staging a tentative store entry. `hashDirectory` ignores a root `.crew.json`, so the hash is identical to the store's; this is behavior-preserving and keeps the store untouched on preview.
- `commands/update/render.ts` row formatting moved to `rows.ts`; `install/update/entry.ts` per-agent reinstall loop moved to `reinstall.ts` (200-line cap).
- `crew help update`, README command table, and the site command reference now mention `--dry-run`.

## Tests

`tests/e2e/update-dry-run/` — `preview.test.ts` (2), `guards.test.ts` (3), `clone-lock.test.ts` (5) — plus `tests/unit/tap-scan-cache.test.ts` (4) and `tests/unit/tap-reexpand.test.ts` (2). All use real `file://` repos; none touch the network.

The preview tests install from a local git tap, move it upstream (an edit plus a new sibling), run `update --dry-run` in human and JSON mode, and assert the installed `SKILL.md` bytes, markers, store listing, and `state.json` are byte-identical before and after, that the new skill is not installed, and that a subsequent real `update` still applies the change. A path-kind tap case covers content-hash comparison and `new_sha: null`.

The clone-lock tests (C-UPD-18d) hold a tap's lock and assert that a real run, a dry run, and `crew tap update` each block and exit `state_locked`, that an unrelated tap's lock does not block, and that `crew cache clean` cannot delete a held lock. The scan-cache tests pin which failures are the skill author's fault (`invalid_skill`) and which are the local filesystem's (rethrown as-is).

Coverage is back to 100% on both metrics, which was the round-three blocker (the required `check` job had been failing at 99.99%).


stevekrenzel:
## `/review`

Verdict: **not ready — 0 blockers, 4 majors, 4 minors, 1 nit**

### Major findings

1. **The documented “without writing anything” promise is false** — `README.md:168`, `site/components/sections/Commands.data.tsx:30`, `src/commands/help/content/update.ts:16`, `src/install/update/types.ts:19`, `tests/e2e/update-dry-run.test.ts:62`  
   Dry runs intentionally call `refreshTaps`, which fetches and checks out tap clones. The PRD describes that exception, but these surfaces promise no writes and the test snapshot silently excludes tap clones.  
   **Fix:** state the boundary precisely: tap clones refresh, while installed skills, markers, store entries, and `state.json` remain unchanged.

2. **An empty-home dry run creates `state.json`** — `src/commands/update/index.ts:66`  
   `update --dry-run` still enters `withStateLock`; lock acquisition touches a missing state file. This violates the PRD’s narrower claim that `state.json` is not written.  
   **Fix:** avoid the mutating state lock for the read-only post-refresh plan, or make dry-run locking non-creating. Add an empty-config/no-state-file test that asserts the file remains absent.

3. **New tap children are reported as `would_add` without full validation** — `src/install/tap-reexpand.ts:168`  
   Discovery validates the declared name only. The dry-run branch skips the install callback, so a child with a valid name but invalid required frontmatter can be reported as `would_add` with exit 0 even though the real update would fail. Error-handling and testing independently flagged this.  
   **Fix:** run the same full skill validation before emitting `would_add`; record the same failure/hard-failure semantics a real run would produce. Add an invalid-new-child regression test.

4. **Changed path skills are fully hashed an extra time** — `src/install/update/entry.ts:137`  
   The new preflight `hashDirectory(skillDir)` is followed by `stageIntoStore`, which hashes the same unbounded source tree again and then hashes the staged copy.  
   **Fix:** retain the preflight source hash and pass it into staging/short-SHA derivation, while still hashing the staged copy for its canonical content hash.

### Minors

- `src/commands/update/rows.ts:68` and `src/commands/update/render.ts:135`: both use catch-all failure handling for `Outcome`. Handle `failed` explicitly and add a `never` exhaustiveness assertion so future outcomes cannot silently render/count as failures.
- `tests/e2e/update-dry-run.test.ts:42`: derive the JSON test shape from `UpdateRow` and `TapReexpandRow` instead of weakening both discriminants to `string`.
- `src/install/update/entry.ts:140`: add a short comment explaining that `loadSkill` intentionally runs before the dry-run return so invalid upstream content still fails validation.
- Add the explicitly documented pinned `--force --dry-run` case; line coverage does not verify that combination.

### Nit

- Rename or narrow the “nothing on disk changed” test wording because the snapshot intentionally excludes refreshed tap clones.

### Reviewer coverage

- clarity: 1 finding
- naming: no findings
- type safety: 3 findings
- database: no findings
- testing: 2 findings
- security: no findings
- observability: no findings
- performance: 1 finding
- concurrency: no findings

Deduplicated overlaps are counted once. CI is green, but the majors prevent approval.

stevekrenzel:
All 9 findings addressed in `6ef2477`. One finding turned out to be inverted — details under Major 3.

## Majors

**1. "Without writing anything" was false.** A dry run does refresh tap clones, and four surfaces claimed otherwise. Reworded all of them to state the boundary: collections still refresh, nothing installed changes. `README.md` (command table), `site/components/sections/Commands.data.tsx`, `src/commands/help/content/update.ts`, and the `would_update` docstring in `src/install/update/types.ts`. The test helper is now `snapshotInstalledState` with a docstring saying tap clones are deliberately excluded and why, so the exclusion is explicit rather than silent.

**2. Empty-home dry run created `state.json`.** Confirmed: `acquireStateLock` does `ensureDir` + `touch` when the file is absent. Followed #107's pattern — tap refresh, re-expansion, and the per-skill loop moved into `planUpdate` in `src/commands/update/index.ts`, shared by both paths. A dry run calls it directly; a real run calls it inside `withStateLock` and then persists and GCs. One decision function drives both, so preview and real run cannot disagree. New test C-UPD-18a asserts both `state.json` and its `.lock` stay absent.

**3. `would_add` without full validation — the premise was inverted, and the real bug is worse.** I built the case before fixing: a tap gains a child with a valid `name` but no `description`. The review says dry run reports `would_add` "even though the real update would fail". The real update does not fail — it exits 0 and **installs the invalid skill**. `installNewTapChild` calls `stageIntoStore` directly, and neither it nor `installSkillIntoAgents` validates. So the dry run was accurately previewing a real defect: §9 step 4 validation was being skipped entirely on the re-expansion path, in violation of §10.1.1's "runs the install algorithm (§7.3)".

Fixed the real path rather than only the preview. Both now run `loadSkill` before staging, report `invalid_skill` as a per-child failure, and exit 1. The invalid child no longer reaches disk. Verified end to end: before, `~/.claude/skills/bad/SKILL.md` existed with no description; after, it does not, on both paths. New test C-UPD-18c covers both.

**4. Path skills hashed twice.** `shortShaFor` / `stageIntoStore` take an optional `sourceHash`; `entry.ts` computes the preflight hash once (only for path-kind taps, where `newSha === null`) and threads it through. The staged copy is still hashed separately for its canonical content hash, so store layout and marker `content_hash` are unchanged — confirmed by the existing store and dry-run tests.

## Minors

- **Exhaustive `Outcome` handling** — `formatRowParts` and `symbolFor` in `rows.ts` and `tally` in `render.ts` now handle `failed` explicitly, guarded by `o satisfies Extract<Outcome, { kind: "failed" }>`. I used `satisfies` rather than an `assertNever` helper because its `throw` body is unreachable and would cost a line under the 100% gate — CLAUDE.md's "don't test dead code" rule. A newly added `Outcome` kind is still a compile error.
- **Typed test JSON** — `UpdateJson` derives from `UpdateRow` and `TapReexpandRow` instead of `string` discriminants.
- **`loadSkill` comment** — added in `entry.ts`, noting it runs before the dry-run return on purpose so broken upstream content still surfaces as `failed` in a preview.
- **Pinned `--force --dry-run`** — new test C-UPD-18b: without `--force` a pinned entry whose upstream moved is `skipped`; with `--force` it previews `would_update` and still writes nothing.

## Nit

- `snapshot` renamed to `snapshotInstalledState`; the test docstring and the describe-block wording no longer claim "nothing on disk changed" unqualified.

## Pre-fix verification

Stashed `src/` only (keeping the new tests) and ran against the pre-fix commit: **C-UPD-18a and C-UPD-18c both fail**, the other three pass. So both new guards are genuine regression tests, and the existing three prove no behavior change.

## Structure

`tap-reexpand.ts` would have hit 224 lines with validation inline, so the additions loop moved to `src/install/tap-additions.ts` (101 lines); `tap-reexpand.ts` is back to 184.

## PRD

§10.1.1 now records the no-lock guarantee, names refreshed tap clones as the sole exception with a MUST NOT on claiming otherwise, and makes child validation mandatory on both paths. Criteria C-UPD-18a, C-UPD-18b, C-UPD-18c added next to C-UPD-18 (C-UPD-19 was already taken).

`bun run check`: 990 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean. All `src/` files under the 200-line cap.

stevekrenzel:
## Updated `/review`

Verdict: **not approved — 1 blocker, 5 majors, 3 minors**

### Blocker

- **Dry runs mutate shared tap clones without synchronization** (`src/commands/update/index.ts:66`). A real update can resolve SHA A, then a concurrent dry run checks out B before the real update copies source bytes, causing it to record A for B’s installed content. Add per-tap locks spanning refresh through source read/staging, acquired in stable tap-name order.

### Majors

- `PRD.md:146`, `PRD.md:1855`, help, and tests still say dry run changes/writes “nothing,” despite intentionally refreshing clones. State explicitly that clones refresh while installed skills, markers, store, and state stay unchanged.
- Invalid newly discovered children lose their name and validation message in human output and totals (`src/commands/update/render.ts:87`, `src/install/tap-additions.ts:48`). Render the child, error code/message, and count `tap_error` failures; test human output.
- `tests/e2e/update-dry-run.test.ts` is 229 lines, violating the hard sub-200-line cap. Split scenarios or extract snapshot/setup helpers.
- Each `(tap, scope, project_root)` group rescans and revalidates every child (`src/install/tap-reexpand.ts:131`, `src/install/tap-additions.ts:48`). Cache one tap snapshot and per-child validation result per run.
- The `would_update` type docs claim a tap clone was refreshed (`src/install/update/types.ts:20`), but path-tap updates also emit it. Qualify that statement to Git taps or remove it.

### Minors

- The byte-purity snapshot (`tests/e2e/update-dry-run.test.ts:63`) can miss changes to other installed resources or existing store contents. Recursively snapshot the installed/store trees or include a resource-file regression.
- `tap-reexpand.ts` plus `tap-additions.ts` violates the directory-over-prefix convention. Use `tap-reexpand/index.ts` and `tap-reexpand/additions.ts`.
- `validationErrorFor` (`src/install/tap-additions.ts:93`) casts every thrown value to `CrewError`, which can mislabel I/O failures as `invalid_skill`. Narrow expected validation errors and rethrow/report unexpected failures separately.

All eleven lenses reran against updated head `6ef2477`; CI is green. I did not approve because a blocker and majors remain.

stevekrenzel:
All 9 findings addressed in `c985aeb`. Nothing declined. One of my own fixes introduced a regression that an existing test caught — details under Major 4.

## Blocker — dry runs mutate shared tap clones without synchronization

Confirmed by tracing the window: `acquireTap` resolves `newSha` from the clone's HEAD, then `stageIntoStore(loaded.path, …)` copies bytes out of the clone's **working tree**. In between, another run's `refreshTaps` → `ensureRepo` → `fetchAndCheckout` does a detached checkout of that same tree. So state records commit A while the installed bytes are B — silent provenance corruption, and the dry-run path made it worse by fetching without taking any lock at all.

Every update run now holds a **per-tap advisory lock** spanning refresh → re-expansion → per-skill source read and staging. Dry runs take them too, because a preview still fetches. Acquisition is in sorted tap-name order so two runs touching the same set can't deadlock.

Rather than write a second lock implementation, I extracted the retry/stale/timeout loop from `state/lock.ts` into `state/advisory-lock.ts` and built both on it — the state lock is now a thin wrapper. Lockfiles live under `cache/locks/` so nothing in `taps/` is mistaken for a tap directory, and the tap name is hashed into the filename so a name persisted in config can't escape that directory.

**Ordering:** the state lock is taken *outside* the tap locks on the real path. I grepped every `withStateLock` call site; no other command takes tap locks, so state-then-tap is the only ordering in the codebase and there's no cycle to deadlock on.

**Could I test it deterministically?** Yes, without racing two processes. A true interleaving test would need deliberate timing and would flake. Instead `tests/e2e/update-dry-run/clone-lock.test.ts` holds a specific tap's lock and asserts that both a real and a dry run block on it (exit 7, `state_locked`), and that an *unrelated* tap's lock doesn't block. If `update` didn't lock, those runs would sail past and succeed. **Verified load-bearing:** replacing `withTapLocks` with a pass-through makes both blocking tests fail; restored, they pass. The 30 s timeout is overridable via `CREW_LOCK_TIMEOUT_MS` so the tests don't wait — an env seam like `CREW_HOME`/`CREW_NOW`, rather than new option surface on `runCli`.

PRD §14 point 4 previously claimed clone access was "serialized under the state lock", which is what the dry-run path broke. It now specifies per-tap locks, the ordering rule, and that previews take them. New criterion C-UPD-18d.

## Majors

**1. Surfaces still promising "nothing".** §5.2's global `--dry-run` description and C-UPD-18 both still said it. Both now state the boundary explicitly — no installed skill, marker, store entry, config, or `state.json` is written, with read-cache refresh called out as an exception a command's own section must document. Also fixed the `crew help update` example gloss ("without changing anything" → "without touching anything you have installed"), which round one missed.

**2. Invalid children lost their name and message.** Reproduced: a child failing validation rendered as `⚠ tap acme (invalid_skill)` — no child name, no reason, and uncounted in totals. Since the user is the one who has to go fix it, output now names the child, prints the validation message, and counts it as a failure. `guards.test.ts` asserts the human output contains the child name, the code, *and* the message text.

**3. File cap.** The 229-line test file became `tests/e2e/update-dry-run/` — `helpers.ts` (76), `preview.test.ts` (103), `guards.test.ts` (116), `clone-lock.test.ts` (99).

**4. Repeated scan/validate per group.** One tap backs several `(tap, scope, project_root)` groups, so acquisition, the child walk, and per-child validation ran once per group. Now memoised per run: **at 8 groups over 12 children, 8 acquisitions and 96 validations drop to 1 and 12.**

> **My first version of this broke something, and an existing test caught it.** Folding the child walk into the cached `scan()` moved it *inside* the `try` that catches acquisition failures — so an unreadable tap directory stopped propagating and was reported as a `tap_error` row instead. That's precisely the mislabelling flagged in the third minor, reintroduced elsewhere. `reexpandTaps > tap expansion rethrows filesystem errors` failed, which is how I found it. The cache now splits `acquire()` (whose `CrewError` is legitimately caught) from `children()` (whose filesystem errors propagate), preserving the original boundary.

**5. `would_update` docstring.** Claimed a tap clone was refreshed; path-tap updates also emit it and have no clone. Qualified to git taps, noting path taps compare by content hash.

## Minors

- **Snapshot depth.** The byte-purity check compared only `SKILL.md` per skill plus a store *listing* — a changed resource file or a restaged store entry would have slipped through. It now compares every file in both the installed and store trees, and the fixture carries a `reference.md` that changes upstream, so the assertion exercises the gap.
- **Directory convention.** `tap-reexpand.ts` + `tap-additions.ts` → `tap-reexpand/{index,additions,scan-cache}.ts`.
- **`validationErrorFor` cast.** Confirmed reachable: `loadSkill` calls `readText`, so an EACCES/EIO surfaces as a plain `Error` and would have been reported as `invalid_skill` — blaming the skill author for a local filesystem fault. It now returns only `CrewError` verdicts and rethrows anything else.

## Verification

Targeted suites: **19 pass, 0 fail** across `update-dry-run/`, `tests/unit/tap-reexpand.test.ts`, `update-dep-closure`, `update-fetch-scope`. Typecheck and Biome clean; 100% line and function coverage on every new file; all files under the 200-line cap.

I'm **not** quoting a full-suite number, because it wouldn't mean anything on this branch. The suite is non-deterministic without #131: the failing set differs run to run and clusters in tests that clone from remotes. I checked that none of it is mine — `tests/e2e/multi-skill-install.test.ts` fails the same way on the unmodified base `6ef2477` (4–6 failures, varying, with 5 s timeouts taking 19–150 s wall time), so those failures predate this change. The one failure that *was* mine is the scan-cache regression above, now fixed.

stevekrenzel:

**Verdict:** 0 blockers, 2 majors, 8 minors, 1 nit. **Not approved.**

Reviewed `72e6acf..2b578c5` through all 11 `/review` lenses. CI is green at this head.

### Majors

1. **Installs still do not participate in the new tap-lock protocol** — `src/install/flow.ts:77`, `src/install/resolve/index.ts:118`, `src/commands/tap/index.ts:136`

   Install can resolve commit A, then a concurrent `tap update` can advance the shared checkout to B before install stages it. State and markers then record A for B's bytes. Hold every relevant tap lock across install's complete acquire/read/stage span and audit the remaining clone readers against the new §14 guarantee.

2. **A live tap lock can expire during synchronous Git work** — `src/util/advisory-lock.ts:51-54`, `src/git/exec.ts:72`

   The lock is stale after 60 seconds, but `proper-lockfile` refreshes through timers and crew performs fetch/clone with synchronous subprocesses. A Git operation lasting over 60 seconds blocks the heartbeat, allowing another process to reclaim the live lock and mutate the checkout concurrently. Use a lock whose ownership cannot expire while the process is alive, or move protected operations to an execution model that can maintain the lease.

### Minors

1. `tap remove` deletes a clone without its tap lock, so it can remove a checkout while a dry-run update is reading it (`src/commands/tap/index.ts:82-90`).
2. The lock set is based on initially selected taps, but re-expansion can consult a same-named skill from another tap; ensure every clone re-expansion may read is locked (`src/commands/update/plan.ts:63`).
3. `TapReexpandRow` permits `kind: "tap_error"` without an error payload. Make it a discriminated union that requires code/message for that kind (`src/install/tap-reexpand/index.ts:33-39`).
4. Add a dry-run test where an already-installed skill becomes invalid upstream; assert `failed/invalid_skill`, nonzero exit, and unchanged installed bytes (`tests/e2e/update-dry-run/guards.test.ts:44`).
5. The byte-identical fixture decodes resources as UTF-8, which cannot distinguish all binary changes. Compare buffers and include a binary resource (`tests/e2e/update-dry-run/helpers.ts:61`).
6. Restore the previous `CREW_LOCK_TIMEOUT_MS` value during test teardown instead of always deleting it (`tests/e2e/update-dry-run/clone-lock.test.ts:35`).
7. Cache failed tap acquisition as well as successful acquisition so one unreachable tap is not retried for every project group (`src/install/tap-reexpand/scan-cache.ts:52`).
8. Dry-run documentation omits transient lockfile writes, and `plan.ts` incorrectly says update never installs new skills even though re-expansion does (`src/commands/update/index.ts:30`, `src/commands/update/plan.ts:53`).

### Nit

1. The on-disk-layout ownership paragraph is duplicated in `PRD.md:336`.

The previous coverage failure is fixed. The install-lock issue remains from the last pass and is still release-relevant.

stevekrenzel:
Round three addressed in `67385b3`. All eight reproduced first; nothing declined.

**Blocker — coverage gate.** Confirmed `check` was failing. The uncovered line is the non-`CrewError` rethrow in `scan-cache.ts`, and I checked whether it is dead before reaching for a test: `loadSkill` → `readText` → `readFileSync`, which raises a raw Node `Error` on EACCES/EIO. So the branch is live, and it is exactly what stops a local filesystem problem being reported as `invalid_skill` and blaming the skill's author — the mislabelling round two fixed. Deleting it would reintroduce that, so I covered it instead.

Reaching it without a read-only directory (which hung four processes for another worker on a sibling PR): a `SKILL.md` that exists as a *directory* passes the existence guard, then `readFileSync` raises `EISDIR`. Verified the thrown value is an `Error` that is not a `CrewError`, with `code === "EISDIR"`.

**Overall coverage is back to 100.00% on both metrics**, and `scan-cache.ts`, `tap-lock.ts`, and `advisory-lock.ts` are each at 100%.

**Major 1 — unlocked clone consumers.** `crew tap update` fast-forwards working trees while taking no lock, so it could move a checkout out from under an install that had already resolved a SHA from it. It now wraps `refreshTaps` in `withTapLocks`.

I deliberately did **not** push the lock down into `refreshTaps` or `acquireTap`, which looks like the tidier fix. The locks are not reentrant — I measured it: a nested acquire blocks for the full timeout and then fails — so a helper-level lock would deadlock `crew update` against itself, since its plan already holds the same locks. The lock belongs at the command boundary because the hazard is the window between resolving a SHA and reading bytes at it, and that window spans several calls.

`crew info` is left unlocked on purpose. It never fetches (no `ensureRepo`/`fetchAndCheckout` on its path), so it cannot move a checkout; it only reads whatever is currently there. Locking a read-only preview would make `crew info` block behind a long `crew update` for no correctness gain. Said here rather than left silent.

**Major 2 — `cache clean` could delete a live lock.** Round two put lockfiles under `cache/locks/`, and `cache clean` deletes that tree wholesale. They now live in `locks/` at the crew-home root. I checked every deletion path rather than just the reported one: `cache clean` removes only `cacheDir`; tap removal and auto-tap GC remove individual clones under `taps/`; doctor and store GC only walk `storeDir`. Nothing touches `locks/`.

**Minors.**
1. The dry-run contract now states exactly what is and is not written — including the transient lockfiles, which the previous wording omitted while claiming refreshed clones were "the only thing" written. Fixed in both the code comment and PRD §10.1.1.
2. `tap-lock.ts`'s docstring no longer claims every clone reader locks, and now records *why* the lock wraps a command rather than a helper, so the next person doesn't re-attempt the deadlocking version.
3. `state_locked` in §13 now covers both lock kinds rather than reading as state-file-specific.
4. The advisory-lock primitive moved from `state/` to `util/`, since it serves two kinds of lock. Git tracked it as a rename, so history is preserved.

PRD §6's layout gains `locks/`, with a note that it must not live under `cache/` and why — the constraint is externally observable, and a second implementation that put locks in the cache tree would hit this same bug.

**Nit — description.** The Tests section named a single file that is now a directory of four and omitted the lock and scan-cache work. Updated to the real inventory.

**Pre-fix verification.** Stashing `src/` only and running the new tests against the old source fails exactly the two new lock tests — `crew tap update` takes the same lock, and `cache clean` does not delete a held lock — with the seven pre-existing tests still passing. The four scan-cache tests pass either way, which is expected: the rethrow already existed and was merely uncovered, so they close a coverage gap rather than catching a regression. I'd rather say that than let "9 new tests" imply nine caught bugs.

**Targeted results.** `update-dry-run/` + `tap-update` 18 pass; `tap-scan-cache` + `clone-lock` 9 pass; `tap` + `tap-reexpand` + `gaps` 79 pass / 1 fail. That one failure is a `search` test timing out at ~14 s, and a different `search` test failed on the previous run — a shifting set, which is the signature of the network non-determinism rather than a real regression. I confirmed it rather than assuming: stashing all my work and running `tests/e2e/tap.test.ts` on completely unmodified source reproduces the identical failure. That is what #131 removes.

The full suite still shows 14 failures for the same reason, consistent with the 13–18 measured on branches lacking #131. Typecheck and Biome clean; every touched file under the 200-line cap.

**Inherited, not duplicated:** #107's round three fixed centrally in `isSuppressed` the fact that `--dry-run` still wrote `version-check.json` on a TTY. That is not in this base, and I did not add a copy — this branch picks it up at merge.

stevekrenzel:
Rebased onto `main` after #111/#112/#115/#116/#119 merged. One conflict reconciled in `src/commands/tap/index.ts`: #119's `--dry-run` path (`planRefresh`) and this PR's tap-clone locking now coexist — a dry run plans without taking tap locks; a real refresh runs under `withTapLocks`. README rows from both sides kept. All commits re-signed; targeted update tests pass.

stevekrenzel:
Rebased onto `main` after the cap-compliance refactors (#135–#143) merged. Two modify/delete conflicts, both on the deleted `site/components/sections/Commands.data.tsx`: this PR's two edits to the `update` row (the `--dry-run` sentence, then its "taps still refresh" rewording) were re-placed byte-for-byte into `Commands.data/managing.tsx`. Nothing else conflicted. All four commits re-signed. `bun run check`: typecheck, lint (incl. the new file-size gate), full suite 1106 pass at 100% line/function coverage; site lint clean.

PR #110, opened 2026-09-11
[LOG-5754] fix(uninstall): honor --scope instead of removing every scope

stevekrenzel:
## What

`crew uninstall --scope project <name>` removed the **user-scope** install (and plain `crew uninstall` removed project installs too). Selectors matched state entries by name across every scope, so `--scope` was accepted and ignored, even though `crew help uninstall` documents it.

A selector now targets exactly one scope:

- no `--scope` / `--scope user` → the user-scope entry only
- `--scope project` → the project entry whose `project_root` is the cwd; if none matches but exactly one project install exists, that one (keeps C-UNINST-15: schedulers can run from any directory)

When the skill is installed only at other scopes, the command fails with `not_installed_here` (exit 6) and the remedy names where it *is* installed and the exact command to remove it. `--json` details carry `installed_locations: [{scope, project_root}]` (not `installed_at`, which means an ISO timestamp everywhere else in the state and marker contracts). `--force` turns the miss into a no-op as before.

## Also fixed along the way

- `--prune` now only sweeps orphans at the targeted scope and project root. PRD §7.4 step 5 already said "at the same scope"; the implementation searched every scope.
- The uninstall state helpers keyed entries on `(name, scope)`, so removing a project install at `~/a` also dropped the state entry for the same skill at `~/b`. They now key on `project_root` too.
- Adds `scopeGiven` to `CommandFlags` so a command can tell an explicit `--scope` from the default. A sibling PR for `crew list --scope` reuses it.

## PRD

- §7.4: new **Scope** paragraph.
- §18.3: C-UNINST-15a, 15b, 15c.

## Tests

New `tests/e2e/uninstall-scope.test.ts` (8 tests, real installs at user + two project roots): default scope removes only user; `--scope project` from one root leaves the other; lone project install reachable from any cwd; two roots + unrelated cwd errors listing both roots; user-only + `--scope project` errors with hint; project-only + plain uninstall errors with `--json` details and command hint; `--force` no-op; `--prune` scoped. `bun run check` green: 993 tests, 100% line and function coverage.


stevekrenzel:
## `/review`

Verdict: **not ready — 2 blockers, 2 majors, 4 minors, 0 nits**

### Blockers

1. **C-UNINST-15 still promises project uninstall works from any cwd** — `PRD.md:1865`  
   The new §7.4 rule and tests reject an unrelated cwd when the name is installed in multiple project roots. Qualify this criterion to the lone-project-entry fallback or replace it with the new criteria.

2. **C-STATE-02 still defines entry identity as `(skill, scope)`** — `PRD.md:1913`  
   That contradicts §11.1 and this PR’s core fix, where project entries are distinct by `(skill, scope, project_root)`. Update the conformance criterion.

### Majors

1. **Removing one scoped parent strips dependency edges from every scope/root** — `src/commands/uninstall/state.ts:45`  
   `dropScopedEntryAndUpdateRequiredBy` removes `target.name` from every surviving entry’s `required_by`. If `foo → bar` exists in projects A and B, uninstalling foo in A marks B’s bar orphaned even though B’s foo remains; a later prune can delete that required dependency.  
   **Fix:** restrict the scrub to entries with the same scope and project root as `target`, and test user plus two project roots.

2. **`--force --prune` can delete unrelated orphans after a scope miss** — `src/commands/uninstall/index.ts:67,129`  
   On a forced miss, the selected entries are empty. `prunedRoots` then falls back to cwd and pruning runs anyway, so the command is not the promised no-op and may remove an unrelated project orphan. Partial `--agent` removals have the same gating problem.  
   **Fix:** prune only when at least one direct entry was fully removed, and derive roots only from those records. Add a forced-miss regression with a pre-existing orphan.

### Minors

- `src/commands/uninstall/core.ts:29`: `projectRoots` is internal prune-routing data, but `UninstallRecord` is serialized verbatim, exposing a new undocumented JSON field with absolute paths. Return private removal metadata separately.
- `src/commands/uninstall/scope.ts:70`: the promised copyable remedy interpolates `project_root` unquoted. Spaces break it and shell metacharacters can change its meaning. Use a POSIX shell-quoting helper and keep display shortening separate from the executable path.
- `tests/e2e/uninstall-scope.test.ts:213`: the new file violates the repository’s hard `< 200` line cap. Split the prune scenarios into a focused sibling test file/directory.
- `src/commands/types.ts:50`: `scopeGiven` has no consumer in this PR; the description says it is for a sibling PR. Add it with that consumer instead of shipping speculative public context state.

### Reviewer coverage

- clarity: 4 findings
- naming: 1 finding
- type safety: no findings
- database: no findings
- testing: 2 findings
- security: 2 findings
- error handling: 2 overlapping findings
- observability: no findings
- performance: no findings
- concurrency: no findings
- architecture/conventions: 2 overlapping findings

Deduplicated overlaps are counted once. CI is green, but the blockers and majors prevent approval.

stevekrenzel:
Thanks — both majors were real, and I verified each by reverting the fix and watching the new test fail. Pushed as `1d4dd1f`.

## Blockers

**1. C-UNINST-15 contradicted the new rule** (`PRD.md`). Qualified it: a lone project-scope entry is still reachable from any cwd (the fallback this PR keeps), but when several project roots hold the same skill the cwd selects among them and an unrelated cwd is `not_installed_here`. It now points at C-UNINST-15b/15c for those two cases instead of promising something the code rejects.

**2. C-STATE-02 defined identity as `(skill, scope)`** (`PRD.md`). Updated to the `(skill, scope, project_root)` triple, matching §11.1's own wording ("One entry per (skill, scope, project_root) triple") and this PR's fix.

## Majors

**1. Cross-location `required_by` scrubbing** (`src/commands/uninstall/state.ts`). Confirmed: the filter correctly removed only the target entry, but the subsequent `.map` scrubbed the removed name from *every* survivor. With `foo → bar` in projects A and B, uninstalling `foo` in A left B's `bar` with an empty `required_by` while B's `foo` was still installed — so a later `--prune` in B would delete a live dependency. Added a `sameLocation` predicate and restricted the scrub to entries sharing the target's scope and `project_root`. Dependency edges are per-location, which follows directly from the §11.1 identity rule in Blocker 2.

New test: *"uninstalling a parent in one project root keeps the other root's dependency edge"* asserts B keeps `required_by: ["foo"]`. **Fails against the pre-fix commit** (verified by reverting only `state.ts`).

**2. `--force --prune` pruned after a miss** (`src/commands/uninstall/index.ts`). Confirmed: `prunedRoots` did `if (roots.size === 0) roots.add(cwd)`, so a forced miss pruned at cwd despite removing nothing. Pruning is now gated on `removedRoots.length > 0`, and the roots come only from entries that were *fully* removed — which also fixes the partial-`--agent` case, since a surviving entry still needs its dependencies.

New tests: *"a forced miss with --prune leaves unrelated orphans alone"* (**fails pre-fix**, verified by restoring the cwd fallback), plus siblings for a forced *scope* miss and a partial `--agent` removal.

## Minors

- **`projectRoots` leaked into `--json`** (`src/commands/uninstall/core.ts`). Right — `UninstallRecord` is serialized verbatim, so this shipped absolute paths as an undocumented field. `removeOne` now returns a separate `RemovalMeta` carrying `fullyRemovedRoots`; the record is back to its documented shape. This dovetailed with Major 2, since "fully removed" is exactly the gate pruning needs.
- **Unquoted path in the copyable remedy** (`src/commands/uninstall/scope.ts`). Added `shellQuote` to `src/util/format.ts` (single-quote with `'\''` escaping) and used it for the `cd` target, while `shortenHome` still renders the human-readable `~/…` form. The two are deliberately not composed — `~` is shell-expanded, so quoting it would defeat the shortening. New test installs into a path containing a space and asserts the hint reads `cd '…/my project'`. I did not reuse `quoteArg` from `src/autoupdate/systemd.ts`: it is systemd-specific (doubles `%` for specifiers, uses double quotes) and would be wrong here.
- **Test file over the 200-line cap.** Split into `tests/e2e/uninstall-scope-prune.test.ts`, with shared setup in `tests/helpers/scoped-install.ts`. Both files are now 182 and 158 lines.
- **`scopeGiven` has no consumer** — this one is **stale**, and I checked rather than assuming. `gh pr diff 118` and `gh pr diff 124` both show `const scope: Scope | null = ctx.flags.scopeGiven ? ctx.flags.scope : null;` in `src/commands/list/index.ts`. Those PRs are stacked on this branch, so the consumer lands with them; removing the field would break both. Left as-is.

## One note on the helpers

The shared helper returns the install exit code rather than throwing, and callers assert with `expect(...).toBe(0)`. Two constraints pushed me here: an unreachable `throw` guard costs line coverage under the 100% gate, and Biome's `noStandaloneExpect` forbids assertions outside a `test()` body.

`bun run check`: exit 0, 999 pass, 0 fail, 100% line and function coverage.

stevekrenzel:
## Updated `/review`

Verdict: **not approved — 2 blockers, 2 majors, 2 minors**

### Blockers

- **The PRD still specifies the old cross-location dependency behavior** (`PRD.md:519`, `:1272`, `:1856`, `:1920`). It requires global `required_by` scrubbing and scope-wide pruning, contradicting the corrected `(scope, project_root)` implementation and inviting a conforming implementation to recreate cross-project deletion. Update every affected normative section and criterion to use the same full location identity.
- **Prune treats a safety-aborted removal as successful** (`src/commands/uninstall/core.ts:86`, `src/commands/uninstall/index.ts:121`). It drops state ownership and can prune descendants even while the untracked/customized directory remains, potentially collecting its backing auto tap. Preserve failed owners/state and seed recursive pruning only after confirmed physical removal. Test a protected middle node in a `foo → bar → baz` chain.

### Majors

- **`installed_at` is reused for location objects** (`src/commands/uninstall/scope.ts:37`). That name already means an ISO installation timestamp in state and marker contracts. Rename it to `installed_locations` or `locations` and update tests/spec wording.
- **Help omits the lone-project fallback** (`src/commands/help/content/uninstall.ts:16`). It says project scope removes the current project’s copy, but the implementation may remove the only project installation from another cwd. Document that behavior.

### Minors

- The fallback/location tests (`tests/e2e/uninstall-scope.test.ts:73`, `:87`) do not assert physical deletion at the stored root or the exact multi-root JSON payload. Add both assertions.
- Add a lone-project-fallback plus `--prune` case (`tests/e2e/uninstall-scope-prune.test.ts:88`) proving pruning follows the removed entry’s recorded root instead of cwd.

All eleven lenses reran against updated head `1d4dd1f`; CI is green. I did not approve because blockers and majors remain.

stevekrenzel:
Addressed all six findings in `0e64483`. Nothing declined; every one reproduced first.

## Blockers

**1. The PRD still specified the old cross-location behaviour.** Confirmed — §7.4 step 4 said "for every other entry whose `required_by` listed this skill", step 5 said "at the same scope", §11.1 defined `required_by` as "at the same scope", and C-UNINST-06, C-UNINST-08, C-UNINST-12 and C-STATE-10 all said the same. Round one scoped the implementation to `(scope, project_root)`, so a second implementation reading the spec would have faithfully recreated the cross-project deletion we removed. All four normative sites and all four criteria now use the full location identity, and §11.1 states the consequence directly: the same skill installed in two projects has two independent `required_by` lists.

**2. Prune treated a safety-aborted removal as successful.** Reproduced before changing anything, and it was worse than the report. With `foo → bar → baz` and `bar` made untracked:

```
exit: 0
state names: []          ← bar erased from state…
bar on disk: true        ← …while its bytes are still there
baz on disk: false       ← swept, though bar still requires it
```

So state claimed a skill was gone while the install site held it, its dependency was deleted, and the user was told everything succeeded. `removeOne` now keeps ownership for agents whose removal aborted, and `RemovalMeta.aborted` records the outcome so pruning seeds only from a confirmed physical removal. Same run after the fix: `exit 0`, state `["bar", "baz"]`, both present on disk.

### How the prune sweep terminates

Worth stating explicitly, because retaining state breaks the loop's old termination argument. `pruneOrphans` ran `while (findOrphan(...))` and relied on the removed entry *disappearing* from state. An orphan that aborts now stays, so that loop would return the same entry forever and append a record each pass — an unbounded loop.

`findOrphan` now takes the set of entry keys already attempted this run, and `pruneOrphans` records each candidate **before** removing it. The sweep therefore visits each entry at most once regardless of outcome, so termination is a property of the loop rather than of the entry vanishing. I re-read the loop before running anything against it, and added a third test that protects the *orphan itself* — the shape that would previously have spun — which completes in ~7 ms.

**Interaction with #130:** it fixes the sibling case (an aborted uninstall dropping its entry, so a forced retry orphans the install). Both land on the same rule — an abort keeps ownership — so the two should agree at merge; the difference is that this PR also had to bound the sweep that consumes those retained entries.

## Majors

**1. `installed_at` reused for locations.** That key is an ISO timestamp in eight places across `StateEntry`, `Marker`, and `doctor --repair`. Renamed to `installed_locations`; the test now also asserts `installed_at` is absent, so the collision can't come back silently.

**2. Help omitted the lone-project fallback.** `crew help uninstall` now says that a skill installed in exactly one project is still reachable from elsewhere, and that with several you run the command from the project you mean.

## Minors

- The fallback test asserted only state; it now asserts the directory is gone **at the recorded root**, which is the part that would break if the fallback used cwd. The multi-root test asserts both installs survive on disk and pins the exact `installed_locations` payload.
- Added a lone-project-fallback + `--prune` case: the project's dependency is swept at its recorded root while an unrelated user-scope orphan at the cwd is left alone.

## Verification

The three new tests in `tests/e2e/uninstall-abort.test.ts` **all fail against the pre-fix commit** and pass after (verified by stashing only `src/`). `uninstall-scope.test.ts` reached 204 lines with the new assertions, so the "installed elsewhere" cases moved to `uninstall-scope-miss.test.ts` (124 / 116).

Targeted suites: 148 pass, 0 fail across the uninstall, install, error-path and coverage suites; `uninstall/{core,scope,state}.ts` at 100% line and function coverage, `index.ts` covered to 100% once `coverage-closeout` is included. Typecheck and Biome clean.

I'm deliberately **not** quoting a full-suite number. The suite is non-deterministic on any branch without #131 — `origin/main` fails 3/985 on an idle machine and this stack's branches fail 13–18 with a different set each run, all clustered in tests that clone from remotes. One such failure appeared in my batch runs here (`error-paths.test.ts`, 5.2 s); it passes in isolation, passed on a re-run of the identical batch with these changes applied, and doesn't touch this code.

One thing I did **not** change: exit code stays 0 when the *direct* removal succeeds and only a pruned orphan aborts. §7.4 step 5 says skipped orphans are "reported, not forced", and `foo` did come off cleanly, so 0 looks right — but it's a judgement call worth a second opinion, since the run does leave a protected skill behind.

stevekrenzel:
Follow-up in `bf12ae6`: removed NUL bytes from `src/commands/uninstall/state.ts`.

My round-two prune-termination fix built `entryKey` by joining the `(name, scope, project_root)` triple with raw NUL bytes. Git classified the file as binary as a result, so **this PR's `state.ts` diff was hidden from review entirely** — the implementation of the termination guard could not be read by anyone reviewing it. I only found this because the same defect was reported independently on #126 and I checked whether it existed here too.

The key is now a JSON-encoded tuple. That also removes an aliasing hazard: a plain delimiter means a `project_root` containing that delimiter can produce the same key as a different triple, which for a key used to decide what gets removed is worth avoiding.

Verified the committed blob contains zero NUL bytes and git now classifies the file as text. The commit itself still renders as `Bin` because it is compared against the previous binary blob; subsequent diffs read normally.

Targeted: `uninstall-abort` + `uninstall-scope` 7 pass, 0 fail. Typecheck and Biome clean.

stevekrenzel:

**Verdict:** 0 blockers, 3 majors, 9 minors. **Not approved.**

Reviewed `72e6acf..1fbfea6` through all 11 `/review` lenses. CI is green at this head.

### Majors

1. **A later selector error can leave removed bytes recorded as installed** — `src/commands/uninstall/index.ts:58`

   The command removes each selector immediately but writes state only after the whole loop. For example, `crew uninstall foo project-only-bar` can remove `foo`, then the new scope check throws `not_installed_here` for `bar`; `foo` remains in state even though its files are gone. Resolve and scope-check every selector before any removal, or durably persist each confirmed removal before processing the next selector.

2. **Malformed project roots can target the wrong project or silently disappear** — `src/state/load.ts:33-47`, `src/commands/uninstall/scope.ts:54-58`

   Empty and relative strings pass `hasUsableLocation`; the lone-project fallback can then build a removal path under the caller's cwd or outside the intended project. Missing roots are silently filtered and later state writes can abandon their markers. Validate untrusted entries and require nonempty absolute project roots before destructive use; preserve/report entries that cannot be safely recovered rather than silently rewriting them away.

3. **Doctor still conflates independent project installs** — `src/commands/doctor/checks.ts:36-56`, `src/commands/doctor/repair.ts:75-103`

   This PR makes `(name, scope, project_root)` canonical, but doctor still matches and merges by `(name, scope)`. One project's marker can hide another project's missing marker, and repair can merge owners into the wrong project entry. Include `projectRoot` in drift checks, retention, lookup, and merge predicates.

### Minors

1. A protected orphan failure during `--prune` is reported but does not make the command exit nonzero (`src/commands/uninstall/index.ts:130`).
2. `state.json` entries are cast before validation; `installations: [null]` crashes the filter and blocks doctor recovery (`src/state/load.ts:33`).
3. The partial-agent prune test removes an agent that was never installed, so it does not exercise a successful partial removal (`tests/e2e/uninstall-scope-prune.test.ts:172`).
4. The malformed-root test does not run the project-scope destructive path and can pass without the intended guard (`tests/e2e/uninstall-scope-miss.test.ts:161`).
5. `scopeGiven` has no consumer and cites nonexistent `src/commands/list/index.ts` (`src/commands/types.ts:53`).
6. `RemovalMeta.fullyRemovedRoots` is documented as project-only but deliberately contains `null` for user scope (`src/commands/uninstall/core.ts:38`).
7. README/site describe only the current project and omit the intentional single-project fallback from another cwd (`README.md:168`, `site/components/sections/Commands.data/managing.tsx:24`).
8. PRD install/invariant wording still says dependency edges are same-scope rather than same `(scope, project_root)` location (`PRD.md:846`, `PRD.md:1357`).
9. Legacy `removeEntry` still deletes by `(name, scope)` and would erase every project-root entry if reused; remove it or accept the full entry identity (`src/state/load.ts:80`).

No other actionable findings were found.

stevekrenzel:
Round three addressed in `20ad1c9`. Nothing declined.

**Minor 1 — `cd ''` remedy** (`src/commands/uninstall/scope.ts`, `src/state/load.ts`). Reproduced first: a project entry with no `project_root` produced

```
installed at project scope in : cd '' && crew uninstall --scope project demo
```

which is worse than cosmetic, since the remedy is advertised as copyable — pasting it runs the uninstall from the user's home directory. Fixed at the trust boundary rather than the render site: `readState` was casting unvalidated JSON straight to `StateEntry[]`, so `state.json` (user-editable, and possibly predating the field) could introduce an entry that §11.1 forbids. Such entries are now dropped on load, with `doctor --repair` as the documented recovery path. The remedy builder also skips a rootless project entry rather than emitting a broken command — defence in depth, since a bad command is worse than a missing line.

I did not convert `StateEntry` into a scope-discriminated union. That type is referenced in 22 source files and `project_root` is read in 23, so the ripple is far beyond this PR; the invariant is now enforced where untyped data enters instead. Worth its own change if you want it structural.

**Minor 2 — scope documentation** (`README.md`, `site/components/sections/Commands.data.tsx`). Both now say uninstall acts on one scope at a time: the system-wide install by default, the current project's with `--scope project`. The site entry also gained the flag in its signature.

**Minor 3 — `scopeGiven` unread.** Kept, and documented rather than defended again. You flagged this in round one too, so the field's docstring now records *why* it exists: `--scope` is a target for uninstall but a filter for `crew list`, which shows both scopes when the flag is absent and narrows when present. `scope` alone reads as `user` in both cases, so no other signal can express the distinction. It names `src/commands/list/index.ts` as the consumer, which arrives with #118/#124.

**Minor 4 — `RemovalMeta.aborted` unread.** Removed. I checked whether it was load-bearing before deleting: an aborted agent is kept in `remainingAgents`, so the entry takes the survival branch and its root never reaches `fullyRemovedRoots` — which is what actually gates pruning. `aborted: true` and a pushed root are therefore mutually exclusive by construction, so the field could not inform any decision. The invariant is now a comment where the retention happens. Treated differently from `scopeGiven` because no sibling PR consumes it and nothing would restore a use.

**Minor 5 — duplicate identity encoding.** Extracted `src/state/identity.ts` with `entryKey` and `sameLocation`; both uninstall and `update/selection.ts` now use it. I kept the JSON-tuple encoding over the `::`-joined one: a `project_root` is an arbitrary path and can contain any delimiter, so a joined key lets `a::b` + `c` alias `a` + `b::c`. For a value that decides what gets removed, that is worth ruling out by construction.

On convergence: #120's round two added `entryIdentity` for the same purpose with the tap included. At merge these should collapse into this module — the tap-aware variant is the more general one, and both call sites want the same guarantee.

**Minor 6 — `installed_locations` undocumented.** Added to PRD §7.4: an array of `{ scope, project_root }` with `project_root` null at user scope, plus the reason it is not named `installed_at`.

**Minor 7 — single-quote remedy test.** Added. The existing test covered spaces; a single quote is the character that actually breaks naive quoting, since the closing quote lands early and the remainder becomes shell syntax. The test asserts the POSIX close-escape-reopen form and that the raw path is *not* emitted inside plain quotes.

**Nit 8 — naming.** `dropScopedEntryAndUpdateRequiredBy` → `dropInstallLocation`.

**Nit 9 — description.** Corrected to `installed_locations`.

**Verification.** The new malformed-entry test fails against pre-fix source and passes after — I confirmed by reverting both halves of the fix and re-running: exactly that test failed, the other five passed. Targeted suites: uninstall scope/miss/prune/abort 37 pass, dry-run + update-dep-closure + state-subjects 15 pass, 0 fail. `identity.ts`, `scope.ts`, `state.ts` at 100% line and function coverage. Typecheck and Biome clean; every touched file under the 200-line cap.

I re-read `pruneOrphans` before running anything against it, since this is the area of the earlier runaway. Termination is unchanged: `attempted.add(entryKey(orphan))` still runs *before* `removeOne`, and `findOrphan` skips attempted keys, so each entry is visited at most once regardless of outcome. My change only moved `entryKey` to a shared module with an identical encoding.

Two `crew update` tests failed intermittently in batch runs (8s, 20s, 50s). I did not assume flakiness — I stashed this work and reproduced the same failure on unmodified source, where it took 50s. That is the network non-determinism #131 removes, in tests that keep the GitHub-backed default tap.

**Inherited at merge, not duplicated here:** #107's round three fixed `--dry-run` writing `version-check.json` on a TTY, centrally in `isSuppressed` (`src/self-update/notice.ts`). This branch picks that up when the two meet; I did not add a copy.

stevekrenzel:
Rebased onto `main` after the first merge batch; README row conflict only (both rows kept). Uninstall/list tests pass (37).

stevekrenzel:
Rebased onto `main` after the cap-compliance refactors (#135–#143) merged. New tip: `1fbfea6`.

One conflict, re-placed:

- **`site/components/sections/Commands.data.tsx` → `Commands.data/managing.tsx`** — the file was split into one file per command group, so the `uninstall` row edit (the `[--scope {user,project}]` signature and the "one scope at a time" wording) moved into the "Managing skills" group file verbatim.

The `src/cli/args.ts` change (`scopeGiven`) landed cleanly in `src/cli/args/index.ts` — it edits `parseArgs`, which stayed in `index.ts` rather than moving to `tables.ts`.

No behavior changed. `bun run check` is green: typecheck, lint (incl. the new 200-line cap gate), 1112 pass / 0 fail, coverage 100.00 / 100.00. `cd site && bun run lint` clean.

PR #111, opened 2026-09-11
[LOG-5755] fix(install): implement the --from-git flag the PRD defines

stevekrenzel:
## What

`crew install --from-git <url>[@<ref>][//<subpath>]` now works. Before this PR the flag was parsed by `src/cli/args.ts` and silently discarded, so `crew install --from-git https://github.com/acme/skills.git` failed with "needs at least one skill reference".

## Behavior

- The value is always read as a git source. Full URLs, `git@`, `ssh://`, `file://`, and the `gh:`/`gl:`/`bb:`/`@owner/repo` shorthands pass straight through.
- A bare `owner/repo` (optionally `@ref`) is treated as GitHub, i.e. the same as `@owner/repo`. This is the disambiguation purpose PRD §5.3 gives the flag: as a positional, `acme/skills` would be read as `<tap>/<skill>`.
- Anything that doesn't parse as a git source is `invalid_ref` (exit 4) naming the flag.
- Can be combined with positional refs on the same command; `crew install --from-git @acme/skills python-testing` installs both.
- Only `install` accepts the flag. `crew list --from-git x` is now an unknown-flag `usage_error` instead of being silently accepted.

## PRD

- §5.3: expanded the `--from-git` definition (owner/repo → GitHub rule, install-only, `invalid_ref` on unparseable values).
- §18.3: new criterion **C-INST-02b** placed directly after C-INST-02.

## Files

- `src/cli/args.ts` — flag moved from `STRING_GLOBALS` to `STRING_SUB.install`.
- `src/commands/install/from-git.ts` — new normalizer (`normalizeFromGit`).
- `src/commands/install/index.ts` — reads the flag, appends the normalized ref after collision resolution, relaxes the "no refs" usage error when the flag is present.
- `src/commands/help/content/install.ts` — FLAGS entry + example.
- `tests/e2e/install-from-git.test.ts` — six tests, all titled C-INST-02b.

## Testing

Targeted suites: `ref-guards`, `flag-guards`, `tap-root-containment`, and `install-from-git` — 20 pass; the `tapRootDir` consumers (`search`, `tap`, `info`) — 68 pass. Typecheck and Biome clean. The e2e tests use real local git repos over `file://`; the GitHub rewrite is asserted through `parseRef`, so nothing touches the network. README and the site were updated in round two to document `crew install --from-git <source>` and its bare `owner/repo` reading.

Per-suite numbers rather than a headline count: the full suite is non-deterministic on any branch without #131, because the default `core` tap URL is github.com and tests that don't discard it attempt live clones.


stevekrenzel:
## `/review`

Verdict: **not ready — 1 blocker, 2 majors, 3 minors, 0 nits**

### Blocker

**Git subpaths can escape the cloned repository** — `src/commands/install/from-git.ts:17` (acquisition at `src/sources/acquire/index.ts:51`)

The newly documented `//<subpath>` route accepts values such as `file:///tmp/repo//../../../../etc`. The parser preserves `..`, and acquisition joins it to the clone path, normalizing outside the clone before skill discovery. This unsafe parser behavior predates the flag, but the PR explicitly adds and specifies this route.

**Fix:** validate git subpaths centrally: require a normalized POSIX relative path within the repo, reject absolute paths, `..` components, and backslashes, and add a regression through `--from-git`.

### Majors

1. **Bare GitHub values fail when a subpath is present** — `src/commands/install/from-git.ts:19`  
   `acme/skills//tools/demo` and `acme/skills@v1.2.0//tools/demo` hit the generic `//` git rule first; canonicalization of the bare URL fails, so the value is never classified as a tap and never rewritten. Both documented forms return `invalid_ref`.  
   **Fix:** recognize/rewrite the bare `owner/repo` head before generic classification, preserving ref and subpath; add both subpath and ref+subpath tests.

2. **Repeated `--from-git` values are silently discarded** — `src/cli/args.ts:81,113`  
   Yargs returns a `string[]` for duplicate arguments, while `extras` accepts only strings/booleans, so every supplied git source disappears. With a positional ref the command succeeds while ignoring the flags; without one it emits the unrelated “needs at least one” error.  
   **Fix:** reject duplicate scalar flags as `usage_error` (keeping only documented repeatable flags repeatable), and add parser/e2e coverage.

### Minors

- `src/commands/help/content/install.ts:5`: the synopsis still requires a positional `<skill>`, contradicting the new valid `crew install --from-git <source>` form. Express the alternatives.
- Add a real install test pinned to an older local commit/ref, rather than only parsing a rewritten ref, to prove the full flag value reaches acquisition.
- `src/commands/install/from-git.ts:17-19`: avoid classifying the same original value twice; the rewrite fix should parse each candidate once and remove the currently redundant always-true happy-path condition.

### Reviewer coverage

- clarity: 1 finding
- naming: no findings
- type safety: 1 finding
- database: no findings
- testing: 3 findings
- security: 1 finding
- error handling: 2 findings
- observability: no findings
- performance: no findings
- concurrency: no findings
- architecture/conventions: 1 overlapping finding

Deduplicated overlaps are counted once. CI is green, but the blocker and majors prevent approval.

stevekrenzel:
Addressed all six findings in `55380a7`. Nothing skipped; nothing pushed back on.

## Blocker — git subpaths escape the clone

Reproduced first: `parseRef("file:///tmp/repo//../../../../etc")` returned `subpath: "../../../../etc"`, and joining that against a clone at `~/.crew/taps/repo` normalized to `/etc`. An absolute subpath (`gh:a/b///abs`) and a backslash subpath were accepted too.

New `src/refs/subpath.ts` rejects `..` components, leading `/`, and backslashes, and collapses `.` and duplicate separators. It runs inside `parseGit` in `src/refs/git-url.ts`, which is the single choke point for every git-shaped reference — so the guard covers **positional installs, `--from-git`, `crew tap add`, `crew info`, and dependency refs**, not just the flag this PR adds. `tests/e2e/ref-guards.test.ts` proves three of those entry points separately, since each reaches acquisition by its own route.

You were right that §8.4 already called a subpath "any POSIX relative path not starting with `/`" — the code now enforces what the grammar always claimed. I tightened the PRD to name the `..` rejection explicitly rather than leaving it implied, and added C-REF-22a / C-REF-22b. (Suffixed IDs because PRs #113 and #121 take C-REF-23 through 29; this avoids a merge collision.)

## Major 1 — bare owner/repo with a subpath

Confirmed: `acme/skills//tools/demo` and `acme/skills@v1.2.0//tools/demo` both returned `invalid_ref`. They match §8.5's "contains `//` → git" rule first, and canonicalizing the bare `acme/skills` head as a URL fails, so the value never reached the tap branch that does the rewrite.

`src/commands/install/from-git.ts` now recognizes the bare `owner/repo` head *before* generic classification, preserving both tails. Tests cover subpath-only and ref+subpath.

## Major 2 — repeated `--from-git` silently discarded

Confirmed, and it is broader than the report. `duplicate-arguments-array` makes yargs return an array, which `extras` drops — but `--scope user --scope project foo` also silently fell back to `user`, and `--interval 1h --interval 2h` vanished entirely. Every one of those loses a choice the user stated explicitly.

`rejectRepeatedScalars` in `src/cli/args.ts` now fails with a `usage_error` naming the flag. It is keyed off `ARRAY_GLOBALS`, so it covers every current and future scalar flag automatically, and `--agent` stays repeatable. PRD §5.2 states the rule globally; C-CLI-08a covers it.

## Minors

- **Synopsis** (`src/commands/help/content/install.ts`): now `crew install <skill> [<skill>...] | crew install --from-git <source>`.
- **Real pinned install test**: added one against a two-commit `file://` repo with the skill in a subfolder, asserting the recorded `ref` and `pinned`. Only the subpath could have selected `demo` over `later`, so the whole flag value demonstrably reaches acquisition. One caveat: I deliberately did **not** assert `resolved_sha === first`, because on this base it records HEAD instead — that is the pre-existing bug #129 fixes on its own branch, and asserting it here would encode the defect. There is a comment in the test saying so.
- **Double classification**: the rewrite parses each candidate once; the always-true happy-path condition is gone.

## Pre-fix verification

Stashed `src/` only, restored the tests, and ran them against the pre-fix commit: **all 9 new tests fail**, covering the blocker at each entry point, the bare-head rewrite, and the duplicate-flag guard. Restored afterward.

## Note

`tests/e2e/install-from-git.test.ts` would have hit 226 lines, over the hard cap, so the containment and flag-guard cases moved to `tests/e2e/ref-guards.test.ts` (166 and 107 lines). They are a separate concern from the flag itself anyway.

`bun run check`: 1002 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean.

stevekrenzel:
## Updated `/review`

Verdict: **not approved — 1 blocker, 1 major, 2 minors**

### Blocker

- **Lexical subpath checks still allow symlink escape** (`src/refs/subpath.ts:28`, `src/sources/acquire/index.ts:51`, `src/skill/load.ts:44`). A cloned repository can contain `outside -> /absolute/local/skill`; `repo//outside` passes validation and later discovery/install follows it outside the clone. Reject symlink components or resolve the real path and prove it remains beneath the acquired clone before any filesystem operation.

### Major

- **The new scalar-repeat rule misses boolean flags** (`src/cli/args.ts:150`, `tests/e2e/ref-guards.test.ts:91`). Yargs collapses repeated booleans to `true`, so `--json --json`, `--force --force`, and `--recursive --recursive` silently succeed despite §5.2/C-CLI-08a saying only `--agent` may repeat. Detect occurrences from raw argv or preserve counts in parser configuration, and add boolean regression cases.

### Minors

- README and the site still present positional refs as the complete install syntax, although `--from-git` is now a valid no-positional form. Document `crew install --from-git <source>` and its bare `owner/repo` interpretation.
- The comment at `tests/e2e/install-from-git.test.ts:132` relies on external PR #129 context to justify omitting `resolved_sha`. State the local test boundary directly: this suite verifies forwarding/ref metadata while dedicated pinned-ref tests cover SHA resolution.

All eleven lenses reran against updated head `55380a7`; CI is green. I did not approve because a blocker and major remain.

stevekrenzel:
Addressed all four findings in `4a562da`. Nothing declined.

## Blocker — lexical subpath checks still allow symlink escape

Reproduced before fixing. A repo committing `evil -> /tmp/.../outside-skill`, installed via `file://<repo>//evil`:

```
Installing pwned (dry run)
  pwned
    content from outside the clone entirely
    [ok] claude-code → ~/.claude/skills/pwned
    ...
8 installs
EXIT: 0
```

Eight reported installs of a skill whose bytes never existed in the repository, exit 0, no warning. You were right that the round-one guard is necessary but not sufficient: `evil` is a lexically perfect relative path, and only the filesystem knows it isn't one.

New `src/util/symlink-containment.ts` walks every segment from the source root to the resolved location with `lstat` and rejects any symlink — the leaf *and* every directory above it, since a symlinked parent redirects just as effectively. It runs in `acquireTap` (`src/sources/acquire/index.ts`), which is the single point where a subpath stops being a string and becomes a real location, so it covers positional installs, `--from-git`, `crew tap add`, `crew info`, and dependency refs alike. Same reference now:

```
Error (invalid_ref)
  tap `repo-evil` subpath `evil` resolves through a symlink, which would
  read content from outside the source
EXIT: 4
```

**On reusing #129's helper:** I matched its shape deliberately rather than inventing a second approach — same walk-every-segment-with-`lstat` structure, same "a symlinked parent redirects just as effectively" rationale, same error-detail shape naming the offending segment. The two differ only in what they're anchored to (`assertNoSymlinkEscape(exportRoot, …)` for an exported commit tree there, a clone root here) and in error name, since a bad reference is `invalid_ref` while a bad exported tree is `invalid_skill`. **Merge note:** these should collapse into one helper when the branches meet; #129's version is the more general of the two and mine can be dropped in favour of it with a parameterised error name.

**Scope note:** I did *not* extend this to path-kind taps. A path tap points at a directory the user named directly, so there is no "inside the source" boundary to escape — that would be a different finding with a different rationale, and I'd rather not widen a security fix on inference.

## Major — the scalar-repeat rule misses boolean flags

Reproduced: `--json --json`, `--force --force`, `--recursive --recursive`, and `--dry-run --dry-run` were all accepted, while `--scope user --scope project` was correctly rejected. The cause is that the two flag kinds leave different evidence. A repeated *value* flag becomes an array thanks to `duplicate-arguments-array`; a repeated *boolean* collapses to plain `true` and leaves no trace at all in the parsed result, so no amount of inspecting yargs' output can see it. `rejectRepeatedScalars` now also counts occurrences in raw argv.

Care taken not to over-reject: `--agent a --agent b` still accumulates, `--scope=user` (the `=` form) is handled, and tokens after `--` are not treated as flags. All four covered in the new test.

**Merge note:** #114, #116, #117, #124, and #130 all touch `src/cli/args.ts`, and several ported this guard verbatim. I extended the existing function rather than adding a parallel one, so whichever lands first should keep its copy and take this argv loop as the delta.

## Minors

- **README and site** now document `crew install --from-git <source>` as a valid no-positional form, including the bare `owner/repo` → GitHub interpretation.
- **Test comment** no longer leans on external PR context. It now states the local boundary: this suite proves the flag's value is forwarded intact and its ref metadata recorded, while resolving a ref to the installed bytes is a separate concern with its own tests.

## PRD

§8.4 gains the symlink rule (the lexical check is explicitly called out as insufficient, with the `outside` example) plus criterion **C-REF-22c**. §5.2 and **C-CLI-08a** now state that the repeat rule covers booleans, and *why* an implementation must read raw arguments to enforce it — that reason is not obvious from the requirement alone, and a second implementation would hit the same trap.

## Verification

Both new tests fail against the pre-fix commit, one per finding (verified by stashing only `src/`):

```
(fail) C-REF-22c a committed symlink subpath cannot escape the clone
(fail) C-CLI-08a a repeated boolean flag is a usage_error
```

Targeted suites: `ref-guards` + `install-from-git` 15 pass / 0 fail; parser-adjacent suites (`coverage`, `final-gaps`, `project-scope`) 68 pass / 0 fail. Typecheck and Biome clean. `src/util/symlink-containment.ts` at 100% line and function coverage — I removed a `try`/`catch` around `lstatSync` that could not fire, since callers verify the location exists first, rather than writing a test for an unreachable branch. All touched files under the 200-line cap.

Per-suite numbers rather than a headline count, because the full suite is non-deterministic on any branch without #131 (the default `core` tap URL is github.com, so tests that don't discard it attempt live clones).

stevekrenzel:

**Verdict:** 0 blockers, 0 majors, 7 minors. **Approved on substance.**

Reviewed `6428d43..7ee5935` through all 11 `/review` lenses. CI is green at this head.

### Minors

1. Help and site render positional refs and `--from-git` as alternatives even though they can be combined (`src/commands/help/content/install.ts:5`, `site/components/sections/Commands.data.tsx:14`). Use a synopsis that shows both as optional sources while requiring at least one.
2. Persisted `tap.subpath` is joined before applying the lexical grammar (`src/sources/acquire/index.ts:85-88`). Values such as `a/../b` and backslashes survive even though §8.4 rejects them. Run `normalizeSubpath` on stored values before resolving them.
3. A tap with no explicit subpath still allows standard-layout directories such as `skills/` to be symlinks, so discovery can follow a repository symlink outside the clone (`src/sources/acquire/index.ts:84`, `src/sources/expand.ts:71-74`). Enforce the same no-symlink rule on discovery paths; this is a pre-existing gap exposed while hardening containment here.
4. The new containment check is check-then-use: another tap checkout can replace a checked segment with a symlink before indexing or staging reads it (`src/util/symlink-containment.ts:62`, `src/sources/acquire/index.ts:58`). Protect validation and consumption with the tap lock/snapshot work so the path cannot change between them.
5. `lstatSafe` rethrows `ENOTDIR` and other filesystem errors as raw errors (`src/util/symlink-containment.ts:85`). A subpath below a regular file now becomes generic `usage_error`; translate expected path-shape failures to the documented `CrewError`.
6. The repeatable-`--agent` guard test only checks exit 0, which would pass if the parser retained a single value (`tests/e2e/flag-guards.test.ts:76`). Assert the parsed array or both resulting adapter owners.
7. Add a persisted malicious-subpath case through `crew info`, complementing the existing search/config and newly parsed info cases (`tests/e2e/tap-root-containment.test.ts:49`).

No release-holding issue remains. The two earlier majors and their focused containment cases are fixed at this head.

stevekrenzel:
Round three addressed in `3a585b8`. One finding refuted with evidence; everything else fixed.

## Major 1 — refuted

The finding says `from-git` sits in `STRING_GLOBALS`, is therefore swept into `BUILT_IN_FLAGS`, gets stripped from `extras` before `installCommand` reads it, and is silently accepted on unrelated commands. I checked all three claims against this head and none of them hold. Round one moved the flag to `STRING_SUB.install` precisely to avoid this.

`src/cli/args.ts:33` and `:43-48`:

```ts
const STRING_GLOBALS = ["scope"] as const;
...
const STRING_SUB: Record<string, readonly string[]> = {
  autoupdate: ["interval"],
  // `--from-git <url>` is an explicit git source (§5.3); only install takes it.
  install: ["from-git"],
  "self-update": ["version"],
};
```

`BUILT_IN_FLAGS` is built from `BOOLEAN_GLOBALS`, `STRING_GLOBALS`, and `ARRAY_GLOBALS`, so `from-git` was never in it and the extras loop cannot remove it. Parser-level:

```
install extras[from-git] = "acme/skills"
list: rejected -> Unknown argument: from-git
```

End to end, the value reaches acquisition with the bare `owner/repo` correctly rewritten:

```
$ crew install --from-git acme/does-not-exist-xyz --dry-run
Error (source_unreachable)
  couldn't clone `https://github.com/acme/does-not-exist-xyz.git` — ...

$ crew list --from-git acme/skills
Error (usage_error)
  Unknown argument: from-git
```

I have not changed working code here. I did add **nit 2**, a parser assertion pinning both halves, so the behaviour is nailed down either way — `C-INST-02b --from-git reaches install and is rejected elsewhere` in `tests/e2e/flag-guards.test.ts`.

## Major 2 — confirmed, both gaps, and they were worse than described

**(a) Containment was never actually checked.** `assertNoSymlinkEscape` walked upward while `current.length > rootAbs.length` and rejected symlinks, but nothing proved the target was beneath the root to begin with. A target outside the root simply walked past it and returned clean:

```
target below root lexically? false
(a) RESULT: ACCEPTED  <-- escape not caught
```

No symlink anywhere, so the symlink walk had nothing to catch. The helper now proves containment with `relative()` first, then walks for symlinks — the two checks cover different attacks and both are required.

**(b) The check sat at one call site instead of the boundary.** Only `acquireTap` called it. `tapRoot` in `src/install/tap-index.ts` resolves a tap root through `tapRootDir` for `crew search` and `crew info` without it. Both now go through `tapRootDir`, which performs the check as it resolves — install's separate call was removed as redundant rather than left duplicated.

That matters most for subpaths that **never pass through the reference grammar at all**: `subpath` is persisted in `config.yaml` and read straight back. A hand-edited entry, a row written before the guard existed, or a directory that became a symlink after `crew tap add` all reach discovery unvalidated. `tests/e2e/tap-root-containment.test.ts` covers both, writing the config directly.

**One thing the tests found that the review didn't.** Search wrapped indexing in a bare `catch` and reported every failure as `tap ... isn't cloned yet and couldn't be reached — run crew tap update when you're back online`. Treating an unusable tap as a warning is correct per §16.6, but telling a user to retry when online never fixes a subpath that escapes its clone. The two are now distinguished, and the test asserts the message does *not* say `back online`.

## Verification

All four containment tests fail with the check removed from `tapRootDir` and pass with it:

```
(fail) C-REF-22c a committed symlink subpath cannot escape the clone
(fail) C-REF-22c a symlinked PARENT directory cannot escape the clone
(fail) C-REF-22a a hand-edited `..` subpath is skipped, not indexed
(fail) C-REF-22c a subpath that became a symlink is skipped, not indexed
 7 pass, 4 fail
```

## Minors and nits

1. **Symlinked parent** — added; the leaf is an ordinary directory and only the parent is the link, so a leaf-only check passes it.
2. **`crew info` entry point** — added.
3. **Help label** — `--from-git <url>` → `<git-source>`, since it takes shorthands and bare `owner/repo` too.
4. **`acquireTap` docstring** — now names `invalid_ref` alongside `source_unreachable` and `no_skills_found`.
5. **Nit 1** — the description's claim that README and the site were untouched was stale; round two updated both. Corrected, along with the stale test count.
6. **Nit 2** — the parser assertion described above.

`ref-guards.test.ts` reached 205 lines with the new cases, over the hard cap, so the flag guards moved to `flag-guards.test.ts` (reference guards 151, flag guards 84). They were already separate concerns sharing a file.

## PRD

§8.4 gains the rule that containment is re-verified wherever a stored subpath is resolved, not only where a reference is parsed, plus the requirement that a warning-only command names a containment refusal for what it is. New criterion **C-REF-22d**.

## Inherited, not duplicated

#107's round three fixed centrally, in `isSuppressed`, that `--dry-run` still wrote `version-check.json` on a TTY. Not in this base; this branch inherits it at merge rather than carrying a copy.

Targeted suites: 20 pass across the four guard/flag files, 68 across the `tapRootDir` consumers (`search`, `tap`, `info`), 0 fail. `symlink-containment.ts` and `search/configured.ts` at 100% line and function coverage. Typecheck and Biome clean; every touched file under the 200-line cap.

stevekrenzel:
CI regression from `3a585b8` fixed in `7ee5935`.

**What broke.** Round three's containment walk called `lstatSync` on every segment between the target and the source root, with a comment asserting that callers had already proven the target exists. That assumption does not hold for the ordinary case of a skill deleted upstream. The segment is gone, `lstatSync` raises a bare Node `ENOENT`, and since that is not a `CrewError` it bypasses the soft-outcome branch in `update/entry.ts` and lands in the hard-failure branch. Two tests caught it:

```
(fail) update: missing skill at new revision > C-UPD-11 SKILL.md deleted upstream → source_gone soft outcome
(fail) update edge cases > skill removed from tap upstream → source_gone

Received: "Checked 1 skill\n\n  [fail] demo  failed  ENOENT\n[warn] tap …-demo (ENOENT)\n\n1 failure\n"
```

§10.1 requires that case to be the soft `source_gone` outcome with the install preserved, so this was a real behavioural regression, not a test artifact.

**Why.** Containment answers *where* a path points, not *whether* it exists. A missing path points nowhere, and a missing segment cannot be a symlink, so treating absence as an escape conflated two unrelated conditions.

**The fix.** `lstatSafe` absorbs `ENOENT` only. A permission or I/O failure still throws, because a containment check that silently passes on a segment it could not inspect would be worse than no check at all.

**Containment still holds** — confirmed, not assumed. Both round-three guarantees pass unchanged, and I verified they are still load-bearing:

- a hand-edited `..` subpath is still refused (`outside the source`),
- a subpath that became a symlink is still refused (`symlink`),
- a *missing* target that also escapes lexically is still refused — absence is not a way around the lexical check,
- a segment that exists but cannot be inspected still throws.

**On the new tests.** My first attempt pinned this through `crew search` and **passed with the bug still present**, because §16.6 converts indexing failures into warnings and swallowed the difference. I dropped it and asserted the helper directly instead. Verified load-bearing: reverting only the source change fails `a missing but contained target is accepted` while the two guarantee tests keep passing, which is the pairing that matters here.

**Verification.** `update-edges` + `error-paths` + `ref-guards` + `flag-guards` + `tap-root-containment`: 29 pass, 0 fail. Typecheck and Biome clean — note the formatter flagged my import placement, the same drift that is currently blocking #112's CI, so it is worth running `lint` before pushing on these branches.

PR #112, opened 2026-09-11
[LOG-5756] fix: honor --dry-run on agents, autoupdate, cache clean, and doctor --repair

stevekrenzel:
## What

PRD §5.2 says `--dry-run` "describes what would happen without changing anything" and is accepted on any command where it applies. Four mutating commands accepted the flag and mutated anyway. This PR makes each one a real preview:

| Command | With `--dry-run` |
|---|---|
| `crew agents enable\|disable <name>` | Validates the name, reports "Would enable/disable", writes no config. Unknown name is still `usage_error`. |
| `crew autoupdate enable\|disable` | Names the scheduler files it would write and load (or unload and remove) and the interval. No launchctl/systemctl call, no config/plist/unit write. Unsupported platform still fails with `autoupdate_failure`. |
| `crew cache clean` | Reports bytes and orphan store entries that would be freed. Deletes nothing. |
| `crew doctor --repair` | Runs the checks, lists the findings a repair would address, applies none. Exit code follows the plain `doctor` rule because nothing was fixed. |

All four carry `dry_run: true|false` in `--json` output.

Sibling PRs handle `install` (already correct), `uninstall`, `update`, and `tap`.

## Why

A user who types `--dry-run` is explicitly asking not to change anything. Silently mutating on a preview is the worst possible outcome for the flag.

## PRD sections touched

- §6 (`cache clean --dry-run`)
- §7.2 (`agents enable|disable --dry-run`)
- §10.2 (`autoupdate enable|disable --dry-run`)
- §11.2 (`doctor --repair --dry-run`)
- §18.3 criteria: C-AGENT-09, C-AUTO-11, C-STATE-12, C-STATE-13 (added inside their existing blocks)

Help pages for all four commands list the flag and show an example.

## Implementation notes

- `src/commands/autoupdate.ts` became `src/commands/autoupdate/{index,render,duration}.ts` to stay under the 200-line cap. Test imports of `parseDuration` updated.
- `src/autoupdate/scheduler.ts` gains `autoupdateArtifacts(home)` so the dry-run can name platform-specific files without duplicating path logic.
- `src/maintenance/gc.ts` splits `orphanStoreEntries` out of `garbageCollectStore`, so the preview and the real clean measure exactly the same set.
- `renderDoctor` takes a `dryRun` option; the dry-run footer reports how many findings a repair would address.

## How tested

New `tests/e2e/housekeeping-dry-run.test.ts` (15 tests) runs each command through `runCli` in a scratch `CREW_HOME`, using the `setLaunchctlRunner` / `setSystemctlRunner` seams and platform override for autoupdate. Every dry-run test asserts both the preview output and that config, scheduler files, store, and cache are byte-identical afterward; each real-run counterpart asserts `dry_run: false`.

`bun run check` is green: 1000 tests, 100% line and function coverage, typecheck and lint clean.


stevekrenzel:
## `/review`

Verdict: **not ready — 0 blockers, 1 major, 5 minors, 0 nits**

### Major

**Doctor dry-run claims it would repair findings the real repair path cannot fix** — `src/commands/doctor/render.ts:149`

`repairableCount` counts `missing_project_root`, `autoupdate_not_loaded`, and `autoupdate_unexpectedly_loaded` as findings repair would address. The actual `repairState` path rewrites config/state and garbage-collects the store; it does not repair missing project roots or scheduler drift. The preview therefore overpromises, and a real repair can report success while those findings remain.

**Fix:** exclude these codes from the repairable count until repair implements them, or add the corresponding repair behavior. Test preview and real repair against each category.

### Minors

- `src/commands/help/content/autoupdate.ts:21`: help says `--json` is only for `status`, but enable/disable also emit JSON and this PR adds their `dry_run` fields. Change it to “Machine-readable output.”
- `tests/e2e/housekeeping-dry-run.test.ts:1`: the new 262-line file violates the repository-wide hard `< 200` line cap. Split the four command suites into focused files/a directory.
- Fresh-home preview tests should assert `state.json` remains absent, protecting the no-write boundary from accidental `withStateLock` use.
- C-STATE-12 coverage exercises only orphan-store cleanup. Snapshot state/config/store bytes for marker/state drift, tap reconstruction, and non-repairable findings so the full conformance promise is tested.
- Complete autoupdate coverage: assert disable’s exact artifact/JSON output on Linux, and verify repeated enable/disable make zero scheduler calls and leave config unchanged.

### Reviewer coverage

- clarity: 2 findings
- naming: no findings
- type safety: no findings
- database: no findings
- testing: 4 findings
- security: no findings
- error handling: 1 finding
- observability: no findings
- performance: no findings
- concurrency: no findings

Overlapping test recommendations are grouped in the minor list. CI is green, but the major prevents approval.

stevekrenzel:
Addressed all six findings in `fc7f465`. Nothing pushed back on — the major reproduced exactly as described.

## Major — doctor dry-run overpromised

Reproduced first: a lone `missing_project_root` finding rendered `Dry run: crew doctor --repair would address 1 finding`, and `repairState` provably cannot fix it — check 8 in `checks.ts` documents that removing a vanished project's install "isn't doctor's job."

The root cause was that `repairableCount` used a **blocklist** (`NOT_REPAIRABLE`), so any code not explicitly excluded was assumed fixable. That is backwards: a newly added finding code silently claims to be repairable until someone remembers to exclude it.

`src/commands/doctor/render.ts` now uses `REPAIRABLE_CODES`, an allowlist mapping each repairable code to the mechanism that fixes it. A new code is not repairable until someone adds the repair and lists it.

I also found and fixed the second half of the finding, which the review alluded to ("a real repair can report success while those findings remain"):

- **The post-repair summary counted every finding as addressed.** `src/commands/doctor/render.ts` now reports `N findings addressed` plus, when any remain, `M findings left for you — rerun crew doctor`.
- **The exit code was unconditionally 0 after `--repair`.** `src/commands/doctor/index.ts` now keeps a non-zero exit when an *unrepairable, error-level* finding survives the repair. Repairable findings still stop counting, since the repair genuinely resolved them.

### Interaction with #123 — please read before merging either alone

#123 (`fix/doctor-repair-autoupdate`) is stacked directly on this branch and implements repair for `autoupdate_not_loaded` and `autoupdate_unexpectedly_loaded`. So of the three codes the review named, only `missing_project_root` is permanently unrepairable.

I therefore **dropped `missing_project_root`** from the repairable set and **kept the two autoupdate codes**, with a comment pointing at the implementation. Consequences:

- **Merging #112 then #123** (the intended order): correct throughout. #123 adds the repair that the set already promises.
- **Merging #112 alone**: the two autoupdate codes must be removed from `REPAIRABLE_CODES`, or the preview will overpromise on scheduler drift exactly the way this review describes.

I did not implement autoupdate repair here, since that would collide with #123 head-on.

PRD §11.2 now states the unrepairable set and the exit-code rule, with criterion **C-STATE-12a** (suffixed to avoid colliding with #123's C-STATE-11a/b/c).

## Minors

- **`--json` help** — `src/commands/help/content/autoupdate.ts` now says "Machine-readable output." Confirmed enable/disable both emit JSON with `dry_run` (`autoupdate/index.ts:64,95`).
- **262-line test file** — split into `tests/e2e/housekeeping-dry-run/` with one file per command plus a shared `helpers.ts`: 63, 171, 54, 170, and 13 lines. All under the cap.
- **Fresh-home `state.json` assertions** — added to the agents and autoupdate previews, and to cache's fresh-home test. This is the same hazard #107 and #109 hit, where taking the state lock creates `state.json` via `ensureDir` + `touch`. These four commands never took the lock, so the assertions pass both before and after this commit; they are regression guards, not proof of a present bug, and I would rather say that plainly than imply they caught something.
- **C-STATE-12 coverage** — added a test that installs a real skill, orphans its state entry and adds an orphan store entry, then snapshots state, config, and every file under `store/` and asserts byte-equality after the dry run. It then runs the real repair to prove the dry run had genuine work available to skip.
- **Autoupdate coverage** — added Linux `disable --dry-run` asserting the exact JSON (`{enabled: false, artifacts: [service, timer], dry_run: true}`) and that both unit files are unchanged, plus a test running enable/disable/enable previews in sequence asserting zero scheduler calls and byte-identical config and plist.

## Pre-fix verification

Stashed `src/` only, kept the new tests:

- `C-STATE-12a a non-repairable finding is excluded from the preview count` — **fails** pre-fix (reports 2 findings instead of 0).
- `C-STATE-12a a real repair reports only what it addressed and leaves the rest` — **fails** pre-fix (claims "3 findings addressed", no remainder line).

The fresh-home and byte-snapshot tests pass both ways, as noted above.

One honest note on process: my first versions of both regression tests asserted the wrong counts, because deleting a project directory also orphans its marker and store entry. The implementation was right and my fixture was naive; I rebuilt the fixtures to isolate `missing_project_root` rather than adjusting the expectations to match whatever the code printed.

`bun run check`: exit 0, **1006 pass**, 0 fail, **100%** line and function coverage.

stevekrenzel:
## Updated `/review`

Verdict: **not approved — 3 blockers, 1 major, 1 minor**

### Blockers

- **Autoupdate drift is falsely reported as repaired** (`src/commands/doctor/render.ts:161`, `src/commands/doctor/index.ts:56`). Scheduler findings are allowlisted as repairable even though no scheduler reconciliation runs. Implement reconciliation and derive summary/exit status from actual repairs, or remove those codes and the PRD promise.
- **`doctor --repair` throws after detecting invalid config** (`src/commands/doctor/index.ts:56`, `src/commands/doctor/repair.ts:25`). Config parsing is first converted into a nonrepairable finding, then repair reads config again and throws instead of returning doctor-style results. Skip repair when config is unavailable and preserve the finding/nonzero response.
- **Repair deletes state for a missing project root** (`src/commands/doctor/repair.ts:52`). `missing_project_root` is defined as nonrepairable and requires user action, yet the state entry is removed as unbacked. Preserve it and add a state-retention regression.

### Major

- `tests/e2e/housekeeping-dry-run/doctor.test.ts` is 202 lines, exceeding the repository’s hard sub-200-line file cap. Split or extract shared scenarios.

### Minor

- Cache preview coverage (`tests/e2e/housekeeping-dry-run/cache.test.ts:14`) contains only an orphan store entry. Add a live referenced store entry and assert it is excluded from `removed_store` and `freed_bytes`.

All eleven lenses reran against updated head `fc7f465`; CI is green. I did not approve because blockers and a major remain.

stevekrenzel:
Addressed all five findings in `8b47a38`. Nothing declined — all three blockers reproduced first.

## Blocker 1 — autoupdate drift claimed as repaired

Reproduced: with the scheduler loaded and config saying disabled, `crew doctor --repair` printed `1 finding addressed`, and a follow-up `crew doctor --json` still showed `autoupdate_unexpectedly_loaded`. The allowlist entry pointed at `repairAutoupdateDrift`, a function that does not exist on this branch.

**Decision: removed both scheduler codes from `REPAIRABLE_CODES` here.** Round one kept them on the reasoning that #123 implements the reconciliation, so #112 would never merge alone. You're right that this is unsafe — either PR can land first, and the one that lands alone is wrong.

With this change the two branches are each honest about themselves:

- **#112 alone**: reports scheduler drift, never claims to fix it. `0 findings addressed` / `1 finding left for you`.
- **#123 on top**: adds `repairAutoupdateDrift` *and* re-adds the two codes in the same change, so the promise and the implementation arrive together.

I did not implement the reconciliation here; that is #123's diff and duplicating it would collide head-on.

**PRD §11.2 was promising it too**, which the review didn't name but is the same defect one layer up — a second implementation reading the spec would have built the reconciliation and found nothing to match. The bullet is gone from the `--repair` list, replaced by an explicit statement that check 7 is reported and not reconciled. #123 restores the stronger promise alongside its code.

## Blocker 2 — `--repair` threw on invalid config

Reproduced: exit 4, empty stdout, and a raw `config_invalid` CrewError on stderr. The finding was raised correctly and then thrown away, because `repairState` re-reads config and lets the parse error escape.

`src/commands/doctor/index.ts` now gates the repair on config being readable. The run reports the finding as a normal doctor report at exit 1, and `renderDoctor` takes an explicit `applied` flag so it can't print "Repaired what was fixable" for a repair that never ran.

The test asserts the config file is byte-for-byte unchanged afterward. That matters specifically here: repair *rewrites* `config.yaml` to reconstruct taps from markers, so running it against a file we failed to parse risks overwriting whatever the user was mid-edit on.

## Blocker 3 — repair deleted state for a missing project root

Reproduced: one project-scope entry whose root had been removed; after `--repair`, state was empty.

`buildMarkerIndex` skips a root that doesn't exist, so the entry has no marker and the orphan filter drops it. But that's "marker unread", not "marker absent" — §11.2 check 8 explicitly defines a vanished root as the user's to resolve, so the entry is the only record they have left. `repair.ts` now preserves it, and the PRD says so directly.

## Major — file-size cap

`doctor.test.ts` was 202 lines. Split by concern: preview behaviour stays in `doctor.test.ts` (100), real-repair behaviour moves to `doctor-repair.test.ts` (131), and the shared scheduler/adapter pinning plus the project-entry fixture move into `helpers.ts` (84) where the other suites in this directory can use them.

## Minor — cache preview with a live store entry

This one found a real bug rather than just a coverage gap.

I added the referenced entry the review asked for, and the test failed: `cache clean --dry-run` listed the *live* skill's store directory in `removed_store`. `referencedStoreEntries` skipped any entry with `resolved_sha === null` — its own comment said such entries are "kept unconditionally" — but a path source keys its store directory by the first 8 chars of the content hash, which state records in `content_hash`. Nothing ever added that key, so every path-source install looked orphaned.

So `crew cache clean` would have deleted the store entry backing a path-installed skill. Not fatal (the next install re-stages it), but it makes the preview lie about what's safe to remove. The referenced set now reconstructs both key shapes.

## One thing beyond the findings

With the scheduler codes gone, an unrepairable **error**-level finding fell through to the footer "These are heads-ups, not errors — crew keeps working." That's wrong for `config_invalid`, which is exactly the error a user needs to act on. Added a branch that says they need attention instead.

## Verification

Blocker regressions fail pre-fix — stashing only `src/`, the three new tests (`C-STATE-12b`, `C-STATE-12c`, `C-STATE-12d`) fail and the rest pass.

Targeted suites: **83 pass, 0 fail** across `housekeeping-dry-run/`, `doctor-render`, `doctor`, `cache-format`, and `coverage`. `repair.ts`, `gc.ts`, and `index.ts` at 100% line and function coverage; `render.ts` at 99.21%, the one uncovered line being the pre-existing large-cluster summary that other suites reach. Typecheck and lint clean (the two `biome.json` schema infos are present on the base too).

I did not quote a full-suite number. Two failures I saw in `tests/misc/update-edges.test.ts` are the known non-determinism without #131 — I confirmed it by stashing my changes entirely and running that file three times on the unmodified base, which gave 2, 3, then 4 passes out of 4.

Loop safety: the paths I changed are bounded filters over state and a fixed walk of adapters and roots, with no recursion whose termination depends on my edits. I re-read them before running anything against them.

stevekrenzel:

**Verdict:** 0 blockers, 0 majors, 6 minors. **Approved on substance.**

Reviewed `6428d43..c4fd00f` through all 11 `/review` lenses. CI is green at this head.

### Minors

1. A missing project root produces `state_entry_without_marker`, but repair deliberately preserves that unreadable entry; the generic code remains classified as repaired, so `doctor --repair` can return 0 and say it addressed a finding that appears again immediately (`src/commands/doctor/repair.ts:75`, `src/commands/doctor/repairable.ts:31`, `src/commands/doctor/index.ts:69`). Exclude unreadable roots from marker-drift matching or keep this instance unrepairable.
2. Autoupdate help still says `doctor --repair` can fix scheduler drift, while this PR explicitly makes that finding non-repairable (`src/commands/help/content/autoupdate.ts:40`). Point users only to disable/re-enable.
3. The new dry-run options are absent from the README/site command summaries even though those surfaces list the affected commands. Add concise preview examples or flags where needed.
4. `cache clean --dry-run` takes no lock, and `dirSize` now lets `ENOENT` escape if a concurrent real clean deletes a file between walk and stat (`src/commands/cache.ts:100`). Ignore only disappearance races while continuing to surface permission/I/O failures.
5. Strengthen the invalid-config dry-run test to compare the exact file contents, and add command-level repairability coverage for `customized`, `agent_missing`, and `config_invalid` (`tests/e2e/housekeeping-dry-run/doctor-repair.test.ts:82`, `doctor.test.ts:64`).
6. Add the regular-file `project_root` case to the new unreadable-root preservation coverage (`tests/e2e/housekeeping-dry-run/doctor-repair.test.ts:63`).

No release-holding issue remains. The previous CI, path-store, repair-policy placement, and filesystem classification findings are fixed at this head.

stevekrenzel:
Round three addressed in `c4fd00f`. All nine reproduced first; nothing declined, though one was resolved by deletion rather than the suggested guard — details below.

**Blocker — CI failing.** Reproduced on a clean checkout: `bun run lint` reported formatter drift in one hunk of `tests/e2e/housekeeping-dry-run/doctor-repair.test.ts`, a `writeState(...)` call hand-split across four lines that Biome wants on one. `bun run format` fixed it. Worth naming the cause: the round-two change was pushed without running lint, and `bun run check` would have caught it in a second. I hit the same thing myself mid-task after a stash restore, and re-ran lint before pushing rather than assuming.

**Minor 1 — two orphan-store predicates.** Confirmed, and it was the detection half of the same path-source bug round two fixed in the collector. `checkOrphanStoreEntries` keyed only on `resolved_sha`, so a path-source install — whose store directory is keyed by its short content hash — was reported orphaned, and a repair then claimed to have addressed an entry the collector correctly preserved. Detection now delegates to `orphanStoreEntries`, so the two cannot drift. That also picked up a case the finding did not mention: the collector only removes directories, so a stray file in the store root was being announced as a cleanup that never happens.

Three regressions in `tests/e2e/housekeeping-dry-run/path-source-store.test.ts`. Two fail against the old predicate; the third asserts the collector's behaviour directly and passes either way, so I am not counting it as a caught regression.

One correction to my own work: my first version of that third test drove it through `doctor --repair` and failed, because repair rebuilds state from on-disk markers and my fixture had none — so the rebuilt state legitimately dropped the entry. That is the marker-authority rule (§11.1), not drift. I fixed the test's premise rather than the code, and left the reason in a comment.

**Minor 2 — stale repair claims.** Both confirmed. `crew help doctor` said `--repair` would "reconcile scheduled updates" and `autoupdate status` offered `crew doctor --repair` as a fix. Neither is true on this branch, which reports scheduler drift without reconciling it. Both now point at `crew autoupdate disable` then `enable`. The comment in `autoupdate/render.ts` records why `--repair` is deliberately not offered, so it is not re-added by someone reading only the status text.

**Minor 3 — preview did not say which findings were repairable.** The dry run printed "would address N findings" without marking them, so with a mixed set the user could not tell what would be left behind. Each cluster is now tagged `— would be repaired` or `— needs you` during a repair preview only; a plain `crew doctor` does not mark anything, since the question was not asked. Two tests pin both halves.

**Minor 4 — project root replaced by a file.** Confirmed on both sides: `checks.ts` and `repair.ts` each used `existsSync`, so a path replaced by a regular file read as usable, and the finding message even says "directory". Both now test for a directory. Keeping them identical matters here — if detection and repair disagreed, repair could drop the state entry for a root it cannot actually read.

**Minor 5 — `dirSize` swallowing every stat failure. Resolved by deletion, not a narrowed guard.** I first implemented the suggested `ENOENT`-only filter, then could not reach the rethrow: `walk` classifies entries with `lstat` and excludes symlinks from `isFile`, so every path reaching the `statSync` was a regular file a moment earlier, and the only way to force a non-`ENOENT` failure is a permission error — which means making a directory read-only, a technique that hung four processes for another worker on this stack. An unreachable defensive branch is dead code by CLAUDE.md's rule, so the handler is gone and `statSync` is unguarded, with a docstring recording why a failure should bubble rather than be absorbed into a wrong byte count. This *improves* coverage: line 92 was already uncovered before this change and no longer exists.

**Minor 6 — policy in the renderer.** Moved `REPAIRABLE_CODES`, `isRepairableCode`, `repairableCount`, and `isRepairable` into `src/commands/doctor/repairable.ts`, next to the repair mechanisms the allowlist describes. `index.ts` no longer imports a behavioural question from the presentation layer. (Note: no `repairable.ts` existed on this branch — that file is on #123's branch, so this creates it here.)

**Minor 7 — path-source regression.** Covered by the new suite above.

**Nit — PR references in comments.** Both replaced with the invariant itself: a preview must not take the mutating state lock, because acquiring it creates `state.json` on a fresh home.

**Inherited, not duplicated:** #107's round three fixed centrally in `isSuppressed` (`src/self-update/notice.ts`) that `--dry-run` still wrote `version-check.json` on a TTY. That is not in this base and I have not added a copy; this branch picks it up at merge.

**Verification.** Targeted only, per the resource discipline on this stack. `tests/e2e/housekeeping-dry-run/` + `tests/unit/cache-format.test.ts` + `tests/commands/doctor-render.test.ts`: 38 pass, 0 fail. `tests/e2e/doctor.test.ts` + both `tests/misc/coverage*` suites: 94 pass, 0 fail. `bun run typecheck` and `bun run lint` both clean — lint now reports only the two pre-existing Biome config-migration infos that are also present on `main`. Every touched file is under the 200-line cap. No loops were changed, so there was nothing new to check for termination.

I did not run the full suite and am not quoting a suite-wide number, given the measured non-determinism on branches without #131.

PR #113, opened 2026-09-11
[LOG-5757] feat(refs): accept GitHub, GitLab, and Bitbucket browser URLs as references

stevekrenzel:
## What

`crew install https://github.com/anthropics/skills/tree/main/skills/internal-comms` now just works. Any URL copied from GitHub, GitLab, or Bitbucket's web UI is rewritten into the canonical `https://host/o/r.git@<ref>//<subpath>` form before the reference parser sees it, so every command that accepts a git reference (`install`, `info`, `tap add`, dependency lists) benefits.

Recognised shapes:

| Pasted | Becomes |
|---|---|
| `github.com/o/r/tree/<ref>/<path>` | `@<ref>//<path>` |
| `github.com/o/r/blob/<ref>/<path>/SKILL.md` | `@<ref>//<path>` |
| `github.com/o/r/commit/<sha>`, `/releases/tag/<tag>` | `@<ref>` |
| `gitlab.com/<group…>/r/-/tree|blob|commit|tags/…` (any host with `/-/`) | same rules |
| `bitbucket.org/o/r/src/<ref>/<path>`, `/commits/<sha>` | same rules |

Plus hygiene for every http(s) URL: `?query`, `#fragment`, trailing `/` dropped; `www.` removed from the host.

## Rules and judgment calls

- An explicit `@ref` or `//subpath` tail appended to a browser URL wins over the browser-derived value.
- A `blob` link to any file other than `SKILL.md` is `invalid_ref` (a reference must be a skill directory).
- The first segment after `tree/` is the ref; branch names containing `/` need the explicit `<url>@<branch>//<path>` form. Documented in the PRD and `crew help install`.
- `crew tap add` previously rejected any `@ref`. It now drops `main`/`master` (taps track the default branch) so a pasted `/tree/main/skills` link registers a subpath tap. Any other ref is still a `usage_error`. This applies to an explicit `@main` too; the one existing test that pinned `@main` as rejected now uses `@v1.0`.
- Unknown pages (`/issues/12`, `/releases/latest`) pass through unchanged and fail at clone time as before.

## PRD

§8.2 gains a "Browser URLs" subsection; §16.3 documents the `main`/`master` rule for `tap add`; criteria C-REF-23 through C-REF-27 added.

## Tests

- `tests/refs/browser-url.test.ts`: table tests through `parseRef` for every shape above, hygiene cases, explicit-tail precedence, nested GitLab groups, self-hosted GitLab with a port, passthrough of unknown pages, and the `blob` error.
- `tests/e2e/browser-urls.test.ts`: real CLI runs for the `blob` error (`invalid_ref`, exit 4) and a pinned browser link on `tap add` (`usage_error`, exit 4), plus `parseTapAddTarget` cases. No network: every case is decided before a clone.
- `tests/e2e/tap.test.ts`: new C-REF-27 test that `@main` is dropped on `tap add`.

`bun run check` is green: 1022 tests, 100% line and function coverage, Biome and tsc clean.

## Surfaces

`crew help install` gains a "Pasted from your browser" group and a note on the branch-name limitation; README reference table and the site's `SkillRefs.tsx` each get one example.


stevekrenzel:
## `/review`

Verdict: **not ready — 1 blocker, 4 majors, 2 minors, 1 nit**

### Blocker

**Rejected browser URLs disclose embedded credentials** — `src/refs/browser-url.ts:101`

The new blob-error path includes the original `raw` URL in both the human message and JSON `details.ref`. A URL containing userinfo or query credentials therefore publishes them to terminal output or captured JSON.

**Fix:** use a deliberately redacted display URL for both message and details, removing userinfo and sensitive query values and escaping control characters. Add a CLI JSON regression with dummy credentials.

### Majors

1. **Normalization strips credentials from valid private-repo URLs** — `src/refs/browser-url.ts:36`  
   Rebuilding `origin` from protocol, hostname, and port silently removes `username:password@`. Authenticated HTTPS clone URLs that previously worked become unauthenticated, contrary to the “any reachable git URL” contract and the documented hygiene list. Preserve userinfo for acquisition while using a separate redacted display form.

2. **Query/fragment text is parsed as explicit ref/subpath syntax** — `src/refs/git-url.ts:67`  
   Tail parsing runs before browser URL hygiene. A query or fragment containing `@v9` or `//evil` overrides the browser-derived ref/subpath instead of being dropped, violating C-REF-26. Strip/isolate query and fragment before grammar-tail parsing and add delimiter cases.

3. **The documented slash-branch workaround does not parse** — `PRD.md:660`, `src/commands/help/content/install.ts:132`  
   `https://github.com/acme/skills@feature/foo//python` leaves `@feature/foo` inside the clone URL because `splitGitRef` only recognizes a slash-free final segment. Implement slash-containing explicit refs or remove the advertised workaround and document the grammar’s actual restriction.

4. **Malformed HTTP URLs map to the wrong stable error** — `src/refs/browser-url.ts:30`  
   The catch passes malformed input unchanged to `canonicalizeUrl`, whose `new URL` throws `TypeError`; the CLI reports `usage_error` rather than required `invalid_ref` (C-REF-17). Translate malformed URLs to `CrewError("invalid_ref", ...)` and assert the JSON name/exit code.

### Minors

- `PRD.md:1735`: C-REF-25 says Bitbucket `/src/` follows C-REF-24’s non-`SKILL.md` file rejection, while the implementation deliberately treats any other leaf as a directory because Bitbucket’s shape is ambiguous. Specify Bitbucket separately and test the policy.
- `src/refs/browser-url.ts:55-68`: the forge maps are sparse, but `Record<string, Shape>` claims every arbitrary URL keyword resolves to a `Shape`. Use literal maps or `Record<string, Shape | undefined>` so undefined handling remains type-enforced.

### Nit

- A `blob/<ref>` URL with no file path says it “points at a file.” Give the empty-path case an accurate invalid-reference message.

### Reviewer coverage

- clarity: 3 findings
- naming: no findings
- type safety: 1 finding
- database: no findings
- testing: 4 findings
- security: 1 finding
- error handling: 2 findings
- observability: 1 overlapping finding
- performance: no findings
- concurrency: no findings
- architecture/conventions: 1 overlapping finding

Deduplicated overlaps are counted once. CI is green, but the blocker and majors prevent approval.

stevekrenzel:
Addressed all 8 findings in `618a40d`. I reproduced each one against the pre-fix parser before changing anything; what I observed is noted per item.

## Blocker — rejected browser URLs disclose embedded credentials

**Reproduced.** `crew install --json https://user:s3cr3t@github.com/acme/skills/blob/main/py/notes.txt` printed `s3cr3t` in both the human message and `error.details.ref`.

New `src/refs/display-url.ts` is the safe renderer: it masks URL userinfo, redacts the values of credential-bearing query parameters (`token`, `access_token`, `private_token`, `api_key`, …), and escapes control characters. Every reference echoed back from `browser-url.ts` now goes through it.

One thing the finding didn't mention, which my own test caught: when a URL is too malformed for `new URL`, `displayUrl` can't parse it either — and a malformed URL is exactly what reaches an error message. A textual userinfo scrub now covers that path too, so `https://user:s3cr3t@exa mple.com/a/b` is also safe.

**Reuse note for whoever merges this:** PR #108 landed `src/util/redact.ts` doing the same job for verbose progress lines. That branch isn't in this base, so I deliberately wrote the minimum this PR needs rather than duplicating its wider surface. **These two should collapse onto one shared renderer when the branches merge** — the file docstring says so as well.

## Major 1 — normalization strips credentials from valid private-repo URLs

**Reproduced.** `https://user:s3cr3t@github.com/acme/skills/tree/main/py` normalized to `https://github.com/acme/skills.git` — the credentials silently gone, so an authenticated clone that previously worked would now fail unauthenticated.

This is the inverse of the blocker, so I solved them together: one internal value that clones (userinfo intact) and one display value that is safe to print. `normalizeBrowserUrl` now rebuilds the origin with userinfo included, and nothing prints a URL except through `displayUrl`.

## Major 2 — query/fragment text parsed as explicit ref/subpath syntax

**Reproduced.** `.../tree/main/py?x=@v9` yielded ref `v9` instead of `main`; `.../tree/main/py#//evil` yielded subpath `evil` instead of `py`.

`stripUrlQueryAndFragment` now runs before `splitSubpath`/`splitGitRef`, so query and fragment text can never supply grammar. Covered for `?`, `#`, both together, and a query on a repo-only URL.

## Major 3 — the documented slash-branch workaround does not parse

**Reproduced.** `https://github.com/acme/skills@feature/foo//python` left `@feature/foo` inside the clone URL, producing `https://github.com/acme/skills@feature/foo` — unusable.

**Decision: implemented it rather than removing the claim.** Removing the workaround would leave a slash-containing branch (`feature/foo`, `release/2.0` — common conventions) with *no* way to be referenced at all, which is a functional gap rather than a documentation fix. The grammar supports it unambiguously: `splitSubpath` runs first, so by the time we look for a ref, any `//<subpath>` is already gone and a remaining `/` after the final `@` belongs to the ref.

The subtlety is that HTTPS userinfo and ssh-style `git@host:` also contain `@` before the host. A new `hostEnd` helper pins where host territory ends, so an `@` before it is never read as a ref delimiter. Verified `git@github.com:acme/skills.git`, `ssh://git@host/owner/repo.git@v2`, and `https://user:tok@github.com/acme/skills` all still parse as before.

PRD §8.4's `git-ref` production said "not containing `/`" — that is now corrected to describe what the grammar actually allows, with `:` and whitespace as the disqualifiers.

## Major 4 — malformed HTTP URLs map to the wrong stable error

**Reproduced.** `crew install --json https://:::/acme/skills` returned `usage_error` with the message *"crew hit an unexpected error: … please report this at …/issues"* — telling a user to file a bug for their own typo.

Malformed `http(s)` URLs now raise `CrewError("invalid_ref", …)` with a remedy hint. The e2e test asserts both the JSON error name and exit code 4.

## Minors

**Bitbucket vs C-REF-24** — C-REF-25 claimed Bitbucket `/src/` follows the same non-`SKILL.md` rejection as GitHub/GitLab `blob`, but the implementation deliberately differs because `src/` serves both files and directories with nothing in the URL to distinguish them. Rather than change behavior, I specified Bitbucket separately in §8.2 and C-REF-25: a trailing `SKILL.md` is stripped, any other leaf is taken as a directory, and a genuine non-skill file fails later with `no_skills_found`. Also renamed the internal shape from `file` to `dir` so the code states the policy it implements.

**Sparse forge maps** — now typed `Readonly<Record<string, Shape | undefined>>`, so `noUncheckedIndexedAccess` enforces the undefined branch instead of the type claiming every arbitrary keyword resolves.

## Nit

A `blob/<ref>` URL with no path said it "points at a file." It now reports that it names a branch but no file or folder.

## Verification

**Pre-fix failures.** I set my source changes aside, restored only the test files, and re-ran: **17 of 18** new parser tests and **all 3** new e2e tests fail against the pre-fix commit. The one that passes is the whitespace-disqualifies-a-ref case, which was already correct — it is a guard for the new slash-ref logic, not a regression test, and I'm flagging it rather than implying it caught something.

**File cap.** `git-url.ts` reached 205 lines, over the hard cap, so the two tail splitters moved to `src/refs/git-tails.ts` (75 lines; `git-url.ts` is now 144). **Merge interaction:** PR #125 extracts the same two helpers into a file of that exact name and teaches them to accept a trailing `@<ref>` after the subpath. Take #125's versions and re-apply the userinfo/`hostEnd` handling on top — the file docstring records this.

**Not fixed, pre-existing:** `tests/refs/parse.test.ts` is 286 lines, over the cap on `main` as well. Out of scope here.

`bun run check`: **1052 pass, 0 fail, 100% line and function coverage**, typecheck and Biome clean.

stevekrenzel:
## Updated `/review`

Verdict: **not approved — 2 blockers, 1 minor**

### Blockers

- **Valid authenticated URLs are exposed throughout tap output** (`src/commands/tap/add.ts:151`, `src/commands/tap/target.ts:46`, `src/commands/tap/index.ts:154`, `src/commands/tap/refresh.ts:46`). Tap-add success/collision output, JSON, list, and update output emit raw credentials. Use `displayUrl` for every outward-facing representation while retaining raw URLs only for config and Git operations.
- **Failure paths still expose raw authenticated URLs** (`src/refs/git-url.ts:77`, `src/commands/tap/target.ts:34`, `src/git/repo.ts:30`). Invalid-reference, pinned-tap, and clone failures leak raw URLs through messages/JSON details, and Git stderr can repeat the remote. Centralize redaction for user-visible URL fields and URL-bearing subprocess diagnostics, with human and JSON regression tests.

### Minor

- New comments in `src/refs/display-url.ts:13` and `src/refs/git-tails.ts:9` contain PR-specific merge instructions that will become stale. Replace them with durable local rationale or move integration notes out of source.

All eleven lenses reran against updated head `618a40d`; CI is green. I did not approve because blockers remain.

stevekrenzel:
Addressed all three findings in `3123e28`. Both blockers reproduced first; nothing declined.

## Blocker 1 — authenticated URLs exposed throughout tap output

Reproduced with a credential-bearing tap URL written into config. Four surfaces printed the secret verbatim:

| Surface | Before |
|---|---|
| `tap list` (human) | `acme  registered  https://user:s3cr3tT0k3n@…` |
| `tap list --json` | `"target": "https://user:s3cr3tT0k3n@…"` |
| `tap add` no-op | `pointed at https://user:s3cr3tT0k3n@…` |
| name collision | `…configured at \`https://user:s3cr3tT0k3n@…\` — … run \`crew tap add https://user:s3cr3tT0k3n@… <tap-name>\`` |

The collision message is the one I'd highlight: it hands back a **runnable command** containing the credential, so the remedy itself publishes the secret, twice.

Fixed at the chokepoints rather than the call sites — `displayTarget` and `payloadOf` in `src/commands/tap/target.ts` are what every human and JSON tap surface renders through, so redacting there covers `tap list`, `tap add` (all four outcomes), and the collision message at once. `src/commands/tap/refresh.ts` redacts when it builds each row, covering `tap update`'s human and JSON output together. The raw URL still reaches `config.yaml` and git untouched, so private-repo clones keep working — that was round one's fix and it stands.

## Blocker 2 — failure paths still exposed raw URLs

Reproduced two of the three sites. The blob-link path (`src/refs/git-url.ts`) was already redacted by round one; the other two were not:

- **Failed clone** (`src/git/repo.ts`) leaked through *both* channels — the interpolated URL and git's own stderr, which repeats the remote.
- **Pinned-tap error** (`src/commands/tap/target.ts`) echoed the user's raw argument in the message and in `details.raw`.

Also fixed `fetchAndCheckout` and `checkoutSha`, which quote stderr identically and weren't reported.

### Reconciliation with #108

As instructed, I did not write a third implementation. I read `src/util/redact.ts` on `feat/verbose-flag` and extended this branch's `display-url.ts` to **the same shape and division of responsibility**:

| #108 (`util/redact.ts`) | here (`refs/display-url.ts`) | role |
|---|---|---|
| `redactUrl` | `displayUrl` | a string that *is* a URL |
| `redactText` | `displayText` | URL-shaped substrings inside prose |
| `redactDetails` | `displayDetails` | mask values, preserve §13 keys |

They should collapse to one module at merge — `util/redact.ts` is the better home, since the concern isn't ref-specific. **#108 already covers `src/git/repo.ts`** (`cloneRepo` and `checkoutSha`), so expect a textual conflict there; either side's version is correct, and keeping #108's costs nothing here.

Two differences worth knowing when you reconcile:

- **SSH userinfo.** #108 deliberately keeps `git@host` (a username, not a secret). This branch's `displayUrl` masks any userinfo. #108's rule is better; adopt it.
- **This branch handles a case #108 doesn't.** See below.

### One gap my own first fix missed

My initial `displayText` used a URL-shaped regex, and the e2e test for a failed clone still failed. Git doesn't always echo the remote as we supplied it — for a `file://` URL it **strips the scheme** and reports the bare path:

```
fatal: '/repo.git?token=s3cr3tT0k3n' does not appear to be a git repository
```

No scheme to anchor on, so the URL match never fired. `displayText` now also redacts a bare `?query` tail. Worth porting into the merged renderer.

### One regression I introduced and fixed

Escaping controls across the whole string turned git's real newlines into `\x0a`, collapsing a readable multi-line diagnostic onto one line. Escaping is now per line: layout survives, ESC and CR are still neutralised.

## Minor — stale merge instructions in source

Both removed. `display-url.ts` now documents the three entry points and why each exists; `git-tails.ts` explains why `hostEnd` exists (HTTPS userinfo and scp-style SSH are both legitimate non-delimiter `@` uses) instead of pointing at #125.

## One thing beyond the findings

`src/git/repo.ts` was at 200 lines before this change and my additions pushed it to 203. Rather than split it arbitrarily, I used **#129's exact split** — `resolveRef`/`classifyRef` into `src/git/refs.ts`, same filename, same re-export from `repo.ts` for import stability. The two branches should now reconcile as a duplicate rather than a conflict. `repo.ts` is 142 lines, `refs.ts` 79.

## Verification

**Pre-fix:** all 5 e2e regressions in `tests/e2e/tap-credentials.test.ts` fail against `618a40d` (verified by stashing only `src/`); the unit file fails to load, since the two new helpers don't exist yet.

**Post-fix:** `display-url.ts` unit suite 15/15, `tap-credentials` 5/5, `tap-update` + `tap-unify` + browser-url suites 82/82, `git.test.ts` + `tap.test.ts` 49/50. `src/refs/display-url.ts` at 100% line and function coverage. Typecheck and Biome clean.

**The one failure is pre-existing.** `C-TAP-08 search --json` fails only in a multi-file run, at 11–17s — the signature of a live clone attempt. I confirmed it by running `git.test.ts` against the **completely unmodified base** (everything stashed) and reproducing the identical failure at identical timing, and `tap.test.ts` passes 42/42 in isolation. This is the network non-determinism #131 removes, not a regression here. Per-suite numbers are reported above rather than a full-suite total for that reason.

I re-read the two loops I touched (the argv-free `URL_IN_TEXT` replace and the per-line escape split) before running anything against them; both are bounded string operations with no recursion.

stevekrenzel:

**Verdict:** 2 blockers, 0 majors, 2 minors, 1 nit. **Not approved.**

Reviewed `72e6acf..7657f10` through all 11 `/review` lenses. CI, site checks, CodeQL, and the Vercel preview are green at this head.

### Blockers

1. **`tap update --dry-run` still prints configured tap credentials** — `src/commands/tap/refresh.ts:50-59`

   `planRefresh()` puts the raw `tap.url` into each pending row, and `tapUpdate()` forwards those rows directly to both the human renderer and `--json`. The real refresh path correctly calls `displayUrl()`, but dry-run bypasses it, so a configured `https://user:token@host/...` URL is exposed in terminal scrollback or CI output. This violates the new §8.2 guarantee that credentials never appear in any human message or JSON payload. Render pending URLs with `displayUrl(tap.url)` (or sanitize `TapRefreshRow` at one shared presentation boundary) and cover dry-run in both output modes.

2. **Common credential query parameters bypass the new redactor** — `src/refs/display-url.ts:32-43`, `src/commands/tap/add.ts:108-115`

   Query redaction is an exact allowlist and omits standard credential names such as `client_secret`. A reachable name-conflict path echoes the original `rawArg`; for example, a URL containing `?client_secret=SECRET` reaches both the human error and JSON error without masking `SECRET`. That breaks §13/C-REF-28's no-secret-output contract. Prefer redacting every query value in displayed URLs (queries are discarded from git references anyway), or implement a complete conservative credential-key policy shared by URL and bare-query redaction, with a `client_secret` regression test.

### Minors

1. **Primary docs overpromise slash-containing branch support** — `README.md:106`, `site/components/sections/SkillRefs.tsx:29-32`. Both say pasted browser links work as-is, while `/tree/feature/foo/...` is deliberately ambiguous and requires the explicit `@feature/foo//path` form. Add the caveat/workaround already documented in help and the PRD, or narrow the claim.
2. **C-REF-25 lacks the GitLab rejection half of its test** — `tests/refs/browser-url.test.ts:71-91`. The criterion requires a GitLab `blob` link to a non-`SKILL.md` file to return `invalid_ref`, but the table only covers the valid `SKILL.md` case. Add one negative GitLab case so a `file` → `dir` mapping regression cannot pass the conformance suite.
### Nit

1. **The PR description points reviewers to a removed test path.** Its third test bullet names `tests/e2e/tap.test.ts`; the C-REF-27 test now lives in `tests/e2e/tap/add.test.ts`.

No other actionable findings survived verification and deduplication. The two blockers are concrete credential disclosures, so they meet the release-holding bar rather than being polish.

Reviewer coverage: clarity, naming, type safety, database/state, testing, security, error handling, observability, performance, concurrency, and architecture/conventions.

stevekrenzel:
Round three addressed in `c5baaae`. Both majors and both minors reproduced first; one major is **refuted with evidence**, the rest are fixed.

## Major 1 — refuted: `@owner/repo@feature/foo` already parses

I could not reproduce this. `looksLikeAtShorthand` strips an optional `@ref` tail *before* it counts `owner/repo` segments (`git-url.ts:57-59`), so the slash inside the ref is never counted as a third segment. Every variant resolves, with and without a subpath:

| Reference | url | ref | subpath |
|---|---|---|---|
| `@acme/skills@feature/foo` | `https://github.com/acme/skills.git` | `feature/foo` | `""` |
| `@acme/skills@release/2.0` | `https://github.com/acme/skills.git` | `release/2.0` | `""` |
| `@acme/skills@feature/foo/bar` | `https://github.com/acme/skills.git` | `feature/foo/bar` | `""` |
| `@acme/skills@feature/foo//python` | `https://github.com/acme/skills.git` | `feature/foo` | `python` |

I changed no parsing code. **The coverage gap behind the finding is real, though**: the no-subpath slash-ref case was tested only for the `gh:` shorthand, not for `@owner/repo`, which is the form where nothing else marks where the repo body ends. That case is now covered by three table-driven assertions, and they pass against the unmodified source — stated plainly rather than counted as caught regressions.

## Major 2 — confirmed, both sites

A parseable URL that fails the repository-shape check leaked in both modes:

```
mode=human leaks=true
  `https://user:s3cr3tT0k3n@github.com/onlyowner` isn't a valid git reference
mode=json  leaks=true   details.ref carried the same string
```

`acquireTap`'s `no_skills_found` interpolated `tap.url` raw the same way.

Fixed at the **shared error boundary** — `writeError`/`writeMessageBlock` in `src/cli/output.ts` is the single sink every human message, remedy hint, and JSON payload passes through, so a future error cannot leak by forgetting. Both construction sites (`git-url.ts`, `acquire/index.ts`) also render safe values themselves, because a redacted URL reads better in a sentence than a scrub of the finished text. Re-running the probe after the fix gives `leaks=false` in both modes.

## Minor 1 — the docstring now describes something that happens

`displayDetails` had no production caller, so the claim of JSON payload protection was aspirational. Wiring it at the boundary made it true; the docstring now names `cli/output.ts` as the enforcement point and explains why construction sites still call `displayUrl` directly.

## Minor 2 — the regression that was missing

Added to `tests/refs/browser-url-safety.test.ts` and `tests/e2e/tap-credentials.test.ts`. You were right that the existing malformed and blob cases take branches round one had already redacted. Reverting only `src/` fails exactly the three new credential assertions (22 pass / 3 fail) while the slash-ref tests keep passing — which is what separates the confirmed major from the refuted one.

## #108 convergence

This branch's `src/refs/display-url.ts` and #108's `src/util/redact.ts` now solve the same problem at the same seam: #108's round three put `redactText` in `writeMessageBlock`, and this change puts `displayText`/`displayDetails` in `writeError`. **They will conflict in `src/cli/output.ts`, and that is the right place to resolve to one renderer** — keep `src/util/redact.ts` as the home, and carry across two rules from here: mask the *whole* userinfo rather than the password half (a token is usually the username), and redact a bare `?query` tail with no scheme, because git strips the scheme when it echoes a `file://` remote.

## Inherited, not duplicated

PR #107's round three suppressed the `version-check.json` write for `--dry-run` centrally in `isSuppressed`. Not in this base; no copy added here — this branch inherits it at merge.

## Verification

Targeted only, per the resource directive. `tap-credentials` + `browser-url-safety` + `browser-urls` + `error-paths`: **50 pass, 0 fail**. `display-url.ts` and `acquire/index.ts` at 100% line and function coverage; the two remaining lines in `output.ts`/`git-url.ts` are subset artifacts — I measured the same gaps on the unmodified branch before concluding that. Typecheck and Biome clean.

One process note: the formatter flagged my new test and I caught it with `bun run lint` before pushing, which is the drift that turned #112's CI red.

stevekrenzel:
Rebased onto `main` after #111 merged. Conflicts reconciled in `src/refs/git-url.ts`: both imports kept (`normalizeBrowserUrl`/`stripUrlQueryAndFragment` and #111's `normalizeSubpath`), and the parsed source now uses `ref: gitRef ?? browser.ref` with the browser-derived subpath fed through `normalizeSubpath` so §8.4 containment applies to browser URLs too. PRD C-REF rows from both sides kept. Refs and browser-URL tests pass (118).

stevekrenzel:
Rebased onto `main` after the cap-compliance refactors (#135–#143) merged. What was reconciled:

- **`src/git/repo.ts` → `repo/{index,refs}.ts` (#143).** This PR had split the same file itself (`src/git/refs.ts` + a re-export); `main`'s split is the same two functions, so this PR's copy is dropped and its credential-redaction edits land in the split files instead: `displayUrl`/`displayText` in `cloneRepo` and `fetchAndCheckout` (`repo/index.ts`), `displayText` in `checkoutSha` (`repo/refs.ts`). Import-only conflicts in `tap/refresh.ts` and `sources/acquire/index.ts` resolved to the split paths.
- **`tests/e2e/tap.test.ts` → `tests/e2e/tap/` (#140).** The `@ref`-tail edit and the new `C-REF-27` test moved to `tests/e2e/tap/add.test.ts` (151 → 163 lines).
- **Cap gate (#137).** `src/commands/tap/add.ts` was 199 on `main`; this PR's redaction lines took it to 202, so the three "make a new tap" helpers (`deriveName`, `cloneNewTap`, `newTapOf`) moved to `src/commands/tap/new-tap.ts` (40 lines; `add.ts` now 172). Folded into the commit that grew the file.

Verification: `bun run check` — typecheck, lint (incl. the cap gate), 1178 tests, 100% line/function coverage; site lint clean; targeted refs/tap/browser-URL suites 251 pass.

PR #114, opened 2026-09-11
[LOG-5758] feat(cli): accept --help/-h and --version/-v like a standard CLI

stevekrenzel:
## What

`crew --help`, `crew -h`, `crew install --help`, `crew install -h`, `crew --version`, `crew -v`, and `crew -V` all exited 4 with `usage_error`. Every other CLI accepts these spellings, so now crew does too.

Implemented as a pure argv rewrite in `src/cli/args.ts` before yargs runs:

- `--help` / `-h` anywhere on the line → `help <command>`, where `<command>` is the first non-flag token (a leading `help` is skipped). Subcommand words are ignored, so `crew tap remove --help` shows the `tap` page. Only `--json` survives the rewrite, so `crew install --help --json` still emits structured help.
- `--version` / `-v` / `-V` as the first token → `version`, honoring `--json`. After a command name (`crew install -v`) it stays an unknown-flag `usage_error`, leaving `-v` free for a future `--verbose` short form.

## PRD

- §5.5 "Invocation surface" gains bullets for both forms.
- §18.3: C-CLI-15, C-CLI-16, C-CLI-17 (inserted after C-CLI-14).

## Surfaces

- `crew help` / `crew version` help pages and the overview footer mention the flag spellings.
- README command table and `site/components/sections/Commands.data.tsx` rows for help/version mention them.

## Testing

New `tests/e2e/help-flags.test.ts` (8 tests) asserts byte-for-byte equivalence with `crew help`, `crew help <command>`, and `crew version` for every form, structured `--json` output, the unknown-command fallback, and the `crew install -v` negative case. `bun run check` green: 993 pass, 100% line and function coverage.


stevekrenzel:
## `/review`

Verdict: **not ready — 0 blockers, 1 major, 4 minors, 2 nits**

### Major

**The rewrite drops valid JSON boolean spellings** — `src/cli/args.ts:67`

`rewriteConventionalFlags` preserves JSON mode only when the argv array contains the exact token `--json`. The existing yargs parser also accepts `--json=true`, so `crew help --json=true` emits structured JSON while `crew --help --json=true` rewrites to plain `help` and emits human text; version behaves the same. Duplicate/final boolean value semantics can also change.

**Fix:** preserve the effective parsed JSON value, or at minimum recognize and preserve supported `--json=<boolean>` forms and final-value semantics. Add equivalence tests for help and version.

### Minors

- `src/cli/args.ts:71`: skip only the leading `help` token. `crew help help --help` should match `crew help help`, but filtering every literal `help` rewrites it to the overview.
- `PRD.md:198`: define precedence when both conventional flags appear. `--help` applies “anywhere,” while first-token `--version` also applies; current behavior is order-dependent. Add a conformance case.
- `tests/e2e/help-flags.test.ts:55`: compare structured help byte-for-byte with canonical `help … --json`, or validate the complete schema; checking only `name` and `flags` permits truncated JSON.
- `tests/e2e/help-flags.test.ts:65`: this new flag-rewrite test is labeled C-CLI-10; label it C-CLI-15 so conformance failures trace to the right criterion.

### Nits

- `src/commands/help/content/version.ts:6` mentions `--version` and `-v` but omits supported `-V`.
- README and the site likewise mention only a subset of the documented equivalent spellings; list all three consistently.

### Reviewer coverage

- clarity: 1 finding
- naming: no findings
- type safety: no findings
- database: no findings
- testing: 4 findings
- security: no findings
- observability: no findings
- performance: no findings
- concurrency: no findings

Deduplicated overlaps are counted once. CI is green, but the major prevents approval.

stevekrenzel:
Addressed all 7 findings in `efe67a4`. I reproduced each one against the pre-fix branch before changing anything; notes below say what I actually observed.

## Major — rewrite drops valid JSON boolean spellings

Reproduced, and the final-value case was worse than reported:

| argv | before | canonical equivalent |
|---|---|---|
| `crew --help --json=true` | human text | `crew help --json=true` → JSON |
| `crew --version --json=true` | `crew 0.10.1` | `crew version --json=true` → `{version}` |
| `crew --help --json --json=false` | **JSON** | `crew help --json --json=false` → human text |

That last row is inverted rather than merely missing: the rewrite saw the bare `--json` token and forced JSON on, where the canonical form correctly honored the trailing `--json=false`.

Rather than widen the pattern match, `rewriteConventionalFlags` now asks the real parser via a new `effectiveJson()` (`src/cli/args.ts`). I probed yargs first to confirm the semantics worth preserving — last-occurrence-wins, and `--json=TRUE` is *false* since only lowercase `true` is accepted. Hand-rolling a second implementation would have gotten that wrong eventually. Extracting `baseParser()` for both call sites also removes the duplicated parser configuration, so the rewrite and the real parse cannot drift.

## Minors

- **Leading `help` only** (`src/cli/args.ts`). Confirmed: `crew help help --help` printed the overview instead of the `help` command's page, because the predicate filtered *every* literal `help`. Now only `argv[0]` is skipped.
- **Precedence** (`PRD.md` §5.5). Confirmed order-dependent: `--version --help` gave version, `--help --version` gave help. Chose **`--help` wins in either order** — it is the broader request, and a user who types both is asking what the tool can do. §5.5 states it; C-CLI-17b covers it.
- **Byte-for-byte structured help** (`tests/e2e/help-flags.test.ts`). Now compares full output against `crew help <cmd> --json` and `crew help --json`. To be precise about what this buys: this assertion **passes both pre- and post-fix**, so it is a guard against truncated JSON, not a regression catcher. I would rather say that than let the test count imply it caught something.
- **Mislabeled criterion**. `C-CLI-10` → `C-CLI-15` on the unknown-command fallback test.

## Nits

- `src/commands/help/content/version.ts` now lists `-V` alongside `--version` and `-v`.
- `README.md` and `site/components/sections/Commands.data.tsx` list all spellings consistently (README's version row gained `-V`; the site's help row gained `-h` and its version row gained `-v`/`-V`).

## Pre-fix verification

Stashed `src/` only, keeping the new tests, and ran the file: **8 pass, 4 fail**. The four failures are C-CLI-17a (both), C-CLI-17b, and C-CLI-17c — every new behavioral test fails against the pre-fix commit. Restored afterward and re-confirmed green.

## Merge note

This touches `rewriteConventionalFlags` and the yargs setup in `src/cli/args.ts`. **PR #111** adds a duplicate-scalar-flag guard in the same file, keyed off the repeatable-flag list. The two changes are in adjacent but distinct regions — #111 guards the main `parseArgs` path, this guards the pre-parse rewrite — so I expect a textual conflict around the parser configuration rather than a semantic one. Whoever merges second should keep `baseParser()` and re-apply the duplicate guard on top of it.

## Gate

`bun run check`: 997 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean. (The two Biome "migrate configuration" infos are pre-existing on `main`.)

stevekrenzel:
## Updated `/review`

Verdict: **not approved — 0 blockers, 1 major**

### Major

- **Help-target detection mistakes flag values for command names** (`src/cli/args.ts:106`). For example, `crew --scope project install --help` rewrites to `crew help project` and falls back to the overview instead of install help. Yargs-supported spaced boolean values have the same issue, such as `crew --help --json false install`. Derive the target from parser positionals after configuring flags, or make the scan consume known scalar/array/boolean values. Add `--scope`, `--agent`, and spaced `--json true|false` equivalence tests for both `--help` and `-h`.

All eleven lenses reran against updated head `efe67a4`; CI is green. I did not approve because the major remains.

stevekrenzel:
Addressed the major in `e906ed9`.

## Major — help-target detection mistakes flag values for command names

Reproduced first. Every one of these rewrote to a non-command and fell back to the overview:

| argv | rewrote to | showed |
|---|---|---|
| `crew --scope project install --help` | `help project` | overview |
| `crew --scope project install -h` | `help project` | overview |
| `crew --agent codex install --help` | `help codex` | overview |
| `crew --help --json false install` | `help false` | overview |

The cause is that `rest.find((a) => !a.startsWith("-"))` cannot distinguish a command from a flag's *value* — `project` looks exactly like `install` to a string scan.

**Approach: parser positionals, not a value-aware scan.** I probed both before choosing. A scan needs a hardcoded list of which flags consume a following token, and that list rots every time a flag is added — `--interval`, `--from-git`, and `--tap` would each need remembering. The parser already knows, so `positionalsOf()` asks it.

It runs two passes, and the second one earns its keep:

```
["uninstall","--prune","foo","--help"]  one-pass: ["uninstall"]  two-pass: ["uninstall","foo"]
```

With only the global tables, `--prune` is unknown and swallows `foo` as its value. The first pass learns the command, the second re-parses with that command's own tables. Only the first positional is used, but the accuracy is free.

I also confirmed the parser handles subcommand-scoped *string* flags it hasn't been told about — `["autoupdate","enable","--interval","4h","--help"]` yields `["autoupdate","enable"]`, because yargs infers `4h` belongs to `--interval`.

## Tests

New `tests/e2e/help-flag-targets.test.ts` (C-CLI-17d), asserting byte-identical output against the canonical `crew help <command>` for scalar flags (`--scope`, `--from-git`), the repeatable `--agent`, spaced `--json true|false`, and subcommand-scoped flags (`--interval`, `--prune`) — each for both `--help` and `-h`.

**Pre-fix: 3 of 4 fail** against `efe67a4` (stashing only `src/`). The fourth — repeatable-flag values — passes either way, because `--agent` was already declared with `nargs`, so its values never reached the scan. I'd rather say that than let "4 new tests" imply four caught regressions.

Round one's cases still hold: `--json=true`, trailing `--json=false`, `help help --help`, and `--version --help` all remain byte-identical to their canonical forms.

## File split

The change pushed `src/cli/args.ts` to 235 lines, over the hard cap, so it is now `src/cli/args/`:

- `tables.ts` (71) — flag tables and the shared parser configuration
- `conventional-flags.ts` (73) — the §5.5 rewrite
- `index.ts` (118) — the real parse

One concept per file, and the tables now have a single home that both the rewrite and the real parse read, so they cannot drift.

## No PRD change

§5.5 already specifies `<command>` as the help target. This makes the code match what the spec always said, which per CLAUDE.md is a bug fix rather than a spec change.

## Sibling PRs touching `src/cli/args.ts`

Worth flagging, since the file moved:

- **#111** counts flag occurrences from **raw argv**. No conflict with this change — I read positionals from the *parsed* result, and raw-argv counting is unaffected by where the parser is configured. Its guard belongs in `args/index.ts` alongside `parseArgs`.
- **#114 → #116/#124** (`FLAG_TABLE_ALIASES`, `aliasFlagKey`) and **#117/#130**: those add to the flag tables, which now live in `args/tables.ts`. Expect the move rather than the content to conflict.

Whoever merges second: keep the directory, re-apply the hunks into `tables.ts` or `index.ts` as appropriate.

## Verification

Targeted suites only, given the known non-determinism without #131: `help-flag-targets` 4 pass, `help-flags` + `help` + `aliases` 29 pass, 0 fail. All three new files at 100% function and line coverage. Typecheck and Biome clean (the 2 infos are the pre-existing config-migration notices on `main`).

`positionalsOf` is two fixed parses and an index — no loop or recursion — so there is nothing whose termination depends on state; I checked that before running anything against it.

stevekrenzel:
## Updated `/review` (latest head)

**Verdict:** 0 blockers, 0 majors, 2 minors, 1 nit. **Ready from a code-review standpoint.**

Reviewed frozen diff `72e6acf..7dbabb1` through all 11 `/review` lenses. The prior command-target and malformed-argument failures are fixed. Existing CI is green at this head (`check`, site CI, and CodeQL).

### Minors

1. **The PR description still documents the pre-fix implementation** — PR description
   It says the target is the “first non-flag token” in `src/cli/args.ts`, lists only C-CLI-15–17, and reports the original test/check counts. The implementation now deliberately uses the first parser positional in `src/cli/args/conventional-flags.ts`, adds C-CLI-17a–f, and has a larger regression suite. Refresh the description so future reviewers do not infer the exact bug the later commits fixed.

2. **Command-keyed flag registries do not validate their keys** — `src/cli/args/tables.ts:19,27`
   `BOOLEAN_SUB` and `STRING_SUB` use `Record<string, …>`, while the new `EVERY_COMMAND` path flattens every entry. A future misspelled command key would silently affect help discovery. Consider `satisfies Partial<Record<CanonicalCommand, readonly string[]>>` so TypeScript checks keys while preserving sparse tables.

### Nit

- `src/cli/args/conventional-flags.ts:62-64` still says repeated `--json` uses “last occurrence wins,” while §5.2 and the implementation now reject repeats. Update that comment to describe only the accepted single-occurrence spellings.

### Reviewer coverage

- clarity: 1 finding
- naming: no findings
- type safety: 1 finding
- database/state: no findings
- testing: no findings
- security: no findings
- error handling: no findings
- observability: no findings
- performance: no findings
- concurrency: no findings

GitHub approval was attempted, but GitHub rejected it because the authenticated account is the PR author; GitHub does not allow self-approval.

stevekrenzel:
Round three addressed in `9acae1c`. All six reproduced first; nothing declined.

**Major 1 — a command-scoped flag before the command breaks help targeting.** Reproduced: `crew --prune uninstall --help` rewrote to `["help"]` (overview) instead of `["help","uninstall"]`, and the same for `-h`, `--recursive install`, and `--recursive tap`. The discovery parse knew only the global tables, so the unknown `--prune` consumed `uninstall` as its value.

Round two moved target discovery to parser positionals to stop a flag's *value* being read as the command. This is the same class one level down: the flag's arity is command-scoped, but the command isn't known yet — a chicken-and-egg the two-pass design didn't cover when the flag comes first.

The first discovery pass is now widened to every command's tables (`EVERY_COMMAND` in `src/cli/args/tables.ts`). That is sound precisely because this pass only answers *which positional is the command*; the second pass still narrows to that one command's tables so arity resolves exactly. **The real parse in `src/cli/args/index.ts` is untouched** — it must keep rejecting a flag belonging to a different command, and widening it would have been the tempting wrong fix. I added a case for `--recursive`, which belongs to both `install` and `tap`, to pin that widening doesn't blur which command was named.

**Major 2 — malformed value flags escape as raw yargs errors.** Reproduced end to end: `crew --help --agent` returned `exitCode = undefined` and printed `Error (undefined)` — a crash with no error name, exactly as reported. The discovery parses run *before* the real parse installs its `.fail()` handler, so nothing translated the `YError`. They now route through a `discover()` wrapper that maps parser failures to `usage_error`; the same input gives exit 4, the stable name, and the usual remedy hint.

One note on the residual stderr line yargs prints internally: that is pre-existing and not introduced here — `crew install --agent` (no rewrite involved) produces the identical line via the real parse. The captured, user-facing output now matches between the two paths.

**Minors.**
1. §5.5 said "first non-flag token", which wrongly includes flag values. It now specifies the first parser *positional*, and states that resolving it accounts for command-scoped arity even when such a flag precedes its command — i.e. the spec now describes what round two and this change actually implement.
2. `C-CLI-17d` genuinely did not exist in §18.3. Added it, plus `C-CLI-17e` (leading command-scoped flag) and `C-CLI-17f` (malformed value flag → `usage_error`/4) for the new behaviour.
3. `BOOLEAN_SUB`/`STRING_SUB` are now `Readonly<Record<...>>` and `BUILT_IN_FLAGS` a `ReadonlySet`.
4. `CLAUDE.md` pointed at the removed `src/cli/args.ts` in two places. Both now name `src/cli/args/tables.ts`, and I added a row for adding a *command-scoped* flag, which the table never covered.

**Pre-fix verification.** Reverting only the two source files fails exactly the two new tests (C-CLI-17e, C-CLI-17f) while the four pre-existing ones pass — so the new tests are load-bearing rather than merely additional.

**Verification.** `help-flag-targets` 6 pass; with `help-flags` + `help` + `aliases` + `error-paths`, 39 pass, 0 fail. `conventional-flags.ts` and `tables.ts` both at 100% line and function coverage. Typecheck and Biome clean (the 2 infos are the pre-existing config notices, also on `main`). Targeted runs only, per the known non-determinism on branches without #131.

**Inherited at merge:** #107's round three fixed centrally, in `isSuppressed` (`src/self-update/notice.ts`), that `--dry-run` wrote `version-check.json` on a TTY. That is not in this base, so this branch picks it up on merge rather than needing its own copy.

stevekrenzel:
Rebased onto `main` (`15210a0`) after #111 (`--from-git`, repeated-flag rejection) and #116 (command aliases) merged. Both of those changed `src/cli/args.ts`, which this PR splits into `src/cli/args/{index,tables,conventional-flags}.ts`, so the rebase had to carry their changes into the split. Head is now `d206c57`; every commit is signed.

**Reconciled into the split (behavior identical to `main` plus this PR's feature):**

1. `--from-git` moved out of `STRING_GLOBALS` into `STRING_SUB.install` (`tables.ts`), so it stays install-only and reaches `ctx.flags.extras` as on `main`.
2. Alias-aware flag tables: a new `subFlags(table, command)` in `tables.ts` resolves through `aliasFlagKey` + `lookup`, and is used by both the real parse (`index.ts`) and the `--help` discovery parse (`withFlagTables`), so `crew rm --prune --help` targets the same page `crew uninstall --prune --help` does. The `EVERY_COMMAND` widening from this PR is unchanged.
3. `rejectRepeatedScalars` / `repeatedFlagError` ported verbatim into `index.ts`, called right after positional parsing as on `main`.
4. `main.ts` keeps both imports (`canonicalCommand` from `./aliases.ts`, `parseArgs` from `./args/index.ts`); `tests/e2e/flag-guards.test.ts` (from #111) now imports `args/index.ts`.

**One real semantic collision, fixed in a separate commit (`d206c57`):** #111's §5.2 rule makes `crew help --json --json=false` a `usage_error`, but this PR's rewrite collapsed `--help --json --json=false` to one effective value and printed the page — a rewritten form laundering what the canonical form rejects. The rewrite now forwards every spelled `--json` occurrence verbatim when repeated, so the real parse raises the identical error; a single occurrence still carries its parser-derived value. The last-occurrence-wins promise in §5.5 and C-CLI-17a contradicted §5.2/C-CLI-08a, so it's dropped and C-CLI-17a now asserts the shared `usage_error` for `--help` and `--version` forms.

Verified locally: `bun run lint`, `bun run typecheck`, and the targeted suites (`alias-flags`, `aliases`, `flag-guards`, `help-flag-targets`, `help-flags`, `help`, `install-from-git`, `install-known-tap-flags`): 67/67 pass. No conflict markers; every file under the 200-line cap.

stevekrenzel:
Rebased onto `main` after the cap-compliance refactors (#135–#143) merged.

- `main` adopted this PR's own `src/cli/args/{index,tables}.ts` layout (#135), so the diff here shrank to the feature itself: `conventional-flags.ts` (new), `withFlagTables` / `EVERY_COMMAND` and the `command: string | undefined` widening in `tables.ts`, and the `rewriteConventionalFlags` hook in `index.ts` (+7 lines).
- The homepage rows for `help` and `version` moved from the deleted `Commands.data.tsx` into `Commands.data/meta.tsx` (#136); the first commit now carries the final wording rather than the intermediate one.
- `CLAUDE.md`: this PR's fuller `yargs` / flag-table paragraphs kept, on top of `main`'s new cap-gate paragraph; `help/content/alias.ts` keeps `main`'s docstring (`tables.ts` is where the alias resolves).

Every source and test file this PR owns is byte-identical to the previous head. `bun run check`: typecheck, lint (now including the 200-line gate), 1110 tests, 100% line and function coverage; site lint clean. All commits signed.

PR #115, opened 2026-09-11
[LOG-5759] docs(prd): reconcile exit-code 2 with §5.4/§9 and fold in §18.6 resolutions

stevekrenzel:
## What

PRD-only change. Makes the exit-code spec self-consistent with itself and with the code.

- **§15** — exit 2 was described as "nothing was attempted (e.g. install with only already-installed skills)". §5.4 and §9 step 9 say that case exits 0, and `src/commands/install/index.ts` returns 0. A grep of `src/` and `tests/` finds no path that emits 2. The row now reads "Reserved. Not emitted by any command today" and points at where the old examples actually land (0 for already-installed, 4 `no_skills_found` for an empty expansion).
- **§9** — removed the trailing "Exit code: 0 … 1 … 2 …" sentence after step 11, which duplicated and contradicted step 9's table. Step 9's table now states the mixed-outcome rule explicitly (any root with zero successful agents → exit 1).
- **§9 step 6 and §12.1** — folded in the two remaining §18.6 resolutions: topological install order (deps before dependents, cycles broken in any valid order) and the hash of an empty directory.
- **§18.6** — retitled "Ambiguities resolved during drafting" and rewritten as a short changelog pointing at the sections that now carry each answer, instead of a note saying they "should be folded into the next revision".

## Why

Two conformant implementations reading §15 vs §5.4/§9 would disagree on whether an already-installed skill exits 0 or 2. The code has always done 0.

## Sections touched

§9 step 6, §9 step 9, §9 trailer, §12.1, §15, §18.6.

## Verification

- No `src/` or `tests/` changes. README's only exit-code mention is a generic pointer to §15; the site has none.
- `bun run check` green: 985 pass, 100% coverage, tsc and Biome clean.

## Judgment call

Kept §18.6 rather than deleting it, so readers of older drafts can find where each answer moved.


stevekrenzel:
## `/review`

Verdict: **not ready — 0 blockers, 1 major, 2 minors, 1 nit**

### Major

**The new topological-order guarantee is false for resolved names that differ from reference tails** — `PRD.md:748` (implementation at `src/install/topo.ts:21`)

`topoSort` rebuilds graph edges by applying `extractDepName` to each raw dependency reference. For a path dependency such as `./dep-directory` whose `SKILL.md` declares `name: dep`—explicitly allowed by the reference/source model—the extracted name is `dep-directory`, no `byName` entry matches, and the dependency edge disappears. The already-enqueued root can then be installed before `dep`, contradicting the new normative promise.

**Fix:** build the ordering graph from resolved dependency identities already recorded during resolution (for example `requiredBy`), rather than guessing from raw refs. Add a dependency whose directory/ref tail differs from its declared name and assert install order before making the guarantee normative.

### Minors

- Add an observable C-DEP-01 ordering test. Existing tests assert only that both skills exist afterward, so reversing their install sequence still passes.
- Add an e2e mixed-outcome test with two direct root arguments, one succeeding and one failing on every selected agent, and assert exit 1 plus both summary rows. The existing expansion case does not pin the newly explicit direct-root rule.

### Nit

- `PRD.md:761`: “any root with zero successful agents makes the run exit 1” overstates the partial-success row because when all roots fail and one has a validation failure, exit is 4. Qualify it as applying to a mixed-outcome run.

### Reviewer coverage

- clarity: no findings
- naming: no findings
- type safety: no findings
- database: no findings
- testing: 2 findings
- security: no findings
- error handling: no findings
- observability: no findings
- performance: no findings
- concurrency: no findings
- architecture/conventions: 1 finding

CI is green, but the major spec/implementation mismatch prevents approval.

stevekrenzel:
Addressed all four findings in `e801042`. This PR now carries a code change — details below.

## Major — topological guarantee was false for resolved names differing from reference tails

Reproduced before touching anything. Fixture: directory `dep-directory` containing a `SKILL.md` that declares `name: dep`, and a `root` skill depending on it by path.

```
ORDER: root -> dep
requiredBy: [["dep",["root"]]]
```

The dependent was ordered **before** its dependency, while `requiredBy` had the correct edge all along. `topoSort` (`src/install/topo.ts`) rebuilt edges by running `extractDepName` over each raw dependency reference: the tail `dep-directory` is absent from `byName`, so `.filter(… byName.has(n))` discarded the edge silently. §9 step 4 explicitly permits a declared name to differ from its directory, so this was reachable through ordinary use, and the same applies to git refs and subpaths whose tail differs from the declared name.

**Fix:** `topoSort` now takes the resolver's `requiredBy` map — which records *resolved* names on both ends of every edge, built during the dependency walk in `src/install/resolve/index.ts` — and inverts it into dependent → deps. No name is ever re-derived from a reference string. `extractDepName` had no other caller and is deleted.

**Why a code change rather than narrowing the spec:** dependency-before-dependent is load-bearing for marker, state, and adapter writes seeing a consistent view, and the resolver already had the correct data. Weakening the spec to match a filtering bug would have documented the defect instead of fixing it.

Post-fix the same fixture yields `dep -> root`.

## Minor — observable ordering test

The existing C-DEP-01 tests asserted only that both skills exist afterward, so reversing the sequence still passed. New `tests/e2e/install-order.test.ts` asserts on the order the resolver actually produces:

- `C-DEP-01` plain dependency ordering.
- `C-DEP-01` ordering when the dep's directory differs from its declared name (the regression above).
- `C-DEP-08` a cycle still yields every skill exactly once.

## Minor — mixed-outcome e2e test

Added `C-INST-21a`: two directly-named roots where `good` succeeds and `blocked` fails on the only selected agent (an untracked directory at its destination). Asserts exit 1, both roots present in the output, `good` installed, and the untracked file untouched. New criterion `C-INST-21a` added beside `C-INST-21`, which states the exit-code table but did not pin the direct-root case.

## Nit — §9 step 9 partial-success row

Qualified: the "any root with zero successful agents makes the run exit 1" rule now explicitly applies to runs where at least one root succeeded, and notes that when no root succeeded a validation failure exits 4 rather than 1.

## Pre-fix verification

Stashed `src/` only, keeping the new tests: **1 of 4 fails** — `C-DEP-01 ordering holds when the dep's directory differs from its declared name`. The other three pass both ways; they are guards against future reordering, not proof of a present bug, and I'd rather say so than let a count imply otherwise.

## Spec updates

- §9 step 6 — ordering uses resolved names, with the differing-tail case spelled out.
- §9 step 9 — partial-success row qualified.
- C-DEP-01 — extended to cover the differing-tail case.
- C-INST-21a — new.

`bun run check`: 989 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean. `src/install/topo.ts` is 61 lines and the new test file 124, both under the cap.

stevekrenzel:
## Updated `/review`

Verdict: **0 blockers, 0 majors, 2 minors**

### Minors

- `PRD.md:1406` says exit 0 means every requested skill succeeds, while §9 step 9 and the implementation calculate outcomes across every attempted skill, including dependencies. Use “every attempted skill” or explicitly include dependencies.
- The cycle test at `tests/e2e/install-order.test.ts:69` proves each cycle member appears once, but not that acyclic edges around a cycle remain ordered. Add a node `c` depending on cycle member `a`, resolve from `c`, and assert both cycle members precede `c`.

All eleven lenses reran against updated head `e801042`; CI is green. There are no blockers or majors, so this PR qualifies for approval.

stevekrenzel:
Both minors addressed in `8ce509f`. Nothing declined.

**Minor 1 — exit-0 wording** (`PRD.md:1406`). Confirmed the mismatch: the §15 table said "every **requested** skill", while §9 step 9 and `src/commands/install/index.ts` compute the outcome over every **attempted** skill, which includes dependencies pulled in with a root. So a run where a root succeeded but its dependency failed everywhere would exit 1 while the table implied 0. Reworded to name both explicitly: "every root named on the command line plus every dependency pulled in with it."

**Minor 2 — cycle ordering** (`tests/e2e/install-order.test.ts`). You were right that the existing test only proved membership. I added `C-DEP-08 acyclic edges around a cycle stay ordered`: `a` and `b` form a cycle, `c` depends on `a` from outside it, resolve from `c`, and assert both cycle members precede `c` rather than merely appearing in the set.

I checked the new assertion is load-bearing rather than trivially true. Dropping dependency edges entirely in `topoSort` fails it:

```
(fail) C-DEP-01 dependency is ordered before its dependent
(fail) C-DEP-01 ordering holds when the dep's directory differs from its declared name
(fail) C-DEP-08 acyclic edges around a cycle stay ordered
 2 pass, 3 fail
```

The original cycle test passes under that same breakage, which is exactly the gap you identified.

Targeted: `tests/e2e/install-order.test.ts` 5 pass, 0 fail. Typecheck and Biome clean. Per-suite numbers rather than a full-suite total, given the known non-determinism on branches without #131.

PR #116, opened 2026-09-11
[LOG-5760] feat(cli): add remove/rm, ls, and upgrade aliases

stevekrenzel:
## What

Adds the top-level aliases people reach for from Homebrew muscle memory:

| Alias | Canonical |
|---|---|
| `crew remove …`, `crew rm …` | `crew uninstall …` |
| `crew ls` | `crew list` |
| `crew upgrade [...]` | `crew update [...]` |

Each was a `usage_error` before.

## Why it needed more than a table entry

`src/cli/args.ts` looks up per-subcommand flag tables by the raw command word, so `crew rm --prune foo` would have rejected `--prune` as unknown. The alias table now lives in `src/cli/aliases.ts` and the parser resolves an alias to its canonical command before consulting the flag tables. The dispatcher imports the same table.

## PRD

- §5.1: four alias lines next to the existing `skills` / `taps` / `untap` aliases.
- §18.3: new criterion **C-CLI-01b** (aliases behave identically to their canonical command, including accepting its flags).

## Surfaces

- Help: one entry per alias (`ls`, `remove`, `rm`, `upgrade`), registered in the "Managing skills" group with "Alias for …" one-liners, following the existing alias-page pattern.
- README command table and `site/components/sections/Commands.data.tsx` gain matching rows.

## Tests

`tests/e2e/aliases.test.ts`: `ls --json` equals `list --json`; `upgrade --json` equals `update --json`; `remove`/`rm` produce uninstall's `not_installed_here`; `rm --prune --agent claude-code <name>` accepts uninstall's own flags and removes the skill; `ls --prune` is still rejected. `tests/e2e/help.test.ts` now includes the four aliases in the every-command help checks.

`bun run check` green: 990 pass, 100% line and function coverage. `site` lint clean.


stevekrenzel:
## Review

Verdict: **not approved — 0 blockers, 1 major, 3 minors, 1 nit**

### Major

- **Alias help omits supported canonical flags** (`src/commands/help/content/remove.ts:14`, `rm.ts:14`, `ls.ts:14`). The pages say the aliases take the canonical command's flags, but `remove` and `rm` omit both `--scope` and `--force`, while `ls` omits `--scope`. Those flags are parsed and used by the canonical handlers, and PRD §5.5 requires useful per-command help. Mirror the canonical flag lists and cover them in the help tests.

### Minors

- **Prefixed aliases now accept a meaningless `--recursive` flag** (`src/cli/args.ts:61`, `src/cli/aliases.ts:21`). Canonicalizing every alias for flag lookup also changes the existing `taps` and `untap` aliases into the `tap` flag table, so `crew taps --recursive` and `crew untap --recursive` are accepted even though `tap list`/`tap remove` ignore that flag. Preserve strict rejection for prefixed aliases, or select flags from the effective command plus subcommand.
- **Alias targets are not constrained to real commands** (`src/cli/aliases.ts:13`). `canonical` is a plain `string`, so a typo such as `uninstal` compiles and fails only at runtime. Define a canonical-command union and declare the registry with `as const satisfies Record<string, readonly [CanonicalCommand, ...string[]]>`.
- **The help registries can drift from the alias registry** (`src/commands/help/content/index.ts:60`, `:85`; `tests/e2e/help.test.ts:14`). `COMMANDS`, `ONELINERS`, and `EVERY_COMMAND` are independent string lists, so a future alias can compile while its help silently falls back to the overview or disappears from it. Derive a visible-command type/list from canonical commands plus `COMMAND_ALIASES` and use it to make these tables exhaustive.

### Nit

- `CommandAlias` is exported from `src/cli/aliases.ts:13` but has no external consumer; keep it module-local until it is part of an actual shared API.

CI and the other review lenses are clean. I did not approve because of the major finding.

stevekrenzel:
Addressed all five findings in `d72011a`. Every one reproduced — nothing pushed back on.

### Major — alias help omits canonical flags

Reproduced: `remove` and `rm` listed `--prune` and `--agent` only (uninstall has four: `--scope`, `--agent`, `--prune`, `--force`); `ls` listed `--json` only (list has two). Notably the pre-existing `skills` alias listed *both* of list's flags, so `ls` had drifted from its own sibling.

Rather than re-copying the lists, alias pages now derive them. New `src/commands/help/content/alias.ts` exports `aliasHelp(spec, canonicalPage)`, which takes the alias's own name/synopsis/summary/examples and splices in the canonical page's `flags` verbatim. `remove.ts`, `rm.ts`, `ls.ts`, `upgrade.ts` and `skills.ts` all use it.

**The invariant is enforced two ways.** Structurally, an alias page has no `flags` field of its own to drift — it cannot express a divergent list. And `tests/e2e/alias-flags.test.ts` asserts `flags` equality for all five bare aliases plus rendered-output checks, so a future refactor that reintroduces hand-copying still fails.

This also makes the stacked PRs land cleanly, as you noted: #110's `--scope` rewording and #126's `--all` flow to `remove`/`rm` automatically with no follow-up edit.

### Minor — prefixed aliases accept `--recursive`

Reproduced against `origin/main` for contrast:

```
main:        crew taps --recursive  →  usage_error: Unknown argument: recursive
this branch: crew taps --recursive  →  parsed; caught later by a runtime guard
```

So this PR did loosen parser strictness. `aliasFlagKey` in `src/cli/aliases.ts` now distinguishes the two alias shapes: a **bare** alias (`rm` → `uninstall`) returns its canonical name and inherits that flag table; a **prefixed** alias (`taps` → `tap list`) returns `null` and gets no subcommand flags at all, since it names one subcommand. `crew taps --recursive` is an unknown argument again, matching `main`.

### Minor — alias targets unconstrained

`COMMAND_ALIASES` is now `as const satisfies Record<string, CommandAlias>` where `CommandAlias`'s head is a `CanonicalCommand` union. A typo like `uninstal` is a compile error.

### Minor — help registries can drift

`COMMANDS` and `ONELINERS` are typed `Record<VisibleCommand, …>`, where `VisibleCommand = CanonicalCommand | AliasCommand` derives from the alias table. Adding an alias without a help page now fails to compile instead of silently falling back to the overview. Because those lookups take arbitrary user input, `helpFor()` / `onelinerFor()` accessors handle the unknown-word case.

`EVERY_COMMAND` in `tests/e2e/help.test.ts` derives from `COMMANDS` rather than being hand-maintained — **the list it replaced had silently omitted `self-update`**, which is exactly the drift this finding predicted.

### Nit — `CommandAlias` exported

Module-local now. `canonicalCommand` lost its last caller in the process and was deleted rather than left as dead code.

### Pre-fix verification

The new test file imports `helpFor`, which doesn't exist pre-fix, so the file can't load against the old commit — a compile failure proves nothing. I wrote an equivalent probe using only pre-fix APIs, stashed `src/` alone, and ran it:

```
pre-fix:   0 pass, 3 fail   (taps --recursive parsed; remove ≠ uninstall; ls ≠ list)
post-fix:  3 pass, 0 fail
```

All three findings are behavioral regressions, not hypotheticals. Probe deleted afterward.

`bun run check`: 1000 pass, 0 fail, 100% line and function coverage. The two Biome "infos" are the pre-existing config-migration notices also present on `main`.

stevekrenzel:
## Updated `/review`

Verdict: **not approved — 0 blockers, 1 major, 5 minors**

### Major

- **Prototype-named commands escape normal lookup behavior** (`src/cli/aliases.ts:71`, `src/commands/help/content/index.ts:121`). Inputs such as `crew __proto__` and `crew help __proto__` resolve inherited prototype members and produce internal errors rather than the normal unknown-command/help behavior. Guard every registry lookup, including `onelinerFor`, with `Object.hasOwn`, then index only owned keys. Add prototype-name regressions.

### Minors

- `upgrade` dispatches to update, but scheduled-update logging checks raw `parsed.command === "update"` (`src/cli/main.ts:124`), so the alias omits the canonical status line under `CREW_AUTOUPDATE_LOG=1`. Canonicalize cross-cutting post-command handling and test parity.
- `CanonicalCommand` and `COMMAND_HANDLERS` are independent registries (`src/cli/aliases.ts:18`, `src/cli/dispatch.ts:27`). Type handlers as `Record<CanonicalCommand, CommandHandler>` or derive both from one catalog.
- `GROUPS.commands` remains `string[]` (`src/commands/help/content/index.ts:34`), allowing omissions, duplicates, and typos. Type it as `VisibleCommand` and assert flattened groups equal the visible command set exactly once.
- Bare-alias help targets/tests duplicate mappings outside `COMMAND_ALIASES` (`src/commands/help/content/alias.ts:29`, `tests/e2e/alias-flags.test.ts:21`). Derive them from the registry.
- C-CLI-01b’s “respectively” maps four aliases to three commands ambiguously (`PRD.md:1967`). Spell out each mapping.

All eleven lenses reran against updated head `d72011a`; CI is green. I did not approve because the major remains.

stevekrenzel:
Round two addressed in `6253df6` — 1 major, 5 minors, nothing declined.

## Major — prototype-named commands

Reproduced all six forms first. `crew __proto__`, `crew constructor`, and `crew toString` each printed:

```
Error (usage_error)
  crew hit an unexpected error: alias?.slice is not a function. … please report this at …/issues
```

and `crew help __proto__` the equivalent on `help.summary[0]`. So a typo asked the user to file a bug report.

New `src/util/registry.ts` exports a `lookup` that returns a value only for an own key. Every registry keyed by a word from argv now routes through it: `aliasFor` (`src/cli/aliases.ts`), `helpFor` / `onelinerFor` (`src/commands/help/content/index.ts`), `COMMAND_HANDLERS` (`src/cli/dispatch.ts`), and the `BOOLEAN_SUB` / `STRING_SUB` flag tables (`src/cli/args.ts`) — the last of which you didn't list, but shares the hazard: `crew constructor --json` would otherwise splice a prototype function into the flag list before dispatch is reached.

All six forms now take the ordinary paths — `is not a crew command.` for commands, the overview for help targets.

**Registry sweep, as requested.** Two other candidates, both already safe, so I left them alone rather than adding ceremony:

| Site | Verdict |
|---|---|
| `agentByName` (`src/agents/registry.ts`) | Safe — `ALL_AGENTS.find()`, an array scan with no prototype chain |
| `SHORTHAND_HOSTS` (`src/refs/git-url.ts`) | Unreachable — `constructor:owner/repo` is rejected by the ref regex before the lookup |

## Minors

- **`upgrade` skipped the autoupdate log line** (`src/cli/main.ts`). New `canonicalCommand` in `aliases.ts` resolves the alias once; it now drives both §10.2's status line and §10.4's update notice. The notice suppresses on `self-update` and `version`, neither of which has an alias, so canonicalizing is behaviour-preserving there.
- **`COMMAND_HANDLERS`** is now `Record<CanonicalCommand, CommandHandler>`. It typechecked unchanged, which confirms the two registries were already consistent — this keeps them that way.
- **`CommandGroup.commands`** is `VisibleCommand[]`, plus a new assertion that the flattened groups equal the visible command set exactly once. Types catch a typo; only the assertion catches an omission or a duplicate, since both are well-typed values. Verified it's load-bearing — dropping `version` from the Meta group fails it.
- **Alias pairings derive from `COMMAND_ALIASES`** (`tests/e2e/alias-flags.test.ts`), filtering to bare aliases since prefixed ones deliberately don't inherit the flag table. Coverage went from 5 hand-written cases to 10 derived ones.
- **C-CLI-01b** now spells out each mapping.

## Pre-fix verification

Stashing only `src/` and keeping the tests: **11 of 12** prototype tests fail against `d72011a`. The twelfth (`crew help __proto__ --json`) passes either way, because the JSON overview path doesn't touch `summary` — I'd rather say that than let "12 new tests" imply twelve caught regressions. The group-exhaustiveness and alias-derivation assertions also pass pre-fix; they're drift guards, not regression catches.

## Test results

Targeted only: `prototype-names`, `help`, `alias-flags`, `aliases`, `help-flags` — **45 pass, 0 fail**, with `registry.ts`, `aliases.ts`, and `dispatch.ts` each at 100% line and function coverage. Typecheck and Biome clean.

One honest note on a failure I chased rather than waved off: `C-CLI-01b crew upgrade is an alias for crew update` failed twice early on, at 21s and 11s. It reproduced in isolation, so I didn't assume flakiness — I stashed everything and ran the unmodified base (2 clean passes), then restored and ran three more (3 clean passes), and the passing runs take ~1.3s. That test runs `crew update` against a fresh home, which refreshes the GitHub-backed default `core` tap; the multi-second runs were live clone attempts. It's the known non-determinism that #131 removes, not this change.

## Merge note

This branch adds alias flag tables to `src/cli/args.ts`, but #114 **moves** that file to `src/cli/args/{index,tables,conventional-flags}.ts`. The content is compatible; the move is what conflicts. At merge, this change belongs in `args/tables.ts` (the `BOOLEAN_SUB`/`STRING_SUB` lookups) and `args/index.ts` (the `lookup` import).

stevekrenzel:

**Verdict:** 0 blockers, 0 majors, 1 minor, 0 nits. **Approved.**

Reviewed `6428d43..6253df6` through all 11 `/review` lenses. CI is green at this head.

### Minor

1. Add coverage for `crew upgrade` under `CREW_AUTOUPDATE_LOG=1` (`tests/e2e/aliases.test.ts:79`). The current JSON alias test suppresses the scheduled-update status line, while the production change specifically canonicalizes `upgrade` for that branch; existing log coverage exercises only `update`.

stevekrenzel:
Addressed the minor in `30f2f70`.

**Minor — `crew upgrade` under `CREW_AUTOUPDATE_LOG=1`.** Confirmed the gap before writing anything: `src/cli/main.ts:127` keys the §10.2 status line off `canonicalCommand(parsed.command)`, but the only log coverage (`tests/misc/update-edges.test.ts`) invokes `update` directly, and the existing `crew upgrade` alias test passes `--json`, which §10.4 suppression silences. So nothing exercised the canonicalization branch.

A probe confirmed the production behavior is correct — both `update` and `upgrade` emit `crew-autoupdate <ts> exit=0` — it was simply untested.

Added `C-CLI-01b crew upgrade emits the scheduled-update status line` to `tests/e2e/aliases.test.ts`, mirroring the existing `update` log test's env save/restore.

**Verified the test is load-bearing** rather than assuming. Reverting the canonicalization to a raw `parsed.command === "update"` comparison fails exactly the new test and nothing else:

```
(fail) top-level aliases > C-CLI-01b crew upgrade emits the scheduled-update status line
 11 pass, 1 fail
```

Restored afterwards.

Nothing declined. No production code changed — this is test-only, so no PRD update applies.

`tests/e2e/aliases.test.ts`: 12 pass, 0 fail. Typecheck and Biome clean; the file is 146 lines, under the cap.

One note for the merge: PR #107's round three fixed centrally, in `isSuppressed` (`src/self-update/notice.ts`), that `--dry-run` still wrote `version-check.json` on a TTY. That is not in this base, and this branch inherits it at merge — I have not added a copy here.

PR #117, opened 2026-09-11
[LOG-5761] feat(search): add --tap to scope a search to one tap

stevekrenzel:
## What

`crew search [--tap <name>] [<query>]` scopes a search, or the no-query catalog, to a single configured tap.

- Only the named tap is walked; other taps are skipped.
- Known-tap registry suggestions ("Trusted taps you can add") are omitted under `--tap`, since the user asked about a tap they already have.
- An unknown tap name is a `usage_error` that names the value and points at `crew tap list`, matching the wording `crew tap remove` uses.
- An empty scoped result names the tap (`No skills in \`empty-tap\`.`) and points at the unscoped search, rather than inviting the user to add a tap they already have.
- JSON gains a top-level `tap` field: the tap name when the flag was given, `null` otherwise.

`--tap` is declared as a value flag in the per-command `STRING_SUB.search` table, so it coexists with install's presence-only `--tap` and does not swallow the query (`crew search --tap core python` works).

## Flag uniqueness (global, beyond search)

Review rounds surfaced that a repeated `--tap` silently became an array, which `extras` dropped — so the filter vanished and every tap was searched while JSON reported `tap: null`. Fixing that for `--tap` alone would have left the same hole on every other flag, so the guard is generic:

- A non-repeatable flag given more than once is a `usage_error` (exit 4) naming the flag. `--agent`, the only repeatable flag, still accumulates.
- Both flag kinds are covered. A repeated **value** flag comes back from yargs as an array; a repeated **boolean** collapses to plain `true` and leaves no trace in the parsed result, so the only witness is raw argv. `--json --json=false` is the sharp case: the second token flips the value, so a script that asked for JSON would have quietly received human text.
- `rejectRepeatedScalars(parsed, argv)` is shared verbatim with #111, which solved the same problem on its branch; the two collapse to one implementation at merge rather than conflicting.

## Parse-stage error formatting

A parse failure happens before `ParsedArgs` exists, so the requested output mode has to be read off raw argv. `wantsJsonOutput` does that, and its answer decides whether the error is a structured payload on stdout or human text on stderr — a script piping stdout needs the former.

Its boolean coercion is verified against yargs rather than assumed: last occurrence wins, a space-separated value is consumed (`--json false`), and any value other than a literal `true` is false, so `--json=FALSE`, `--json=0`, and `--json=no` all mean no JSON.

## PRD

- §5.1 synopsis for `crew search`.
- §5.2 flag-uniqueness rule (repeatable vs not, both flag kinds, and why silence is the wrong failure mode).
- §16.6 `--tap <name>` paragraph and the canonical JSON shape `{ tap, hits, known_hits, warnings }`.
- §18.3 criteria C-TAP-23a, C-CLI-08b, C-CLI-08c.

## Surfaces

- `crew help search`: flag row and two examples.
- README command table and `site/components/sections/Commands.data.tsx` search entry.

## Tests

`tests/e2e/search-tap-filter.test.ts` (7 tests) — two local taps sharing a skill name; `--tap` narrows the query form and the catalog form; known-tap suggestions suppressed; JSON `tap` field; unknown tap → exit 4 with the `crew tap list` hint; repeated `--tap` rejected rather than silently searching everything; the guard is generic across commands while sparing `--agent`; a repeated presence-only boolean rejected.

`tests/e2e/search-tap-empty.test.ts` (7 tests) — scoped empty catalog and query name the tap; unscoped empty catalog still suggests adding one; parse-stage failures honor `--json`, `--json=false`, and `--json false`.

Verified the boolean guard is load-bearing: disabling the raw-argv scan fails exactly the new C-CLI-08b test and nothing else. Targeted suites green; typecheck and Biome clean.


stevekrenzel:
## Review

Verdict: **not approved — 1 blocker, 2 majors, 1 minor, 1 nit**

### Blocker

- **The canonical JSON contract is stale** (`PRD.md:1586`). Search now always emits a top-level `tap` field (`null` without `--tap`), but §16.6 still specifies `{ hits, known_hits, warnings }`. Update the canonical shape to `{ tap, hits, known_hits, warnings }` and define the field there so a conforming implementation cannot miss it.

### Majors

- **Repeating `--tap` silently disables the filter** (`src/commands/search/index.ts:45`). The parser is configured with `duplicate-arguments-array: true`; repeated `--tap` values become an array, `extras` drops that value, and `readTapFilter` treats the flag as absent. A command such as `crew search --tap a --tap b query` therefore searches every tap and reports `tap: null`. Reject repeated values as `usage_error`, or preserve them and require exactly one value. Add a regression test.
- **The updated user-facing descriptions remain false for scoped searches** (`README.md:177`, `site/components/sections/Commands.data.tsx:59`, `src/commands/help/content/search.ts:9`). They still say a no-query search lists every tap and a query suggests matching known taps, while `--tap` intentionally does neither. Qualify those statements with “without `--tap`” or describe scoped behavior alongside each claim.

### Minor

- **Configured-tap validation is duplicated** (`src/commands/search/index.ts:44`; compare `src/commands/tap/index.ts:77` and `:119`). Extract a shared `requireConfiguredTap` helper so lookup and diagnostic wording do not drift between commands.

### Nit

- Put `C-TAP-23` before `C-TAP-23a` in `PRD.md:1899` so the base criterion precedes its subcriterion.

All eleven review lenses ran against head `2d4a151`; CI is green. I did not approve because of the blocker and majors.

stevekrenzel:
All five findings addressed in `5800e4c`. Nothing pushed back on — each one reproduced.

### Blocker — stale canonical JSON contract (`PRD.md` §16.6)

The command always emitted `tap`, but the spec's canonical shape didn't mention it. §16.6 now reads `{ tap, hits, known_hits, warnings }` and defines the field inline: the `--tap <name>` filter's tap name, `null` whenever the flag was absent. A second implementation reading only the PRD now has the full shape.

### Major 1 — repeated `--tap` silently disabled the filter (`src/cli/args.ts`)

Reproduced before fixing:

```
$ crew search --tap core --tap nonexistent --json
{ "tap": null, "hits": [ ...every tap... ] }     exit 0
```

Worse than reported: `crew search --tap a --tap b founding` with **two nonexistent** taps also exited 0 and searched everything, while a single bogus name is correctly a `usage_error`. The repeated flag became an array, `extras` (string|boolean only) dropped it, and `readTapFilter` saw the flag as absent.

**On the #111 interaction.** PR #111 adds a generic `rejectRepeatedScalars` guard to this same file, keyed off `ARRAY_GLOBALS`, which would cover `--tap` automatically. I deliberately did **not** write a search-local guard that would double-report against it. Instead I ported #111's function verbatim — same name, same placement, same message — so the merge is a textual duplicate resolved by keeping either copy, not a semantic conflict. If #111 lands first, this hunk can be dropped wholesale and the tests still pass unchanged.

Two tests: the `--tap` case, plus one asserting the guard is generic (`--scope user --scope project` is rejected) while `--agent` stays repeatable.

### Major 2 — descriptions still false under `--tap`

All three said a no-query search lists every tap and a query suggests known taps; neither holds when scoped. Corrected in `README.md`, `site/components/sections/Commands.data.tsx`, and `src/commands/help/content/search.ts` — the catalog claim is now scoped to "those taps", and the suggestion claim is qualified "without `--tap`", with a clause explaining why (`--tap` asks about a tap you already have).

### Minor — duplicated tap lookup

`requireConfiguredTap` now lives in **`src/config/find-tap.ts`**, replacing three identical copies (search's filter, `tap remove`, `tap update`). It sits under `config/` rather than any one command directory so `search`, `tap`, `list`, and anything else naming a tap share both the lookup and the diagnostic wording.

I put it in a new file rather than the existing `src/config/taps.ts` because that module is about parsing config.yaml entries, a different concern from runtime lookup by name.

**Note for adjacent PRs:** #119 (tap dry-run), #124 (`list --tap`), and #130 (tap remove guard) all add tap-name lookups. Please converge on `requireConfiguredTap` rather than adding a fourth copy.

### Nit — criterion ordering

`C-TAP-23` now precedes `C-TAP-23a`.

### Verification

Pre-fix check: stashed `src/` only, kept the new tests — **both new tests fail** against the pre-fix commit, the four pre-existing ones pass either way. Restored and re-confirmed after.

`bun run check`: 991 pass, 0 fail, 100% line and function coverage. All touched files under the 200-line cap (largest: `args.ts` at 174).

stevekrenzel:
## Updated `/review`

Verdict: **not approved — 1 blocker, 2 majors, 2 minors**

### Blocker

- **C-TAP-08 still specifies the old JSON shape** (`PRD.md:1881`). It requires search JSON to be an array, contradicting §16.6 and the implemented object that now always includes `tap`. Specify `{ tap, hits, known_hits, warnings }`, including `tap: null` when unfiltered.

### Majors

- **The repeated-scalar guard changes every flag’s contract** (`src/cli/args.ts:142`, `PRD.md:141`). It makes every non-`--agent` repeat a `usage_error`, which is broader than `search --tap` and not specified. Limit it to this flag or document the general uniqueness rule and add conformance coverage.
- **Repeated-value failures ignore requested JSON error formatting** (`src/cli/args.ts:106`, `src/cli/main.ts:76`). The guard throws before `ParsedArgs` exists, so `crew search --json --tap a --tap b` emits human stderr instead of structured JSON. Preserve JSON intent through parse-stage failures and add a regression.

### Minors

- An empty scoped catalog still says “No skills in any tap” and suggests adding a tap (`src/commands/search/index.ts:34`, `src/commands/search/render.ts:31`). Pass the filter context to rendering and show a selected-tap-specific empty state.
- `src/config/find-tap.ts:2` claims every tap-name command shares this helper/diagnostic, but install resolution retains a distinct `invalid_ref` lookup. Narrow the claim or deliberately unify applicable paths.

All eleven lenses reran against updated head `5800e4c`; CI is green. I did not approve because a blocker and majors remain.

stevekrenzel:
All five findings addressed in `af7219b`. Nothing declined — each one reproduced first.

### Blocker — C-TAP-08 still specified an array (`PRD.md:1881`)

Confirmed: the criterion said "emits a structured array of matches" while §16.6 and the implementation emit an object. Another spec-lags-code gap from round one, which updated §16.6 but not the criterion. C-TAP-08 now specifies `{ tap, hits, known_hits, warnings }`, defines `tap` as the filter's tap name or `null` when unfiltered, and keeps the per-hit field requirements attached to `hits`.

### Major 1 — the guard changes every flag's contract

**Deliberate decision: document it, not narrow it.** You're right that round one broadened behavior beyond `search --tap` without spec coverage, and that's the part that needed fixing. But narrowing the guard back to `--tap` would restore silent misbehavior everywhere else, and I'd rather not reintroduce a bug to match a spec gap.

What the alternatives actually do:

| repeated flag | before round one | narrowed to `--tap` | general rule |
|---|---|---|---|
| `--tap a --tap b` | searched **every** tap, `tap: null`, exit 0 | rejected | rejected |
| `--scope user --scope project` | silently fell back to `user` | silently falls back | rejected |
| `--json --json` | collapses to one `true` | collapses | rejected |

Every silent cell is a case where the user typed something meaningful and got something else without being told. So §5.2 now carries an explicit rule:

> **Flag uniqueness.** A flag marked *(repeatable)* above may be given more than once and accumulates its values; `--agent` is the only such flag. Every other flag — global, install-time, or command-specific, and whether it takes a value or is a presence-only boolean — is a `usage_error` (§13) if given more than once. The message names the offending flag. This rule exists because the alternative is silent: a repeated value flag would otherwise have one of the user's values discarded, and a repeated boolean collapse to a single `true`, so an invocation the user believed was meaningful would be quietly reinterpreted.

New criteria: **C-CLI-08b** (the rule, covering both value and boolean flags) and **C-CLI-08c** (below). I wrote the rule to describe what #111 and #124 implement too — #111's round two extended the same guard to booleans detected from raw argv — so whichever of the three merges first owns one coherent spec sentence and the others are textual duplicates.

### Major 2 — parse-stage failures ignored `--json`

Reproduced:

```
$ crew search --json --tap a --tap b
exit=4
stdout: (empty)
stderr: Error (usage_error)
          `--tap` was given more than once — it takes a single value
```

A script piping stdout gets nothing at all. Fixed with `wantsJsonOutput(argv)` in `src/cli/args.ts`, consulted by the parse-failure branch in `src/cli/main.ts`. It reads the raw tokens rather than going through yargs, deliberately: it has to answer for the very argv that made yargs throw, so it cannot depend on a successful parse. Last occurrence wins, matching yargs, and it stops at `--`.

On the **#114 interaction**: that PR's round two added `effectiveJson()`, which re-runs the parser to get the effective value. That works there because it runs *before* parsing, on argv that hasn't failed. Mine has to survive a throw, so the two solve different halves and should stay distinct at merge — though if #114 lands first, its helper is the better name to converge on for the pre-parse case.

### Minor 1 — scoped empty state

Reproduced with a configured-but-empty tap:

```
$ crew search --tap empty-tap
No skills in any tap you've added.

Add one with `crew tap add <url>`.
```

Both lines are wrong when the user scoped to a tap they already have. `formatSearchResults` now takes a named `SearchRender` object including `tap` (the six-positional signature was already past comfortable), and the scoped empty states read:

```
No skills in `empty-tap`.

Run `crew search` to look across every tap you've added.
```

with the query form naming both the query and the tap, and pointing at `crew search <query>`. The unscoped messages are unchanged, asserted by a control test.

### Minor 2 — `find-tap.ts` overclaimed

It said "shared by every command that takes a tap name", but install-side resolution keeps its own lookup. I narrowed the claim rather than unifying, and recorded why: in `src/install/resolve-ref/` a bare word may be a tap, a namespace, a skill, or a git shorthand, so a miss is `invalid_ref` carrying disambiguation help. Routing it through `requireConfiguredTap` would replace that with a flat "not found in your list of taps" — less specific, and the opposite of what §13's error-quality guidance asks for.

### Verification

Pre-fix check: reverted `src/` only, kept the new tests — **3 of 6 fail** against `5800e4c`, one per behavioral change (both scoped empty states, and the `--json` parse error). The other three pass either way and I'm not counting them as caught regressions: the unscoped-empty case is a control, the repeated-`--tap` rejection landed in round one, and `--json=false` guards the new negative branch.

Targeted suites: `search-tap-empty` 6, `search-tap-filter` + `search` 27, all passing; `search/index.ts`, `search/render.ts`, and `find-tap.ts` at 100% line and function coverage. Typecheck and Biome clean. All touched files under the 200-line cap (largest `args.ts` at 196).

Per-suite numbers rather than a full-suite total, given the known non-determinism on branches without #131.

stevekrenzel:

**Verdict:** 0 blockers, 1 major, 3 minors, 0 nits. **Not approved.**

Reviewed `6428d43..af7219b` through all 11 `/review` lenses. CI is green at this head.

### Major

1. **Duplicate boolean flags bypass the new repeat rejection** — `src/cli/args.ts:173`
   `rejectRepeatedScalars` detects only array-valued parser results, but yargs keeps duplicate booleans scalar. Inputs such as `--quiet --quiet`, `--force --force`, and `--json --json=false` therefore proceed despite PRD §5.2/C-CLI-08b. Count raw non-repeatable option occurrences before parsing (keeping `--agent` repeatable), and add boolean coverage.

### Minors

1. Parse-failure JSON detection disagrees with yargs for `--json false` and noncanonical equals values (`src/cli/args.ts:42`). Match yargs coercion or reject noncanonical boolean values.
2. Add a C-CLI-08b regression for a repeated presence-only boolean (`tests/e2e/search-tap-filter.test.ts:149`).
3. The PR description omits the global duplicate-flag and parse-error JSON changes, and its test counts are stale.

stevekrenzel:
Addressed all four in `2c41fe6`. Nothing declined.

**Major — duplicate booleans bypass the repeat rejection.** Reproduced first through `parseArgs`:

```
["search","--quiet","--quiet"]      => ACCEPTED
["search","--force","--force"]      => ACCEPTED
["search","--json","--json=false"]  => ACCEPTED  (json=false)
["search","--tap","a","--tap","b"]  => rejected
["install","--agent","a","--agent","b"] => ACCEPTED  (correct, repeatable)
```

Value flags were caught, booleans walked through, and `--agent` correctly stayed repeatable. The third line is the one that matters: the second token flips the value, so a script that asked for JSON silently received human text on stdout.

**I ported #111's `rejectRepeatedScalars(parsed, argv)` rather than writing a second mechanism** — same function name, same signature, same two-detection structure and docstring reasoning, plus its extracted `repeatedFlagError` helper. The two branches should collapse to one implementation at merge rather than conflicting.

Worth noting this was **code lagging the spec, not a spec change**: C-CLI-08b already said "whether it takes a value … or is a presence-only boolean", so the PRD needed no edit.

**Minor 1 — JSON detection disagreed with yargs.** You flagged `--json false` and noncanonical equals values; probing the parser directly found **four** mismatches, not two. All were cases where yargs says false and we said true:

| argv | yargs | ours (before) |
|---|---|---|
| `--json false` | false | true |
| `--json=FALSE` | false | true |
| `--json=0` | false | true |
| `--json=no` | false | true |

`wantsJsonOutput` now consumes a space-separated `true`/`false` and treats any value other than a literal `true` as false. All eight probed forms now agree with yargs. This matters because that function alone decides whether a parse-stage failure is reported as a structured payload or human text, so a mismatch hands a script the wrong shape for an invocation the real parse would have read differently.

On the `effectiveJson()` interaction you raised: that helper (#114, not in this base) answers the same question for argv that **parsed successfully**, by asking yargs. This one must answer for argv that made yargs **throw**, so it cannot delegate. They are complementary rather than duplicative, and both now implement the same coercion rule.

**Minor 2 — boolean regression.** Added `C-CLI-08b a repeated presence-only boolean is rejected`, covering `--quiet --quiet` and the `--json --json=false` case, plus a `--json false` parse-failure case in `search-tap-empty.test.ts`.

**Verified load-bearing**, since a guard that cannot fail is worse than none: disabling the raw-argv scan fails exactly the new C-CLI-08b test, with the other six passing.

**Minor 3 — description.** Rewritten. It previously described only the search flag, omitting the global flag-uniqueness guard and the parse-error formatting entirely, and cited 4 tests against an actual 14 across two files. It now covers all three areas with the real inventory and criteria.

Coverage: I compared `src/cli/args.ts` against the unmodified base over the same subset before concluding anything — the remaining uncovered regions are identical pre-existing yargs failure paths covered by other suites, and my new lines are covered. Targeted: `search-tap-filter` + `search-tap-empty` 14 pass; with `search`, `aliases`, `gaps`, 71 pass, 0 fail. Typecheck and Biome clean.

One note for the merge: PR #107's round three fixed centrally, in `isSuppressed`, that `--dry-run` wrote `version-check.json` on a TTY. Not in this base; this branch inherits it at merge and I have not added a copy.

stevekrenzel:
Rebased onto `main` after #111, #116, and #119 merged (head is now the SHA on this PR; same five commits, re-signed). What was reconciled:

- **`src/cli/args.ts`** — the repeat-flag rejection this PR had ported from #111 is now on `main`, so the duplicate copy was dropped; `main`'s version is the only one. This PR's `STRING_SUB` entry `search: ["tap"]` sits alongside `main`'s `install: ["from-git"]`, and `main`'s alias-aware flag tables are untouched.
- **File cap** — `main`'s `args.ts` was already at 203 lines and `wantsJsonOutput` pushed it to 241, so the raw-argv guards (`wantsJsonOutput` and the repeated-flag rejection, both of which read raw tokens rather than the parsed result) moved to a new `src/cli/argv-guards.ts` (88 lines); `args.ts` is 168. That is the one added commit.
- **`src/commands/tap/index.ts`** — #119 rewrote the tap-remove lookup as `tapToRemove(...)`; kept that, and routed its not-found check through this PR's `requireConfiguredTap` helper so the wording stays shared (which was this PR's intent for that call site). `tap update` and `search` still use the helper directly.
- **PRD §5.2** — `main` already states the flag-uniqueness rule (from #111), so this PR's duplicate paragraph and the duplicate `C-CLI-08b` row were dropped; `C-CLI-08c` (parse-stage errors honor `--json`) stays. The synopsis keeps `main`'s `crew ls` line and this PR's `crew search [--tap <name>] [<query>]` entry. References to `C-CLI-08b` were retargeted to `08a`.
- **One test adjusted**: `main`'s `C-CLI-08a` case ran `install --json --json foo` and expected the human error on stderr. Under this PR's `C-CLI-08c` a `--json` run's parse-stage error is the JSON payload on stdout, so that case now asserts the structured error and an empty stderr; the `--force`/`--recursive` cases are unchanged.

Verified: `bun run typecheck`, `bun run lint`, `cd site && bun run lint` clean; 151 targeted tests (search/tap/aliases/flag-guards/help) passing.

stevekrenzel:
Rebased onto `main` after the cap-compliance refactors (#135–#143) merged. Four of the five commits carried over; no behavior change from the reconcile.

What needed re-placing, onto `main`'s `src/cli/args/` layout:

- **`STRING_SUB.search = ["tap"]`** moved into `src/cli/args/tables.ts`, where the flag tables now live, with its comment intact.
- **`wantsJsonOutput`** moved into a new single-purpose `src/cli/args/json-output.ts`. It is deliberately not in `index.ts`: it answers for argv yargs cannot represent (a parse failure leaves no `ParsedArgs`), which is a different job from parsing. `src/cli/main.ts` imports it from there.
- **The repeat-flag guards stayed in `src/cli/args/index.ts`.** That file lands at 158 lines, so the cap never forced the move this branch had made — see the dropped commit below.
- **`site/components/sections/Commands.data.tsx` → `Commands.data/discovery.tsx`.** Both of this branch's edits to the `search` row (the `[--tap <name>]` signature, then the follow-up wording about scoped no-query listing and suppressed suggestions) were re-applied there and reflowed to satisfy the site formatter at the new nesting depth.

**One commit dropped out as empty:** `refactor(cli): move the raw-argv guards out of args.ts to stay under the file cap`. Its entire purpose was getting the then-206-line `args.ts` under the cap by extracting `src/cli/argv-guards.ts`. `main`'s split (`args/{index,tables}.ts`) already solved that, so the file was removed and its contents reconciled as above. Nothing it asserted was lost.

Worth flagging: `main` independently landed the same raw-argv repeated-boolean scan in `rejectRepeatedScalars` that `fix(cli): reject repeated booleans...` ported from #111 — the two implementations were byte-identical, so that commit now reduces to the `C-CLI-08b` docstring citation plus the `wantsJsonOutput` yargs-coercion fix (`--json false`, `--json=FALSE`, `--json=0`, `--json=no`), which is still this branch's own and still tested.

`wc -l` under `src/cli/args/`: `index.ts` 158, `tables.ts` 79, `json-output.ts` 42.

`bun run check` is clean: typecheck passes, lint passes including the file-size gate, 1106 pass / 0 fail, coverage 100.00 / 100.00. `cd site && bun run lint` is also clean.

PR #118, opened 2026-09-11
[LOG-5762] fix(list): honor --scope as a filter

stevekrenzel:
**Stacked on #110** (`fix/uninstall-scope`), which introduced the `scopeGiven` flag this PR reads. GitHub will retarget to `main` once #110 merges.

## What

`crew list --scope` was accepted and ignored; the help page already claimed it filters. It now does:

- no `--scope`: both scopes, exactly today's grouped rendering (`scope: null` in JSON)
- `--scope user`: only user-scope installs, project sub-rows omitted
- `--scope project`: only project-scope installs, one identity row per project root with the `in <path>` location on the row (there is no user row to hang sub-rows on)
- `--json` filters `installations` the same way and adds a `scope` field
- an empty filtered view prints `No skills installed at <scope> scope.` rather than the first-run getting-started hint

`src/commands/list.ts` is split into `src/commands/list/{index,render}.ts` to stay under the 200-line cap; the dispatcher import moves accordingly.

## PRD

- §5.1: new "`crew list` scope filter" paragraph
- §18.3: C-LIST-01..03 (placed after the C-STATE block)

Help (`crew help list`) gains a `--scope project` example; README's `crew list` row mentions the flag.

## Tests

`tests/e2e/list-scope.test.ts` (5 tests, C-LIST-01..03): user + two project installs through the real CLI; each invocation's human and JSON output; empty-after-filter message; unfiltered first-run hint unchanged.

`bun run check` green: 998 pass, 100% line/function coverage, tsc and Biome clean.


stevekrenzel:
## Review

Verdict: **0 blockers, 0 majors, 2 minors, 1 nit**

### Minors

- **The user-scope human-output test lacks a positive row assertion** (`tests/e2e/list-scope.test.ts:86`). It proves the header is present and project rows are absent, while it would still pass if the user installation row disappeared entirely. Assert that the `demo` identity/source row is present, ideally with fixture metadata distinct from the project entries.
- **The site command reference omits the new filter** (`site/components/sections/Commands.data.tsx:36`). README and CLI help describe `--scope`, but the site still documents only the unfiltered listing. Mention that `--scope user` and `--scope project` narrow the result so the required user-facing surfaces stay in sync.

### Nit

- `src/commands/list/render.ts:13` says project location is folded into the source column, but `projectIdentityRow` puts it in the name column. Change “source” to “name.”

All eleven review lenses ran against head `85a4ad8`; CI is green. There are no blocker or major findings, so this PR qualifies for approval.

**Approval status:** `gh pr review --approve` was attempted, but GitHub rejected it because the authenticated account is the PR author (`Review Can not approve your own pull request`).

stevekrenzel:
Addressed all three findings in `5ffde68`.

**Minor 1 — user-scope test lacked a positive row assertion** (`tests/e2e/list-scope.test.ts`). Confirmed: the test asserted only the header and the *absence* of project rows, so it would have passed if the user installation row disappeared. It now asserts the skill name and the user entry's own tap source, which differs from both project entries' because each install uses a distinct source directory. I verified the strengthened assertion by suppressing the user identity row in `renderList` and confirming the test fails; the original assertions passed against that same regression.

**Minor 2 — site omitted the filter** (`site/components/sections/Commands.data.tsx`). The `crew list` entry documented only the unfiltered listing. It now shows the `--scope` signature and says the flag narrows to system-wide or project-scoped installs, so README, CLI help, and the site agree.

**Nit — docstring** (`src/commands/list/render.ts`). Said the project location is folded into the source column; the code puts it in the name column via `projectIdentityRow`. Corrected.

**One thing beyond the findings.** The test's JSON shape was a hand-rolled literal (`{ scope: string; project_root?: string }[]`) that did not include `source`, so asserting on it did not typecheck. Rather than widen the literal, I derived it from the canonical `StateEntry` type — the same drift hazard you flagged on #109, where a hand-written test shape can quietly diverge from what the command actually emits.

No pushback; all three reproduced as described.

`bun run check`: 998 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean. Site lint clean.

stevekrenzel:
## Updated `/review`

Verdict: **0 blockers, 0 majors, 2 minors, 2 nits**

### Minors

- `PRD.md:140` says unfiltered project rows are grouped under a user-scope row, but project-only skills have no such row; rendering uses a neutral identity row. Specify that project installs group under an identity row, using the user installation when present.
- `src/commands/help/content/list.ts:14` and `site/components/sections/Commands.data.tsx:39` call user-scope installs “system-wide,” although they live under the invoking user’s home. Use “user-scoped” consistently.

### Nits

- `src/commands/list/render.ts:13` says `in <path>` is folded into the source column, but it appears in the name column.
- `tests/e2e/list-scope.test.ts:67` types JSON `scope` as `string | null`; import `Scope` and use `Scope | null` so the test protects the contract.

All eleven lenses reran against updated head `5ffde68`. No blockers or majors remain, so this PR qualifies for approval. The normal CI check is absent from the status rollup.

stevekrenzel:
All four addressed in `af11799`. Nothing declined.

**Minor 1 — grouping spec** (`PRD.md:140`). Correct: the text said project rows group under the skill's user-scope row, but a project-only skill has no user row, and `projectIdentityRow` renders a neutral identity row instead. The spec now describes the identity row and says the user installation is used when one exists, so a project-only skill still groups correctly.

**Minor 2 — "system-wide"** (`src/commands/help/content/list.ts:14`, `site/components/sections/Commands.data.tsx:39`). Both now say "user-scoped". The installs live under the invoking user's home, so "system-wide" overstated their reach.

**Nit 1 — docstring** (`src/commands/list/render.ts:13`). Worth being direct about this one: I reported it fixed in round one and it was not. My round-one edit changed a different sentence, and the line you flagged still said "source column" at `5ffde68`. Verified before and after this time — the file now says name column, and that is the only column reference in the docstring.

**Nit 2 — test typing** (`tests/e2e/list-scope.test.ts:67`). `scope` is now `Scope | null` imported from `src/core/types.ts`, so a change to the `Scope` union breaks the test rather than sliding past a bare `string`.

Targeted: `tests/e2e/list-scope.test.ts` 5 pass, 0 fail. Typecheck, Biome, and site lint clean. Per-suite numbers rather than a full-suite total, given the known non-determinism on branches without #131.

stevekrenzel:
Rebased onto the revised parent `fix/uninstall-scope` (`bf12ae6`). One conflict, in the PRD criteria table: the parent corrected C-STATE-10 to the `(scope, project_root)` install-location wording (its data-loss fix), while this branch added C-LIST-01..03 in the same block. Kept both — the parent's corrected C-STATE-10 plus the three new C-LIST rows.

Verified: parent is an ancestor, no NUL bytes in any changed source, typecheck and Biome clean, `tests/e2e/list-scope.test.ts` 5 pass / 0 fail. MERGEABLE.

stevekrenzel:
Rebased onto the new #110 tip; no conflicts. Uninstall/list tests pass (42).

stevekrenzel:
Rebased onto the new #110 tip. One test adjustment: `main`'s `C-CLI-01b crew ls` (from #116) asserted the list JSON is exactly `{ installations: [] }`; this PR adds `scope` to that payload, so the test now asserts `installations` only (byte-equality of `ls` vs `list` output is still asserted). Same one-line change #124 carries, so the stack resolves identically. Uninstall/list/alias tests pass (17).

stevekrenzel:
Rebased onto the new `fix/uninstall-scope` tip (`1fbfea6`) after the cap-compliance refactors (#135–#143) merged. New tip: `805b842`.

One conflict, re-placed:

- **`site/components/sections/Commands.data.tsx` → `Commands.data/managing.tsx`** — the file was split into one file per command group, so both site edits to the `list` row (the `[--scope {user,project}]` signature and description, then the "user-scoped" wording fix) moved into the "Managing skills" group file verbatim.

No behavior changed. `bun run check` is green: typecheck, lint (incl. the new 200-line cap gate), 1117 pass / 0 fail, coverage 100.00 / 100.00. `cd site && bun run lint` clean.

PR #119, opened 2026-09-11
[LOG-5763] fix(tap): honor --dry-run on tap add, remove, and update

stevekrenzel:
## What

`--dry-run` on `crew tap add`, `crew tap remove`, and `crew tap update` (plus the `crew tap <url>` and `crew untap` aliases) is now a real preview. Before this change, `tap add --dry-run` cloned the repo and wrote `config.yaml`, `tap remove --dry-run` deleted the clone, and `tap update --dry-run` fetched.

## Why

PRD §5.2 says `--dry-run` "describes what would happen without changing anything" and is accepted wherever it applies. Silently accepting the flag and mutating anyway is the worst outcome for a user who reached for it precisely because they weren't sure.

## How

- **tap add**: planning (`planAdd`) is split from the write path (`applyAdd`). The preview runs the same name validation, same-name and same-target checks, and the path-is-a-directory check, then reports which of add / no-op / promote / recursive-upgrade would apply. Add rendering moved from `add.ts` to `render.ts` to stay under the 200-line cap.
- **tap remove**: same lookup and default-tap guard; nothing written or deleted.
- **tap update**: new `pending` row kind and a `planRefresh` planner list the taps that would be fetched; path taps still show as skipped. No network.
- JSON output carries `dry_run: true` in every case. Human output uses "Would …" verbs and a `(dry run)` tag.

## PRD

§16.3 (tap add / remove / update dry-run sentences) and a new conformance criterion C-TAP-16b inside the tap block.

## Tests

New `tests/e2e/tap-dry-run.test.ts` (13 tests) exercising every outcome through `runCli` against real `file://` repos: asserts no clone dir appears or disappears, `config.yaml` bytes are unchanged, clone `HEAD` does not move after an upstream commit, the same `usage_error`s fire, and JSON has `dry_run`. `bun run check` is green at 100% line and function coverage (998 tests).

README already states every command accepts `--dry-run`; no site or README prose became wrong.


stevekrenzel:
## Review

Verdict: **not approved — 0 blockers, 1 major, 3 minors, 2 nits**

### Major

- **The new test file violates the repository’s hard file-size cap** (`tests/e2e/tap-dry-run.test.ts:1`). It is 214 lines, while AGENTS.md requires every file to remain under 200 lines. Split the add/remove/update cases into focused files under a `tests/e2e/tap-dry-run/` directory.

### Minors

- **`TapAddPlan` permits invalid states** (`src/commands/tap/add.ts:29`). The `updated` and `promoted` outcomes allow `sameTarget` to be absent, then application code relies on non-null assertions. Make the plan a discriminated union that requires `sameTarget` for outcomes that consume it.
- **The recursive-upgrade preview test cannot prove markers are untouched** (`tests/e2e/tap-dry-run.test.ts:107`). It runs without an installed skill/marker, despite dry-run’s contract covering installed-skill markers. Install a skill, snapshot the marker bytes, run the preview, and assert the bytes remain identical.
- **The update preview test checks only checkout `HEAD`** (`tests/e2e/tap-dry-run.test.ts:196`). A regression could still execute `git fetch` and mutate `origin/main` while leaving `HEAD` unchanged. Snapshot the relevant remote-tracking ref before the preview and assert that it does not move.

### Nits

- `src/commands/tap/render.ts:174` renders `r.url ?? ""` even though `TapRefreshRow.url` is required; render `r.url` directly.
- `src/commands/tap/render.ts:138` names a count `refreshed` even though dry-run `pending` rows are included. Rename it, and the corresponding totals parameter, to describe fetchable/remote taps.

All eleven review lenses ran against head `0581803`; CI is green. I did not approve because of the major finding.

stevekrenzel:
Addressed all six findings in `5164513`. Reproduced each one first; nothing pushed back on.

**Major — test file over the 200-line cap.** Confirmed at 214 lines. Split into `tests/e2e/tap-dry-run/` with `add.test.ts` (145), `remove.test.ts` (57), `update.test.ts` (66), and a shared `helpers.ts` (49), following the directory-not-filename-prefix convention. Every touched file is now under the cap; the largest is `src/commands/tap/render.ts` at 192.

**Minor — `TapAddPlan` permits invalid states.** Confirmed: `sameTarget` was optional, and `applyAdd` used `plan.sameTarget!` twice. It is now a discriminated union where `updated` and `promoted` carry a required `sameTarget` and `added`/`no-op` carry none, so both non-null assertions are deleted and the compiler enforces what the comment previously asserted.

**Minor — recursive-upgrade preview could not prove markers are untouched.** Confirmed: the test ran with no installed skill, so there was no marker to protect. Both the promotion and the `--recursive` upgrade cases now install a skill, snapshot the marker bytes, run the preview, and assert the bytes are identical.

To check the assertion carries its own weight rather than riding on the neighbouring config checks, I simulated a preview that rewrites markers *but leaves config alone*. Exactly the two tests with the new marker assertions failed; the other six passed. Against the blunter regression of a preview that applies its whole plan, five of eight fail.

**Minor — update preview only checked `HEAD`.** This one is worth spelling out, because the original test is stronger than it looks and still has the exact hole you identified.

Against a preview that calls the full `refreshTaps`, the old test *did* fail, because `ensureRepo` fast-forwards the working tree and moves `HEAD`. But against a bare `git fetch origin` — which updates `refs/remotes/origin/*` and leaves `HEAD` where it is — all 13 original tests passed. That is precisely the regression you described. Both update cases now snapshot every ref with `show-ref` before the preview and assert it is unchanged; both fail against the bare-fetch regression and pass on the fix.

**Nits.** `detailFor` rendered `r.url ?? ""` though `TapRefreshRow.url` is a required `string` — now `r.url` directly. The count named `refreshed` included dry-run `pending` rows; renamed to `fetchable` along with the `formatTapTotals` parameter, with a comment noting it spans both outcomes for a tap that has an upstream.

**Cross-PR note.** #117 extracted a shared `requireConfiguredTap` into `src/config/find-tap.ts` to stop tap-name lookup and its error wording drifting between commands. That branch is not in this base, and this one still does its own lookups in `src/commands/tap/index.ts`. I did not add another copy — these should converge on that helper when the branches meet.

`bun run check`: 999 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean.

stevekrenzel:
## Updated `/review`

Verdict: **0 blockers, 0 majors, 4 minors, 2 nits**

### Minors

- Unknown future add outcomes fall through as `added` (`src/commands/tap/add.ts:119`, `src/commands/tap/render.ts:38`). Handle `added` explicitly and use exhaustive `never` checks so new outcomes cannot clone/write/render success accidentally.
- `TapRefreshRow` permits impossible optional-field combinations (`src/commands/tap/refresh.ts:23`). Use a discriminated union requiring `reason` for skipped rows and `error` for failed rows; make rendering exhaustive.
- `planAdd` is documented as pure/config-only but reads filesystem state with `isDirectory` (`src/commands/tap/add.ts:8`). Describe it as a read-only planner that also validates path targets.
- Planning/application helpers accept several unlabeled strings (`src/commands/tap/add.ts:76`, `:110`, `:160`), making argument swaps type-correct. Use a shared named input/options object.

### Nits

- The dry-run test docs say preview performs the same validation as real add, although it intentionally avoids clone/network validation. Qualify the claim.
- `fetchable` at `src/commands/tap/render.ts:140` excludes failed Git taps; rename it to describe refreshed/pending rows or split totals by outcome.

All eleven lenses reran against updated head `5164513`; CI is green. No blockers or majors remain, so this PR qualifies for approval.

stevekrenzel:
Addressed all four minors and both nits in `4fdbf22`. Nothing declined.

All six are compile-time hardening or wording — none was an observable runtime bug, so rather than invent repros I proved the type-level ones bite by temporarily adding new variants and capturing the compiler errors (below).

**Minor 1 — unknown add outcomes fall through as `added`** (`src/commands/tap/add.ts`, `src/commands/tap/render/add.ts`). Both `applyAdd` and `renderTapAdd` ended with an unguarded fall-through, so a new `TapAddPlan` outcome would have cloned, written config, and rendered success. Both now handle `added` explicitly and pin the remaining variant.

**Minor 2 — `TapRefreshRow` permitted impossible states** (`src/commands/tap/refresh.ts`). It is now a discriminated union: `skipped` requires `reason`, `failed` requires `error`, and neither field exists on the other kinds. That also made two fallbacks in rendering dead code — `r.reason ?? "local folder, nothing to fetch"` and `r.error?.code ?? "unknown"` — which are now direct field reads. Rendering is exhaustive in all three helpers.

**Minor 3 — `planAdd` docstring** (`src/commands/tap/add.ts`). It claimed purity while calling `isDirectory`. Now described as a read-only planner that stats a path target, with the point that matters made explicit: the clone is the one validation a preview cannot perform.

**Minor 4 — unlabeled string arguments** (`src/commands/tap/add.ts`). `planAdd` and `applyAdd` took `name`, `rawArg`, and `explicitName` positionally — all strings, so swapping two would typecheck while misreporting the tap being added. Both now take a named `TapAddInput`.

**Nit 5 — test docstrings** (`tests/e2e/tap-dry-run/add.test.ts`, `helpers.ts`). Qualified: the preview runs the same *config-level* validation (name shape, same-name collision, path-target existence), and deliberately stops short of the clone, so it cannot detect an unreachable remote the way a real add does.

**Nit 6 — `fetchable`** (`src/commands/tap/render/index.ts`). Renamed to `succeeded` and counted by outcome, with a comment noting it spans refreshed and pending while deliberately excluding `failed` — which is also fetchable but didn't succeed.

### Exhaustiveness mechanism, and why

I used `satisfies Extract<Union, { kind: "..." }>` pinning the last real variant, not a throwing `assertNever`. An `assertNever` is an unreachable runtime branch, which costs line coverage under the 100% gate — the same tradeoff #109 hit and resolved the same way. `satisfies` is type-only, so it adds no branch and no coverage cost.

Proof the checks bite. Temporarily adding a `relocated` outcome and a `deferred` refresh kind:

```
src/commands/tap/add.ts(175,8): error TS1360: Type '{ readonly outcome: "added"; } | { readonly outcome: "relocated"; }'
  does not satisfy the expected type '{ readonly outcome: "added"; }'.
src/commands/tap/render/add.ts(73,11): error TS1360: Type '"added" | "relocated"'
  does not satisfy the expected type '"added"'.
src/commands/tap/render/index.ts(171,5): error TS1360: ... does not satisfy ... { kind: "failed" }
src/commands/tap/render/index.ts(181,5): error TS1360: ... does not satisfy ... { kind: "failed" }
src/commands/tap/render/index.ts(188,5): error TS1360: ... does not satisfy ... { kind: "failed" }
src/commands/tap/render/index.ts(189,22): error TS2339: Property 'error' does not exist on type ...
```

Six errors across all three sites, and the last one shows the required-field half working too. Both variants removed afterwards.

### One change beyond the findings

The exhaustiveness comments pushed `render.ts` to 203 lines, over the hard cap. The add renderer moved to `src/commands/tap/render/add.ts` with the rest in `render/index.ts`, following the directory-over-filename-prefix convention — `add.ts` was the only importer of `renderTapAdd`, so the seam is clean. Now 72 and 143 lines.

### Merge interaction

#130 reworks tap removal into `src/commands/tap/remove/` and #132 makes clone deletion reference-counted. Neither touches `renderTapAdd` or `TapRefreshRow`, so the discriminated union and the `TapAddInput` object should merge cleanly. The one thing to watch: both stacked PRs import from `./render.ts`, which is now `./render/index.ts` — a path update, not a content conflict.

### Verification

Targeted suites only, given the known non-determinism on branches without #131:

- `tests/e2e/tap-dry-run/` — 14 pass, 0 fail
- `tap.test.ts` + `tap-update.test.ts` — 49 pass, 0 fail
- `update-fetch-scope` + `update-edges` + `tap-install` (the other `TapRefreshRow` consumers) — 23 pass, 0 fail
- `render/index.ts` and `render/add.ts` at 100% line and function coverage

I also compared coverage on `add.ts` and `refresh.ts` against the pre-change baseline over the same subset and confirmed the uncovered regions are identical — they are the promotion path and clone-failure branch, which live in suites outside this subset, not a regression from this change.

Typecheck and Biome clean. The only loops I touched are the `rows.filter(...)` counts in `renderTapUpdate`, single bounded passes over a fixed array; I re-read them before running anything, and no test run exceeded 2 seconds.

PR #120, opened 2026-09-11
[LOG-5764] feat(update): accept tap and namespace selectors

stevekrenzel:
**Stacked on #109** (`fix/update-dry-run`). Merge that first; GitHub will retarget this to `main`.

## What

`crew update <selector>` now accepts collections, mirroring `crew install <tap>`:

- **Installed skill name** (bare or `tap/skill`) — unchanged, and always wins.
- **Tap name** (registered or auto) — every state entry attributed to that tap, all scopes. The tap is fetched and, for whole-tap installs (`tracks_tap`), re-expanded so newly added upstream skills come along. A tap with nothing installed from it is fetched and reported (`No skills installed from tap acme — refreshed it anyway.`), exit 0.
- **Namespace** (`acme/marketing`, or bare `marketing` when exactly one tap has installed entries under it) — only the entries under `skills/marketing/`. A namespace selector also **bounds which additions re-expansion installs**: the group spans the whole tap, but a child added under a different namespace is left alone.
- A bare word that is both a tap and a namespace elsewhere, or a namespace installed from several taps → `ambiguous_reference` listing each `crew update <tap>/<ns>` form.
- Anything else → `unknown_skill`, whose message now reads "isn't an installed skill, a tap, or a namespace".

A tap **named as a selector** whose refresh fails is a hard failure (exit 1), even with nothing installed from it — the run did not do what was asked. Per-tap fetch isolation is otherwise unchanged: an unnamed configured tap that fails stays a warning and exits 0.

Human output gets a header per collection (`Updating tap acme (3 skills)`); `--json` gains `selectors: [{ raw, kind, name }]`.

## Why

`crew update <tap>` previously failed with `unknown_skill` even though `tapsToRefreshFor` had a dead branch for bare tap names. Install can target a whole tap; update couldn't.

## Selector identity

Three rounds of review converged on one invariant: a selector binds to the **state entries** it resolved to, by identity, and a run that re-reads its selection by name or by path can silently change what the user asked for.

Entry identity (§11.1) is `(tap, name, scope, project_root)`. `source.path` is deliberately excluded: re-expansion rewrites a relocated skill's tap-relative path in the same run that reads the selection back, so including the path made an entry fail to match itself and drop out, leaving stale bytes installed. The namespace subject refresh had the same defect by a different route — it re-derived membership from the entry's current namespace, so a skill that merely *moved* namespaces fell out of its own selection.

## Where

- `src/state/collections.ts` — `resolveCollectionSubject(s)`, `entryIdentity`; shared so the sibling uninstall-by-collection PR can reuse it.
- `src/commands/update/` — `plan.ts` (selection + lock acquisition), `locked-plan.ts` (everything inside the locks), `selection.ts`, `dep-closure.ts`, `render.ts`, `index.ts`.
- `src/install/tap-reexpand/` — `index.ts`, `survey.ts` (per-group survey), `additions.ts` (namespace bound), `scan-cache.ts`.
- `PRD.md` §10.1 / §10.1.1, criteria C-UPD-26..37.
- Help (`crew help update` examples), README command table.

## Tests

`tests/e2e/update-collections/` — a directory, not a single file:

- `selectors.test.ts` — tap and namespace selectors, ambiguity, precedence, unknown-selector message, named-vs-unnamed tap refresh failure.
- `collisions.test.ts` — selector-identity regressions across taps and scopes, fetch scoping (§16.6), dependency closure through a collection.
- `reexpand-bounds.test.ts` — namespace-bounded additions, and a relocated skill staying selected.
- `helpers.ts` — real local `file://` taps; nothing contacts the network. The `--json` payload is typed against the production row types rather than `any`.


stevekrenzel:
## Review

Verdict: **not approved — 0 blockers, 1 major, 4 minors**

### Major

- **Collection identity is lost across refresh/re-expansion** (`src/commands/update/index.ts:86-110`, `:128`). The code converts resolved subjects to strings, lets `reexpandTaps` reinterpret member/collection names as tap names, then resolves the raw selectors again against mutated state. A selected skill whose name collides with a tap can refresh and re-expand that unrelated tap. Conversely, a tap selector can change meaning after re-expansion if a newly discovered child has the tap’s name, because skill resolution then wins. Carry structured initial selections through refresh, re-expansion, and final targeting (`selectedTapNames`, selected member identities, namespace scope) instead of re-resolving ambiguous strings. Add both collision regressions.

### Minors

- **User-facing command references still describe name-only selection** (`site/components/sections/Commands.data.tsx:25`, `src/commands/help/content/update.ts:5`, `README.md:168`). Use `<selector>` and document skill, tap, and namespace forms consistently.
- **The whole-tap test does not prove every existing member updates** (`tests/e2e/update-collections.test.ts:85`). Only `alpha` changes and is asserted; change both existing members and verify both installed bytes or JSON rows.
- **Dependency closure for collection members is untested** (`tests/e2e/update-collections.test.ts:85`). Add a selected collection member with an installed dependency and assert the dependency updates and carries the expected `transitively_required_by` attribution.
- **Whole-tap selection can repeatedly acquire and index the same tap** (`src/commands/update/index.ts:89`). Expanded member names are passed to `reexpandTaps`, which rescans per `(tap, scope, project_root)` group and linearly checks the selection array. Cache acquired tap contents per tap for the run and use sets for membership tests.

All eleven review lenses ran against head `b1ecbf9`; CI is green. I did not approve because of the major correctness finding.

stevekrenzel:
Addressed all 5 findings in `56feb1d`. Nothing pushed back on — every finding reproduced.

### Major — collection identity lost across refresh/re-expansion

Reproduced both directions before changing anything:

**(a)** With tap `docs` holding a skill named `widgets`, and a *different* tap itself named `widgets`, running `crew update docs/widgets` produced:

```
selectors:          [{"raw":"docs/widgets","kind":"skill","name":"widgets"}]
tap_reexpand_rows:  [{"name":"two","tap":"widgets","kind":"added"}]
```

Updating a skill installed a new skill from an unrelated tap, purely on a name collision.

**(b)** With tap `acme` whole-tap installed (member `alpha`), after upstream added a child whose declared name is also `acme`:

```
selectors: [{"raw":"acme","kind":"tap","name":"acme"}]
rows:      ["acme"]          <- alpha silently dropped
```

The selector resolved as a tap, then the second `chooseEntries` pass re-resolved the raw string against post-re-expansion state, where rule 1 (installed skill wins) demoted it to a skill.

**Fix.** Structure is now carried end to end instead of being flattened to strings:

- `ReexpandScope` (`src/install/tap-reexpand.ts`) replaces the flat `restrictNames` array with `{ memberNames, tapNames }`, so a skill name can never be matched against tap identity. `null` still means "every group".
- `tapsToRefreshFor` (`src/commands/update/selection.ts`) takes the resolved subjects and adds a tap to the refresh set only for `kind === "tap"`, rather than testing resolved skill names against the tap list.
- `refreshCollectionSubjects` (`src/state/collections.ts`) recomputes membership against the mutated state while preserving each subject's resolved kind and name. `src/commands/update/index.ts` uses it instead of re-resolving `rawNames`.

Namespace membership is re-read within the tap it first narrowed to, so a same-named namespace appearing elsewhere mid-run can't widen the set either.

PRD §10.1 gains a normative **"Selector identity is fixed at resolution"** paragraph, plus criteria **C-UPD-31** and **C-UPD-32**.

### Minors

- **Whole-tap test didn't prove every member updates** (`selectors.test.ts`). Confirmed: only `alpha` moved. Both `alpha` and `beta` now change and are asserted on installed bytes, and the JSON row set is asserted to be exactly `["alpha","beta","gamma"]`.
- **Dependency closure untested for collection members.** Added a test where tap `acme`'s member `app` depends on `lib` installed from a *different* tap; selecting the tap must pull `lib` into the closure. Asserts `lib` updates and carries `transitively_required_by: ["app"]`.
- **Repeated acquisition/indexing per group.** A tap installed at user scope plus three projects forms four groups over one clone, and both `acquireTap` and `currentTapChildren` (a full tree walk via `indexTap`) ran per group. New `src/install/tap-scan.ts` memoizes the scan per tap per run; the config lookup and membership tests are now `Map`/`Set` instead of linear scans. Measured with a counting git runner on that four-group fixture: **13 → 10 `rev-parse` calls**, plus three fewer tree walks. Scales as one scan per tap rather than one per group.
- **User-facing surfaces described name-only selection.** `crew help update`, the README command table, and `site/components/sections/Commands.data.tsx` now all use `<selector>` and spell out the skill / tap / namespace forms, including that a skill name wins over a same-named collection.

### Pre-fix verification

Stashed `src/` only, keeping the new tests: **C-UPD-31 and C-UPD-32 both fail** against the pre-fix commit. The dependency-closure test passes either way — it closes a coverage gap rather than catching a live bug, so I'm not counting it as a regression guard.

### Cross-PR note

This does change the shape of the shared collections module: `src/state/collections.ts` gains `refreshCollectionSubjects`. #126 (`feat/uninstall-collections-v2`) independently created the same module with a trailing command-verb parameter on the resolver. Those are orthogonal — when the branches meet, keep #126's parameterized signature and carry `refreshCollectionSubjects` across alongside it. Uninstall doesn't mutate state mid-run, so it has no need to call the new function, but the collision hazard is structural: if uninstall ever grows a mutate-then-retarget step, it will need the same treatment.

### Layout

`tests/e2e/update-collections.test.ts` became `tests/e2e/update-collections/` (`selectors.test.ts`, `collisions.test.ts`, `helpers.ts`), and `tap-reexpand.ts` shed the scan logic to `tap-scan.ts` — every touched file is under the 200-line cap.

`bun run check`: 997 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean. Site lint clean.

stevekrenzel:
Rebased onto the revised parent (`fix/update-dry-run` @ `6ef2477`), which had itself changed for its own review: it extracted a lock-free `planUpdate`, added `would_update`/`would_add` handling, and moved the additions loop into `src/install/tap-additions.ts`. Branch is now `5611ea6`; base unchanged, PR is MERGEABLE.

**Five conflicts across two commits, all resolved keeping both sides' intent:**

|---|---|---|
| `README.md` | Both rewrote the `crew update` row | Kept my selector prose **and** the parent's corrected dry-run wording |
| `src/commands/update/index.ts` (×2) | Parent hoisted the body into `planUpdate`; mine still had the logic inside `withStateLock` | Kept the parent's lock-free plan/apply split, re-applied my collection work *inside* `planUpdate` |
| `site/…/Commands.data.tsx` | Both rewrote the update description | Same as README — selectors plus accurate dry-run wording |
| `src/install/tap-reexpand.ts` (×2) | Parent extracted `collectAdditions`; mine had the old inline loop | Kept the parent's extracted call (it also **validates** new children, which my inline loop did not) and re-applied my scan memoization at its new home |

**One wording fix beyond mechanical resolution.** Both my commits described `--dry-run` as previewing "without writing anything" — the exact claim the parent's review corrected, since tap clones *are* refreshed. The README and site copy now state the real boundary. I also swept for that phrasing elsewhere; the only other hits are `crew install`'s help, where it is accurate.

**No criteria renumbering needed.** The parent's C-UPD rows stop at 25 and use the `18a/18b/18c` suffixes; mine are 26–32. No collision.

**Verified both features survived, not just that the suite is green.** A clean rebase proves nothing on its own, so I confirmed each of the four behaviours this branch added is still wired in (`reexpandScopeFor`, subject-keyed `tapsToRefreshFor`, `refreshCollectionSubjects` on re-read, `makeScanCache`) and that the parent's new-child validation is still reached via `collectAdditions`. Targeted runs: parent's dry-run suite 5/5, this branch's collision regressions 3/3, selectors 7/7.

`bun run check`: 1000 pass, 0 fail, 100% line and function coverage.

stevekrenzel:
## Updated /review

**Verdict:** 0 blockers, 1 major, 3 minors. Not approved.

### Major

- `src/state/collections.ts:97`, `src/commands/update/selection.ts:71-77`, `src/install/tap-reexpand.ts:117-120` — Skill selectors collapse to a bare name during refresh and re-expansion, losing the resolved tap and scope identity. For example, `crew update a/foo` can re-expand and update an unrelated `b/foo`, including one installed in another project scope. Preserve a stable state-entry identity through selection refresh and re-expansion, and cover the cross-tap/cross-scope collision.

### Minors

- `src/commands/update/selection.ts:115,132,141` — Collection dependency expansion repeatedly scans all installations and uses `queue.shift()`, making large collection updates quadratic. Build the reverse dependency index once and traverse with a cursor or stack.
- `tests/e2e/update-collections/` — The new per-tap scan cache is not exercised with the same whole tap installed at both user and project scopes. Add a multi-scope test that introduces a sibling and verifies every group receives it from the cached scan.
- `src/install/tap-reexpand.ts:65,76`, `src/commands/update/selection.ts:66` — `ReexpandScope`/`scope` describes a selection filter while `Scope` already means user/project installation scope in this flow. A name such as `ReexpandSelection` would make the distinction clear.

stevekrenzel:
Rebased onto the revised parent (`c985aeb`) and addressed all four findings. Branch is now `603e545`; PR is MERGEABLE and the base is unchanged. Nothing declined.

## Rebase

Two commits, five conflicts. The parent had moved planning into `update/plan.ts` with a lock-wrapped split (`planUpdate` → `withTapLocks` → `planLockedUpdate`), so the collection work folded into that structure rather than the old inline `planUpdate` in `index.ts`. The parent's second commit also superseded my own round-one fix wholesale, so those hunks resolved to the parent's side. One stale file, `src/install/tap-scan.ts`, was dropped — the parent's `tap-reexpand/scan-cache.ts` replaces it and nothing imported the old path.

## Major — selector identity

Reproduced before fixing. A selector that resolved to a single entry came back holding three after re-expansion:

```
resolved:      [ a/foo@user ]
after refresh: [ a/foo@user, b/foo@user, b/foo@project ]
```

`refreshCollectionSubjects` matched skill selectors by name, so any same-named entry appearing mid-run joined the selection. Two more places had the same weakness: `orderedEntries` deduplicated on a key that omitted the tap, and `reexpandTaps` decided which `(tap, scope, project_root)` groups to touch by testing member *names* against a flat list.

All three now key on a shared `entryIdentity` — tap, tap-relative path, name, scope, `project_root` (§11.1). `ReexpandScope` became `ReexpandSelection` carrying identities rather than names, which also resolves the third minor: the old name read as if it meant installation scope, which `Scope` already means in this flow.

Namespace membership additionally pins the scopes it resolved against, so a project install appearing mid-run can't widen a user-scope namespace selection.

**Pre-fix status, stated precisely.** `C-UPD-33 a project-scope install from another tap is not re-expanded` fails against the pre-fix sources and passes after. The cross-tap sibling test passes either way — the tap-vs-skill separation from round one already covered that direction — so it guards the behaviour rather than catching a live regression, and I'd rather say so than let two tests imply two caught bugs.

**One correction to my own work.** My first attempt at the scope regression asserted that `crew update solo` should touch only the user-scope install. That was wrong: `resolveStateSubject` deliberately spans scopes, so both entries are selected at resolution and updating both is correct. I verified that directly before rewriting the test to target what actually breaks — entries that appear *after* resolution.

## Minors

**Quadratic closure.** `dependencyClosureFor` rescanned every installation per dequeue and used `shift()`. It now builds the reverse dependency index once and iterates the queue in place. Measured on a 50-skill dependency chain: **2500 entry visits → 99**. The loop terminates because a node enqueues its dependents only on first visit, so each of the finitely many names expands once; I re-read it before running anything against it, per the resource directive.

**Multi-scope scan cache.** Added `C-UPD-34`: one tap installed as a whole at both user and project scope must re-expand both groups when a sibling appears upstream, asserting the sibling lands in each.

## Spec

§10.1's "selector identity is fixed at resolution" section gains the identity rule as normative text, plus criteria C-UPD-33 and C-UPD-34.

## Merge shape for `src/state/collections.ts`

This branch now adds **`entryIdentity`** (exported) alongside `refreshCollectionSubjects` and the command-verb parameter. #126 independently adds a `collectionState` parameter. The merged module must keep all four; they are orthogonal. `entryIdentity` is worth hoisting as the shared definition — #110's round two fixed a closely related bug by keying state helpers on `(name, scope, project_root)`, and this is the same idea with the tap included.

## Verification

Targeted suites only, given the known non-determinism on branches without #131: update-collections + tap-reexpand 15 pass, update-dry-run + dep-closure + fetch-scope 17 pass, update-edges + tap-update 12 pass, all 0 fail. `selection.ts` and `collections.ts` at 100% line and function coverage; the remaining gaps in `plan.ts` and `tap-reexpand/index.ts` are identical to the unmodified parent's, which I measured directly rather than assumed. Typecheck and Biome clean.

stevekrenzel:

**Verdict:** 0 blockers, 4 majors, 5 minors, 1 nit. **Not approved.**

Reviewed `c985aeb..603e545` through all 11 `/review` lenses. This stacked PR currently has no `check` run because CI only triggers for PRs based on `main`.

### Majors

1. A namespace selector can install newly discovered skills from other namespaces during whole-tap re-expansion (`src/commands/update/plan.ts:148`). Restrict additions to the selected namespace or document broader behavior.
2. A targeted skill relocated upstream can update state to its new path, then disappear from the post-re-expansion selection, leaving installed bytes stale until a second run (`src/state/collections.ts:131`). Carry relocation identity into selection.
3. A failed refresh of a selected empty tap returns exit 0 (`src/commands/update/plan.ts:141`). Include failed tap rows in `hardFailure` and cover the C-UPD-29 failure path.
4. `sourceGone` is keyed only by skill name (`src/install/tap-reexpand/index.ts:66`), so deletion of `shared` in one selected tap can incorrectly skip another tap's still-present `shared`. Key by full entry identity.

### Minors

1. `src/commands/update/plan.ts` is 212 lines and `src/install/tap-reexpand/index.ts` is 203, exceeding the repository's hard `<200` cap.
2. Update-plan comments/docstrings still describe name-only selection and incorrectly say new children are excluded (`src/commands/update/index.ts:2`, `plan.ts:162`).
3. C-UPD-31 does not assert that the unrelated tap was not fetched (`tests/e2e/update-collections/collisions.test.ts:22`).
4. C-UPD-28 covers a same-named tap but not a same-named namespace (`tests/e2e/update-collections/selectors.test.ts:125`).
5. `RunResult.json(): any` disables schema checking for the new JSON assertions (`tests/e2e/update-collections/helpers.ts:27`).

### Nit

1. The PR description names a nonexistent combined test file and stale test count.

stevekrenzel:
## Round three addressed

All 4 majors, 5 minors, and the nit are handled. Every major was reproduced first; two of them turned out to need a different fix than the one suggested, with evidence below.

### Majors

**1. Namespace selector installs additions from other namespaces — fixed.**
Reproduced: a tap with namespaces `marketing` and `eng`, both whole-tap installed, gains one child in each upstream. `crew update acme/marketing` installed `eng/fmt` as well. `ReexpandSelection` carried no namespace bound, so `collectAdditions` took every new child in the group. It now carries `namespaces`, and `additions.ts` skips a child whose `skills/<ns>/<name>` namespace is not among the selected ones. Naming the tap, or mixing in a skill selector, leaves additions unbounded — a skill selector says nothing about which namespaces are in scope. Covered by C-UPD-36 in `reexpand-bounds.test.ts`, both directions.

**2. Relocated skill drops out of the selection — fixed, and it was worse than reported.**
Root cause was `entryIdentity` including `source.path`. Re-expansion rewrites that path mid-run, so the entry failed to match itself and fell out of the post-re-expansion selection. `source.path` is now excluded; identity is `(tap, name, scope, project_root)`.

Worth flagging: that alone was **not** sufficient. `refreshCollectionSubjects` re-derived namespace membership from each entry's *current* namespace, so a skill that moved from `alpha` to `beta` fell out of an `acme/alpha` selection by a second, independent route. Namespace subjects now carry their resolved entries by identity as well. I confirmed both are load-bearing by reverting each in isolation — each revert alone fails C-UPD-37.

One caution for the next reviewer, because it nearly cost me a wrong fix: my first repro relocated a skill from a flat tap root into an arbitrary `nested/` subdirectory. That is genuinely outside the indexed layout (the index walks one level deep, or `skills/<ns>/<name>`), so `source_gone` there is **correct**, not a defect. The bug only reproduces for a relocation that stays inside the indexed layout.

**3. Failed refresh of a selected empty tap exits 0 — fixed, but narrower than suggested.**
Pushing back on the suggested form. "Include failed tap rows in `hardFailure`" unconditionally **regresses C-UPD-20**, which is normative: §10.1 says per-tap fetch failures warn and do not abort. I implemented it as suggested first and watched `tests/e2e/tap-install.test.ts` go to `15 pass, 1 fail` — a bare `crew update` with one broken tap started exiting 1.

The defensible rule is narrower: a failure is hard only for a tap the user **named as a tap selector**, where the run genuinely did not do what was asked. An unnamed configured tap that fails stays a warning at exit 0. Both directions are covered by C-UPD-35, and `tap-install.test.ts` is green again.

**4. `sourceGone` keyed by bare name — fixed.**
Now keyed by full entry identity. Note this does not reproduce across two taps at the same scope: the name-conflict guard blocks installing the same name twice there. It reproduces **across scopes** — `shared` at user scope from tap `a`, and at project scope from tap `b`.

### Minors

1. **File cap.** Three files crossed 200 while absorbing these fixes. Extracted `src/commands/update/locked-plan.ts` (the work inside the clone locks), `src/commands/update/dep-closure.ts`, and `src/install/tap-reexpand/survey.ts` (per-group survey). Every touched file is now under the cap; the largest is `tap-reexpand/index.ts` at 195.
2. **Stale docstrings.** `commands/update/index.ts` now documents the selector set as skill/tap/namespace and states the namespace bound on additions.
3. **C-UPD-31 fetch assertion.** Added: the unrelated tap must not appear in `tap_rows` at all, per the §16.6 fetch scoping — not re-expanded is not the same as not fetched.
4. **C-UPD-28 same-named namespace.** Added, using a namespaced tap holding `pdf/reader` against a flat tap holding a skill literally named `pdf`.
5. **`json(): any`.** Replaced with a typed `UpdateJson` derived from the production row types. This immediately earned its keep: it surfaced five unguarded `.error` accesses in the selector tests that `any` had been hiding.

### Nit

PR description rewritten. It named a nonexistent combined `tests/e2e/update-collections.test.ts` with a stale count; the real layout is a directory of three suites plus helpers.

### PRD

Three of these are externally observable, so PRD first per CLAUDE.md. §10.1 gains the namespace-bound rule and the named-tap-refresh-failure rule, and drops the tap-relative path from selector identity (that text previously said identity included it, which the fix for major 2 contradicts). New criteria C-UPD-35, C-UPD-36, C-UPD-37.

### Verification

This PR is stacked, so it gets no `check` run — targeted verification is the only evidence, and I ran it deliberately rather than running the full suite.

```
bun run typecheck                                    clean
bun run lint                                         clean
bun test tests/e2e/update-collections/               19 pass, 0 fail
bun test tests/e2e/update-collections/ \
         tests/e2e/tap-install.test.ts \
         tests/e2e/update.test.ts                    35 pass, 0 fail
```

I re-read the re-expansion loop in `tap-reexpand/index.ts` before running anything that drives it, and confirmed it terminates: it iterates a finite group list built once from state, and the inner walk iterates a finite child set from a single tap scan. No run exceeded a few seconds.

One inherited-behavior note: this branch does **not** carry a copy of the `isSuppressed` dry-run fix. It inherits the central one from #107 at merge, which is the right place for it.

stevekrenzel:
Rebased onto the new #109 tip; no conflicts. Update tests pass (38).

stevekrenzel:
Rebased onto the new #109 tip (`2b578c5`). One modify/delete conflict on the deleted `Commands.data.tsx`: the `update [<selector>…]` signature and selector description were re-placed into `Commands.data/managing.tsx`.

One real fix, folded into the second commit: the full suite showed `tests/unit/tap-reexpand.test.ts` › "empty tap roots are treated as source_gone" failing — `sourceGone` was switched to entry identities (`entryIdentity(m)`) in this PR, but that unit test still expected the bare name. It failed on the pre-rebase tip too (verified), unseen because stacked PRs get no CI and only the e2e update suites had been run. The assertion now expects `entryIdentity(trackedEntry())`. `bun run check`: typecheck, lint (incl. the file-size gate), full suite 1125 pass at 100%.

PR #121, opened 2026-09-11
[LOG-5765] feat(refs): accept scheme-less hosts and hint @owner/repo on misses

stevekrenzel:
Stacked on #113 (`feat/browser-url-refs`). Merge that first; GitHub will retarget this PR to `main` automatically.

## What

Three ergonomics for references that look like a repository but aren't spelled as one. All three were confirmed by running the CLI against a scratch home.

1. **Scheme-less hosts.** `github.com/anthropics/skills`, `www.github.com/o/r`, `gitlab.com/o/r`, `git.example.com:8443/o/r`, and `github.com/o/r/tree/main/path` used to fail with `invalid_ref` ("names may only contain letters, digits, and hyphens"). Now, when the first `/`-segment contains a `.` (tap names can't) and there are at least three segments, `https://` is prepended and the argument goes through the existing URL and browser-URL parsing. Works everywhere `parseRef` is used (install, info, tap add, dependency lists).
2. **`owner/repo` misses hint the `@` form.** `crew install anthropics/skills` and `crew info anthropics/skills` now say "If you meant the GitHub repository anthropics/skills, use `@anthropics/skills` instead." When the bundled known-tap registry has a tap whose URL is `https://github.com/<owner>/<repo>`, the install error also names that tap ("This is the tap for the GitHub repo …") with its `crew tap add` command, and JSON `known_tap_suggestions` includes it. The known-tap wrapper now preserves the resolver's remedy rather than suppressing it.
3. **`crew tap add owner/repo`** keeps its `usage_error` but the remedy suggests `crew tap add @owner/repo`.

## PRD

- §8.2: new `<host>/<owner>/<repo>` accepted form and "Scheme-less hosts" paragraph.
- §8.5: new precedence rule 4 (rules renumbered) and a "Two-segment misses" paragraph.
- §9 step 2: known-tap URL match for two-segment refs.
- §13: "Guess the spelling" guideline.
- §16.3: tap-add remedy.
- §18.3: C-REF-28, C-REF-29 (after C-REF-27); C-TAP-24b, C-TAP-24c (after C-TAP-24).

## Tests

- `tests/refs/parse.test.ts`: table of scheme-less forms (github.com, www., gitlab.com, self-hosted with port, `@ref`, `//subpath`, `/tree/` path) plus negatives (`core/python-testing` stays a tap ref; `a.b`, `a.b/c`, `.com/a/b`, `com./a/b` are rejected).
- `tests/e2e/owner-repo-hints.test.ts`: install/info/tap-add hints through the real CLI, with a fixture registry entry at `https://github.com/anthropics/skills.git`; JSON shape; no-known-tap case keeps only the `@` hint.

`bun run check` green: 1037 tests, 100% line and function coverage. No network in tests.

## Judgment calls

- Only `crew install` consults the known-tap registry, as today; `crew info owner/repo` gets the `@` hint but not the tap-add line.
- The `@` hint is phrased command-neutrally ("use `@owner/repo` instead") because the resolver doesn't know which command called it.
- README reference table and the install help page mention the scheme-less form; the site's SkillRefs section already lists enough forms that one more felt like noise, so it was left alone.


stevekrenzel:
## Review

Verdict: **not approved — 0 blockers, 2 majors, 2 minors**

### Majors

- **Scheme-less host detection accepts userinfo and can clone from a different host** (`src/refs/parsers.ts:34`). An input such as `github.com@evil.example/acme/skills` passes the dot check; after `https://` is prepended, URL parsing treats `github.com` as userinfo and the actual host as `evil.example`. Require the first segment to be a valid hostname with an optional numeric port, and reject userinfo, query/fragment delimiters, and malformed authority syntax. Add rejection tests for both `github.com@evil.example/acme/skills` and `github.com:443@evil.example/acme/skills`.
- **The two-segment remedy drops the requested ref** (`src/install/resolve-ref/errors.ts:21`). `owner/repo@v1` is parsed with `ref: "v1"`, but the new hint suggests `@owner/repo`, which may install a different revision. Pass the parsed ref/source into the error helper, preserve `@v1` in the suggested GitHub reference, and cover both install and info.

### Minors

- **Malformed ports escape as the wrong error path** (`src/refs/parsers.ts:39`). A value such as `github.com:abc/o/r` is accepted as a scheme-less host, then `new URL()` throws natively, yielding an unexpected `usage_error` rather than `invalid_ref`. Validate authority syntax before dispatch or convert URL-construction failure to a normal parse miss; add a regression test.
- **The no-`.git` known-tap URL form is untested** (`tests/e2e/owner-repo-hints.test.ts:17`). C-TAP-24b requires matching repository URLs with and without the suffix, while the fixture covers only `.git`. Add/parameterize a suffixless fixture and assert identical human and JSON suggestions.

All eleven review lenses ran against head `8f867fb`; CI is green. I did not approve because of the major findings.

stevekrenzel:
Addressed all four findings in `a209797`. Nothing pushed back on — every one reproduced.

## Major 1 — scheme-less host confusion (spoofing vector)

**Reproduced first.** `parseRef("github.com@evil.example/acme/skills")` returned a git source with `url: https://evil.example/acme/skills`, and `crew install` on it reached an actual clone attempt against `evil.example`. The userinfo is dropped during canonicalization, so the recorded URL keeps no trace of the `github.com` the user saw — nothing downstream would reveal the substitution. `github.com:443@evil.example/...` behaved identically. This was a spoofing vector: a reference that reads as GitHub clones from an attacker-controlled host.

**Fix** (`src/refs/parsers.ts`): the first segment must now match a bare authority — dot-separated alphanumeric/hyphen DNS labels with an optional **all-digit** port.

Rejected (falls through to tap parsing → `invalid_ref`, exit 4): userinfo (`@`), query (`?`), fragment (`#`), non-numeric or empty ports (`:abc`, `:`), leading/trailing-hyphen labels, leading/trailing dots.

Still accepted: `github.com/o/r`, `www.github.com/o/r`, `gitlab.com/o/r`, `git.example.com:8443/o/r`, `my-host.example.co.uk/o/r`, and `/tree/` browser paths.

Explicit URLs are matched earlier in the §8.5 precedence chain, so authenticated `https://user:pw@host/...` clone URLs are untouched by this guard. I also checked the browser-URL normalizer this PR stacks on: it only handles inputs that already carry a scheme and rebuilds `origin` from the parsed `hostname`, so the same confusion isn't reachable there. The scheme-less path was the only place a scheme is prepended to user input.

## Major 2 — the remedy dropped the requested ref

**Reproduced:** `crew install anthropics/skills@v1` echoed the reference as `anthropics/skills` and suggested `@anthropics/skills`, which would install the default branch rather than `v1`.

**Fix:** `twoSegmentMissError` takes the parsed ref (`src/install/resolve-ref/errors.ts`, passed from `index.ts`), so the echoed reference and the `@owner/repo@v1` suggestion both preserve it. The known-tap repo-match suggestion in `match.ts` preserves it too — §8.4 allows `tap-name @ tap-ref` for whole-tap installs, so `crew install anthropic@v1` is valid. Covered for both `install` and `info`, human and JSON.

One rendering detail: the suggestion heading previously reused `installRef` as its label, which would have read `anthropic@v1 (official)` as though the ref were part of the tap's name. The heading now names the tap and the ref appears only in the command below it.

## Minor — malformed ports escaped as the wrong error

**Reproduced:** `crew install github.com:abc/o/r` produced `usage_error` with *"crew hit an unexpected error: … cannot be parsed as a URL — please report this at …/issues"*, asking the user to file a bug for their own typo. The Major 1 guard fixes this by construction: the input never reaches `new URL`, so it now returns `invalid_ref` with exit 4 and the normal ref-shape remedy.

## Minor — suffixless known-tap URL untested

`tests/e2e/owner-repo-hints.test.ts` now parameterizes the registry fixture over the `.git` suffix and asserts a registry entry at `https://github.com/anthropics/skills` (no suffix) produces identical human and JSON suggestions.

## Pre-fix verification

Stashed `src/` only and kept the new tests: **10 of 11 fail against the pre-fix commit** — all three spoofing cases, four malformed-authority cases, and all three ref-preservation cases. The suffixless known-tap test passes either way; it closes a coverage gap rather than catching a live bug, and I'd rather say so than let the count imply more.

## Notes

- Tests live in a new `tests/refs/schemeless-authority.test.ts` rather than extending `tests/refs/parse.test.ts`, which is already 322 lines — over the hard cap on `main`, pre-existing and out of scope here.
- One assertion I wrote was wrong and I corrected it rather than the code: I first asserted the error message must not contain `evil.example`, but the message correctly echoes the user's input verbatim, trailing host included. The property that matters is that no git source is produced, which the test now asserts directly.
- I did not assert that `https://user:pw@host/...` survives canonicalization with its credentials, because on this base it does not — that's #113's Major 1, being fixed on the parent branch. Asserting current behavior would have encoded the defect; the test pins only that explicit URLs still route to the git parser.

`bun run check`: 1052 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean. All touched files under the 200-line cap.

stevekrenzel:
Rebased onto the revised parent (`618a40d`), which fixed its own review findings. Branch is now at `0f770c9` and MERGEABLE.

## Conflicts and resolutions

**Three conflicts across the two commits, in two files.**

|---|---|---|
| `PRD.md` (commit 1) | Both sides claimed `C-REF-28`/`C-REF-29` | Kept the parent's 28/29/30 (redaction, userinfo preservation, query/fragment stripping); renumbered mine to **C-REF-31/32** |
| `src/commands/help/content/install.ts` | Both rewrote the same two `notes` entries | Merged: my scheme-less host syntax and the `@owner/repo` hint, the parent's fuller slash-branch example, and the parent's new private-repo note — all three kept |
| `PRD.md` (commit 2) | My spoofing-guard row was `C-REF-29a`, now taken | Renumbered to **C-REF-32a** |

Test titles referencing the old IDs were updated to match (`tests/refs/parse.test.ts`, `tests/refs/schemeless-authority.test.ts`), so a conformance failure still names the right criterion.

## The spoofing guard still holds

This was the thing most at risk: my guard rejects userinfo in scheme-less refs, while the parent independently changed userinfo handling to preserve it for acquisition and redact it for display. Those could plausibly have collided, so I verified rather than assumed — capturing parser behaviour before the rebase and re-running the same probe after.

Both attack strings are still rejected:

```
REJECT github.com@evil.example/acme/skills
REJECT github.com:443@evil.example/acme/skills
```

Legitimate forms still parse (`github.com/acme/skills`, `www.github.com/...`, `git.example.com:8443/...`, `github.com/o/r/tree/main/py`).

The two changes compose as intended rather than fighting. Pre-rebase, an authenticated URL lost its credentials during canonicalization; post-rebase it keeps them:

```
ACCEPT https://user:s3cret@github.com/acme/skills.git
   url=https://user:s3cret@github.com/acme/skills.git
```

…while the display path still masks them. A rejected blob link carrying a token returns `invalid_ref` (exit 4) with the secret replaced by `***` and no leak in the `--json` payload. So: credentials survive for cloning, are hidden in output, and host spoofing stays blocked.

## One change beyond conflict resolution

`tests/refs/parse.test.ts` was already **286 lines on `main`**, over the hard 200-line cap, and my scheme-less table would have pushed it to 322. Rather than deepen a pre-existing violation in a file I was touching anyway, those cases moved to `tests/refs/schemeless-hosts.test.ts` (47 lines). `parse.test.ts` is now byte-identical to its state on the parent branch. The file remains over cap — that predates this stack and deserves its own split.

## Verification

- Parser safety suites (both sides): `schemeless-authority`, `browser-url-safety`, `browser-url` — **60 pass**.
- End-to-end and parse suites (both sides): `owner-repo-hints`, `browser-urls`, `parse` — **72 pass**.
- `bun run check`: exit 0, **1082 pass, 0 fail**, 100% line and function coverage, typecheck and Biome clean.
- `git merge-base --is-ancestor origin/feat/browser-url-refs HEAD` holds, so the stack is intact.

stevekrenzel:
## Updated /review

**Verdict:** 0 blockers, 1 major, 1 minor, 1 nit. Not approved.

### Major

- `PRD.md:751` — The formal reference grammar still omits the new scheme-less `<host>/<owner>/<repo>` git URL form while §8.2 requires it. The PRD is this project’s implementation contract, so a conforming implementation could reject the newly supported syntax. Add the form to §8.4’s grammar.

### Minor

- `site/components/sections/SkillRefs.tsx:29` — The site’s dedicated Git-source reference card omits scheme-less host references, although the README and CLI help document them. Add an example such as `github.com/acme/skills`.

### Nit

- `tests/refs/schemeless-authority.test.ts:80` — The comment says userinfo is stripped and cites PR #113, but the parser preserves it. Replace it with the actual reason this test asserts routing only.

stevekrenzel:
Rebased onto the revised parent (`3123e28`) and addressed all three findings in `37571f4`. Nothing declined.

## Rebase

One conflict this time, down from three. `src/commands/tap/target.ts` — the parent added `displayText`/`displayUrl` imports for credential redaction while this branch added the `TapSource` type for the `owner/repo` remedy. Both kept; the rest of the file merged cleanly, so the parent's redaction of `raw` in the pinned-tap error and this branch's `ownerRepoRemedy` both survive intact. `git merge-base --is-ancestor origin/feat/browser-url-refs HEAD` holds.

## Major — grammar (`PRD.md` §8.4)

Confirmed the gap: §8.2 line 603 lists the accepted form and §8.5 rule 4 gives it precedence, but the grammar's `git-url` production had only `https://`, `git@`, shorthand hosts, and `@owner/repo`. A second implementation building from §8.4 would have rejected `github.com/acme/skills`.

Added the production and a `bare-authority` rule taken from the code's actual regex rather than my reading of it — dot-separated DNS labels with an optional numeric port:

```
git-url     := … | bare-authority "/" owner "/" repo        (scheme-less, §8.2)
bare-authority := dns-label 1*("." dns-label) [ ":" 1*DIGIT ]
dns-label   := [a-z0-9] [ [a-z0-9-]* [a-z0-9] ]         (case-insensitive)
```

The grammar alone could still read as permissive, so it's followed by a normative note: an implementation MUST NOT widen this to "everything before the first `/`". The scheme-less form resolves by prepending `https://`, so userinfo, query, fragment, and non-numeric ports have to be rejected *before* a URL parser sees them — otherwise `github.com@evil.example/o/r` parses as host `evil.example` with `github.com` demoted to a username, and a reference a human reads as GitHub clones from elsewhere. The note points at C-REF-32a, which already pins that behaviour.

## Minor — site card (`site/components/sections/SkillRefs.tsx`)

The Git-source card listed four examples and no scheme-less one, though the README table and `crew help install` both document it. Added `crew install github.com/acme/skills` and a clause noting you can drop the `https://`. My round-one note called this "noise"; with README and help both carrying it, the site was the odd one out.

## Nit — test comment (`tests/refs/schemeless-authority.test.ts`)

The comment was stale in two ways: it cited a PR number, and it said userinfo "is currently stripped". I checked rather than assuming — `parseRef("https://user:pw@github.com/acme/skills")` now returns that URL verbatim, because the parent's round-one fix preserves credentials so private repos clone. Since the stated reason for asserting routing-only no longer applies, the test now asserts the full URL, with a comment explaining the actual distinction: the guard rejects userinfo in *scheme-less* refs because of the spoofing vector, while an explicit URL is matched earlier by §8.5, was written with its scheme by the user, and must keep its credentials per §8.2.

## Spoofing guard re-verified post-rebase

The parent changed credential handling, so I re-ran the probe rather than assuming the guard survived:

```
REJECT github.com@evil.example/acme/skills
REJECT github.com:443@evil.example/acme/skills
ACCEPT github.com/acme/skills              -> https://github.com/acme/skills
ACCEPT www.github.com/acme/skills          -> https://github.com/acme/skills
ACCEPT git.example.com:8443/acme/skills    -> https://git.example.com:8443/acme/skills
ACCEPT github.com/o/r/tree/main/py         -> https://github.com/o/r.git ref=main sub=py
ACCEPT https://user:s3cret@github.com/acme/skills.git
       url=https://user:s3cret@github.com/acme/skills.git
```

Both attacks still blocked, legitimate scheme-less forms still route to the real host, and an authenticated URL still round-trips with credentials intact for cloning.

## Verification

Targeted suites only, per the known non-determinism on branches without #131: `schemeless-authority` + `schemeless-hosts` 20 pass; `owner-repo-hints` + `browser-url-safety` + `display-url` 44 pass; 0 fail. Typecheck, Biome, and site lint clean. No loops changed — the three edits are grammar text, a string array, and a test assertion.

stevekrenzel:

**Verdict:** 0 blockers, 2 majors, 6 minors, 1 nit. **Not approved.**

Reviewed `3123e281..37571f4` through all 11 `/review` lenses. This stacked PR has no `check` run because CI only triggers for PRs based on `main`.

### Majors

1. **Suggested install commands can execute shell syntax from an unvalidated ref** — `src/commands/install/known-tap-fallback/match.ts:58`
   Refs reject whitespace but can contain `$()` and are interpolated into copy-pasteable `crew install ...` suggestions. A dependency or quoted input can therefore produce a command that executes substitution when pasted. Shell-quote the complete positional argument and terminal-sanitize it at every human/JSON suggested-command surface.
2. **Malformed scheme-less refs can leak query credentials** — `src/refs/parse.ts:57`
   A value such as `github.com?token=secret/acme/skills` fails authority recognition, falls into tap parsing, and is echoed raw in message/details. Redact this failure path centrally and add human/JSON coverage.

### Minors

1. Repository-based known-tap suggestions are suppressed when the same tap also has a matching skill/namespace (`src/commands/install/known-tap-fallback/match.ts:58`). Include both interpretations.
2. `github.com:99999/o/r` throws native `TypeError` rather than `invalid_ref` (`src/refs/parsers.ts:38`). Translate URL construction failures.
3. `tap add foo/bar@v1` omits the PRD-required owner/repo correction with preserved ref (`src/commands/tap/target.ts:50`).
4. PRD precedence rule 4 and the scheme-less test docstring overstate detection as merely “contains a dot”; reference the full `bare-authority` predicate (`PRD.md:798`).
5. Move scheme-less Git authority recognition from the path/tap parser module into `src/refs/git-url.ts` (`src/refs/parsers.ts:27`).
6. Add focused tests for out-of-range ports, ref-preserving `tap add`, and invalid scheme-less query redaction.

### Nit

1. The PR description names the old test location and says `SkillRefs.tsx` was unchanged although this diff changes it.

stevekrenzel:
Round-three feedback addressed in 513b8d8. Details per finding.

## Major 1 — shell metacharacters in the suggested `crew install` command

**Reproduced.** This contradicts the earlier note on this PR that it was not reproducible, and I want to be explicit about why that note was wrong rather than quietly reversing it.

The earlier non-reproduction tested **tap names**, which do reject metacharacters. But the suggestion interpolates a full reference, and a reference can carry an `@<ref>` tail. §8.4 constrains a git-ref only to "no whitespace and no slash" — `$(…)`, backticks, `;`, `|`, `&&`, and quotes are all legal there and all survive parsing:

```
parseRef("pdf@$(id)")  →  { ref: "$(id)", … }
```

End-to-end, `crew install pdf@$(id)` against a home with a known tap printed the suggestion verbatim on both surfaces — human output and `known_tap_suggestions[].install` in `--json`:

```
crew install anthropic/pdf@$(id)
```

I confirmed the consequence in a real shell rather than reasoning about it. Pasting the unquoted suggestion executed the substitution and printed my actual group memberships. Pasting the fixed form passes `anthropic/pdf@$(id)` through as one literal argument.

**Fix.** Added `shellQuote` in the new `src/util/shell.ts`, applied at every site that prints a command for the user to run:

- `src/commands/install/known-tap-fallback/index.ts` — one `installCommand` helper now feeds both the human renderer and `suggestionJson`, so the two surfaces cannot drift apart again.
- `src/commands/tap/target.ts` — the `owner/repo` remedy.

Quoting is conditional: anything matching the unreserved POSIX set is returned unchanged, so ordinary suggestions like `anthropic/pdf@v1.2.0` stay readable. A literal single quote round-trips via close-escape-reopen, the only POSIX construct available inside single quotes.

It landed in its own file because adding it to `src/util/format.ts` pushed that file to 201 lines, one over the hard cap.

**Verified load-bearing.** Reverting the two call sites fails 3 of the 4 new tests in `tests/e2e/suggested-command-safety.test.ts`. The fourth asserts ordinary refs are left unquoted and correctly passes either way.

## Major 2 — credentials in a malformed scheme-less ref

**Already fixed**, by the error-boundary redaction this branch is stacked on. `github.com?token=SECRET/acme/skills` renders `***` on both the human and JSON surface today, so no new code was needed here.

**What I did not do:** I initially added a regression test for it on this branch and then removed it. That test only passes with the parent's redaction commit, which this branch does not itself contain — committing it here would have left the branch red until merge. The parent branch already carries its own coverage for this path (`C-TAP-26` and neighbours in `tests/e2e/tap-credentials.test.ts`). Duplicating it here would buy nothing and break CI on this branch in the meantime.

## Minors

1. **Repo suggestion suppressed by a name collision.** `match.ts` only offered the GitHub-repo reading when no skill matched. Both readings can be true at once — `acme/skills` may name a skill inside tap `acme` *and* be the repo the tap itself lives at. The guard is dropped; both are offered. Covered by a new case in `tests/e2e/owner-repo-hints.test.ts`.

2. **`tap add owner/repo@v1` dropped the correction.** `ownerRepoRemedy` bailed whenever a ref was present, so the user got the generic message instead of the fix. It now carries the ref through, so the suggested command installs the revision asked for. `tap add` still rejects a pinned tap separately with its own message — that is a different error and stays that way.

3. **Any digit run accepted as a port.** `github.com:99999/o/r` passed the authority guard, reached URL parsing, and surfaced a native parser error instead of crew's. Added a `validPort` check bounding it to 1–65535; out-of-range now falls through to tap parsing and reports `invalid_ref`. New cases `C-REF-32b` and `C-REF-32c` in `tests/refs/schemeless-authority.test.ts`.

4. **Recognizer in the wrong module.** `looksLikeSchemelessHost` lived in `refs/parsers.ts` alongside the path and tap parsers. Moved to `refs/git-url.ts` with the other git-shape recognizers, and the `parsers.ts` docstring now says so. `parsers.ts` drops 158 → 120 lines.

5. **PRD §8.5 rule 4 understated the guard.** It described detection as "contains a dot", which reads as a loose heuristic when it is actually load-bearing against host spoofing. Rewritten to reference the `bare-authority` production with that rationale stated, and the grammar gains an explicit `port` production with the numeric range.

## Nit

Addressed as described in the review.

## Verification

- `bun run lint` — clean.
- `tests/refs/`, `tests/unit/shell.test.ts`, `tests/unit/format.test.ts`, `tests/e2e/suggested-command-safety.test.ts`, `tests/e2e/owner-repo-hints.test.ts`, `tests/e2e/tap-credentials.test.ts` — 184 pass, 0 fail.
- `tests/e2e/known-tap-fallback.test.ts`, `tap.test.ts`, `tap-add.test.ts`, `install.test.ts` — 108 pass, 0 fail.
- Every touched file is under the 200-line cap; `src/util/shell.ts` is at 100% coverage.

One note on the push: the branch had been rebased remotely while I worked, so I rebased this commit onto the current head rather than force-pushing over it. The push was a fast-forward and no existing commit was rewritten.

stevekrenzel:
Rebased onto the new #113 tip (which is now on `main`); import merge only, no behavior change. Refs tests pass (140).

stevekrenzel:
Rebased onto the new #113 tip (which is now on `main` after the cap-compliance refactors #135–#143). One reconciliation:

- **`tests/refs/parse.test.ts` → `tests/refs/parse/{git-urls,path-and-tap}.test.ts` (#142).** This PR's first commit added the scheme-less-host cases to the old single file, and its later commit moved them out to `tests/refs/schemeless-hosts.test.ts` because the file was over the cap. On the split layout that sequence resolves to: nothing changes under `tests/refs/parse/` (it is byte-identical to `main`), and `schemeless-hosts.test.ts` carries the `C-REF-31`/`C-REF-32` cases as before.

Verification: `bun run check` — typecheck, lint (incl. the cap gate), 1220 tests, 100% line/function coverage; site lint clean; targeted refs/install-known-tap suites 157 pass.

PR #122, opened 2026-09-11
[LOG-5766] feat: add crew outdated, a preview of what crew update would change

stevekrenzel:
**Stacked on #109** (`fix/update-dry-run`). Merge that first; GitHub will retarget this PR to `main` automatically.

## What

Adds `crew outdated [<selector>...]`, the discoverable name for "what would `crew update` change" (compare `brew outdated`).

It *is* `crew update --dry-run`: same selectors, same `--force` / `--json` behavior, same exit-code rule, and a byte-identical `--json` payload (`dry_run: true`, `would_update` / `would_add` kinds). Only the human rendering differs — it lists just the rows that answer the question (pending updates, pending tap additions, upstream removals, failures), drops `up_to_date` / `skipped` noise, and prints a single `Everything is up to date.` line when nothing would change.

```
$ crew outdated
1 skill would change

  [ok] alpha  would update  b2435f1e

[ok] 1 new skill available from repo: beta

Run `crew update` to apply.
```

## Why

`update --dry-run` was the only way to answer "what's outdated?", and it was broken until #109. A named command is easier to find in `crew help` and easier to remember.

## How

- `crew outdated` calls `planUpdate(ctx, config, home, true)` from `src/commands/update/plan.ts` — the planner #109 extracted — so both commands share one decision path and cannot disagree about what would happen. No argv alias, because the alias table only supports positional prefixes, not flags.
- Because it goes through that planner, it holds the per-tap clone locks (`withTapLocks`) for the whole plan. A preview still fetches, so it still mutates the shared clone; locking it means a concurrent real update cannot check out a different commit between this run resolving a SHA and reading that SHA's bytes. It does **not** take the state lock — acquiring that would create `state.json` on a fresh home, and §14 reserves it for commands that write.
- New `src/commands/outdated/{index,render}.ts`. Row formatting, the name cell, and tap grouping are shared with `crew update` via `src/commands/update/rows.ts`, so the two renderers cannot drift on how a row identifies itself.
- Which outcome kinds count as "would change" is an exhaustive `Record` over the row union, so adding an outcome is a compile error until someone classifies it — rather than a row silently disappearing behind "Everything is up to date."
- Registered in `src/cli/dispatch.ts` and the help index (group "Managing skills").

## Write boundary

Tap clones **are** refreshed — that is how the command learns what moved upstream. Nothing else changes: no installed skill, no marker, no stored copy, and no `state.json`. `crew help outdated` states exactly that.

## PRD

- §5.1 command list: `crew outdated [<selector>...]`
- §10.1.1: the `crew outdated` paragraph defining it by reference to `update --dry-run`
- §18.3: **C-UPD-18f** (behaves as `update --dry-run`, same write boundary), **C-UPD-18g** (leaves `state.json` absent on a fresh home; never takes the state lock), **C-UPD-18h** (a failed refresh reports possibly-stale results instead of an unqualified "up to date")

## Surfaces

- Help: new `outdated` page; `update` help gains a note and see-also.
- README command table and `site/components/sections/Commands.data.tsx`: one row each next to `crew update`.

## Tests

`tests/e2e/outdated/` — 11 tests across three suites plus shared helpers:

- `preview.test.ts` (C-UPD-18f/18g): pending update and addition listed with nothing written (snapshot of installed bytes, markers, store, and state); `--json` equals `update --dry-run --json` for the same fixture; "Everything is up to date." when nothing is pending; selectors; `state.json` absent on a fresh home; help page present.
- `stale.test.ts` (C-UPD-18h): a tap that cannot be refreshed does not read as up to date; a partial answer is flagged; a tap that cannot be acquired exits like `update --dry-run`.
- `outcomes.test.ts` (C-UPD-18f): `--force` on a pinned skill matches `update --dry-run --force` and writes nothing; a skill deleted upstream is listed; a project whose directory is gone is listed. Both outcome tests fail if their kind is dropped from the noteworthy set.

Every fixture is a local `file://` repo; no test contacts the network.


stevekrenzel:
## Review

Verdict: **not approved — 0 blockers, 2 majors, 4 minors**

### Majors

- **`outdated` mutates state despite being a read-only preview** (`src/commands/update/run.ts:48`, called from `src/commands/outdated/index.ts:18`). The shared flow always takes `withStateLock`; on a fresh home, lock setup creates `state.json`, and the command can fail with `state_locked` even though it only inspects. Run the flow without the state lock in dry-run mode and add a fresh-home test asserting neither state nor lock paths are created.
- **A failed refresh can still produce “Everything is up to date.”** (`src/commands/outdated/render.ts:50`). With a retained stale clone, refresh can fail while locally cached rows are `up_to_date`; filtering removes those rows and the renderer prints a false all-clear. Track refresh/re-expansion failures and render an indeterminate/stale-results message instead. Add a retained-clone/unreachable-origin regression test.

### Minors

- `src/commands/outdated/render.ts:46` emits two blank separator lines when warnings exist, no skill rows qualify, and `would_add` rows exist. Add the second separator only when a row section preceded additions.
- `NOTEWORTHY` in `src/commands/outdated/render.ts:25` infers as `Set<string>`, so outcome typos can silently hide rows. Type it as `Set<UpdateRow["outcome"]["kind"]>` or validate the source array with `satisfies`.
- The selector test at `tests/e2e/outdated.test.ts:113` has only one installed skill, so it cannot detect an ignored selector. Add another changed skill, preferably on another tap, and assert it is absent and its tap remains unfetched.
- The failure cases at `tests/e2e/outdated.test.ts:153` and `:165` do not assert the promised update-compatible exit codes. Assert `0` for warning-only stale-clone results and `1` when a tap cannot be acquired.

All eleven review lenses ran against head `2e0537a`; CI is green. I did not approve because of the major findings.

stevekrenzel:
Addressed all six findings in `cd4623d`. Nothing pushed back on — every finding reproduced.

**First, a note on the rebase.** This branch was cut before its parent (#109) landed `6ef2477`, which fixed the identical locking bug in `crew update`. Rather than write a second version of that fix, I rebased onto the updated parent and adopted its approach. Three conflicts came with it, resolved as described below.

---

### Major 1 — `outdated` mutated state despite being a read-only preview

**Reproduced against the pre-rebase commit** (`2e0537a`): running `crew outdated` once against a nonexistent `CREW_HOME` left an empty `state.json` behind. On the rebased branch the same run leaves only `taps/`, which is the documented, intended effect of a tap refresh.

The parent had already extracted `planUpdate` as a lock-free core but inlined it into `update/index.ts`, and deleted `run.ts` — the very file this PR added so `outdated` could share the flow. I resolved that conflict by moving `planUpdate` into `src/commands/update/run.ts` as the shared core. `crew update` wraps it in `withStateLock` and persists; `crew update --dry-run` and `crew outdated` call it directly. So this is the *same* pattern as #107 and #109, not a fourth variant — the lock/persist decision lives with the caller that actually writes.

New test `C-UPD-18e` asserts a fresh home ends with neither `state.json` nor its lock present.

### Major 2 — a failed refresh still produced "Everything is up to date."

**Reproduced:** with a retained clone and a dead origin, `crew outdated` printed the refresh warning and then, as its final line, `Everything is up to date.` The warning scrolls away; the last line is what a user reads, and it asserted something the tool had not verified.

`renderOutdated` now tracks whether any tap refresh or re-expansion failed. When it did:

- the no-changes path says `Nothing known to be out of date, but some collections couldn't be checked.` instead of the all-clear;
- the has-changes path still lists what it knows, plus a closing `Some collections couldn't be checked; results may be incomplete.`

Exit codes are unchanged — a retained clone is still a warning (0), an unacquirable tap is still a hard failure — so this is purely about not overstating confidence.

### Minors

- **Double separator** (`render.ts:46`) — the additions section now separates only from a preceding *row* section; the warning block already emits its own trailing blank.
- **`NOTEWORTHY` typing** — was `Set<string>`, so a renamed outcome kind would silently stop reporting rows. Now `ReadonlySet<UpdateRow["outcome"]["kind"]>` with a `satisfies` check on the literal, making a typo a compile error.
- **Selector test** — installed a second tap with its own pending update. The test now asserts the unselected skill is absent from the output *and* that its tap clone's `HEAD` does not move, which pins the §16.4 fetch-scope promise rather than just the rendering.
- **Exit codes** — both failure cases assert them now: `0` for the warning-only stale-clone case, and for the unacquirable-tap case an assertion that `outdated` exits with exactly what `crew update --dry-run` exits with, since the two share one flow and one exit-code rule.

### Which tests fail pre-fix

Verified by restoring the pre-fix source and keeping the new tests:

| Test | Pre-fix |
|---|---|
| `C-UPD-18f` no false all-clear | **fails** |
| `C-UPD-18f` partial-answer flag | **fails** |
| `C-UPD-18e` fresh home stays clean | **fails** (verified by reintroducing `withStateLock`) |
| selector / exit-code assertions | pass either way — coverage gaps, not live bugs |

I'm calling that out rather than letting the count imply four regressions were caught.

### Also corrected

Three user-facing surfaces still claimed `outdated` writes nothing, which is the same inaccuracy you flagged as Major 1 on #109 — a dry run *does* refresh tap clones. The site command table, `crew help outdated`, and `crew help update` now state the boundary the way the parent corrected it: collections refresh, nothing installed changes.

### Housekeeping

The test file reached 260 lines, over the hard cap, so it is now `tests/e2e/outdated/` — `helpers.ts` (106), `preview.test.ts` (117), `stale.test.ts` (98). Criteria renumbered to `C-UPD-18d` because the parent claimed 18a–18c, plus new `C-UPD-18e` and `C-UPD-18f`.

`bun run check`: 998 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean.

stevekrenzel:
## Updated /review

**Verdict:** 2 blockers, 2 majors, 4 minors. Not approved.

### Blockers

- PR description — The “How”, “PRD”, and “Tests” sections still describe `runUpdateFlow(ctx, dryRun)`, C-UPD-18a, and a single `tests/e2e/outdated.test.ts`. The diff ships `planUpdate(ctx, config, home, dryRun)`, C-UPD-18d/e/f, and split suites under `tests/e2e/outdated/`. Update the description so it accurately identifies the implementation and test inventory.
- `src/commands/help/content/outdated.ts:38-39` — “Nothing is installed or written” directly contradicts the next line and the implementation, which refreshes and writes tap clones. State precisely that installed skills, state, and store content are unchanged while tap clones refresh.

### Majors

- `src/commands/outdated/index.ts:26` — `outdated` bypasses the state lock while `planUpdate` fetches/checks out shared tap working trees. It can race a locked real update so one run resolves SHA A while another checks out B, causing copied bytes and recorded SHA to diverge. Add a source/tap lock shared by preview and update flows, or use immutable per-run checkouts.
- `src/commands/update/run.ts:50-51` — The extracted planner adds default-parameter branches, but all callers pass both values and no test enters either default. Under the repo’s 100% coverage gate this can fail the full check. Remove the unused defaults or exercise both omitted-argument paths.

### Minors

- `src/commands/outdated/index.ts:26` — Add coverage for `outdated --force` on a pinned skill and prove its JSON matches `update --dry-run --force` without mutating installed state.
- `src/commands/outdated/render.ts:34-39` — Add cases for `source_gone`, `missing_project_root`, and per-skill `failed` rows; the current suites only cover pending updates and tap-level failures.
- `src/commands/outdated/render.ts:34` — The `satisfies` list validates included values but is not exhaustive, so a future outcome can compile and disappear behind “Everything is up to date.” Use an exhaustive `Record<UpdateRow["outcome"]["kind"], boolean>`.
- `src/commands/outdated/render.ts:81,97` — Row-cell construction and tap grouping duplicate the update renderer and can drift. Extract shared helpers into the update rendering modules.

stevekrenzel:
Addressed all 8 findings in `f182369`. Nothing declined.

First, the rebase, because it resolves both majors by construction rather than by a change of mine.

**Rebase.** Three conflicts onto the revised parent (`c985aeb`): `PRD.md`, `src/commands/help/content/update.ts`, and `src/commands/update/index.ts`. Resolutions:

- `update/index.ts` — the parent already imports `planUpdate` from its new `./plan.ts`, which is exactly what this branch needed from its own `run.ts`. Took the parent's side and deleted `run.ts`.
- `help/content/update.ts` — both sides reworded the same line, and both implied nothing is touched. Neither was right, so I wrote the accurate version: "without changing any skill you have installed."
- `PRD.md` — both sides claimed `C-UPD-18d`/`18e` for different criteria. Kept the parent's numbering and renumbered this branch's to **C-UPD-18f/18g/18h**, updating the test titles and the renderer docstring so conformance failures still name the right criterion.

`origin/fix/update-dry-run` is an ancestor of the new head; PR is MERGEABLE with base unchanged.

**Major 1 — clone locking. Resolved by the rebase; I verified rather than assumed.** The parent's `planUpdate` wraps the entire plan in `withTapLocks`, and `outdated` calls that planner, so it now holds a per-tap lock across refresh, re-expansion, and every per-skill source read. The window you described — one run resolving SHA A while another checks out B — is closed for the preview path too.

**Lock ordering is unchanged: state lock outer, tap locks inner.** `update` takes `withStateLock` in `index.ts` and the tap locks inside `planUpdate`; `outdated` takes only the tap locks and never the state lock, since acquiring it would create `state.json` on a fresh home (C-UPD-18g). Tap locks are acquired in sorted name order, so two runs touching overlapping tap sets cannot deadlock. No path takes a tap lock and then reaches for the state lock, so there is no cycle. I corrected the `outdated` docstring, which said only "never takes the state lock" and would have read as "takes no locks at all".

**Major 2 — unused default parameters.** Also resolved by the rebase: `src/commands/update/run.ts` no longer exists, so the defaults you flagged are gone with it. The parent's `planUpdate` takes all four arguments explicitly and every caller passes them.

**Blocker 2 — help text.** Reproduced: the note said "Nothing is installed or written — not even state" with the next line saying clones are refreshed. It now states the boundary once — clones refresh, and no installed skill, marker, stored copy, or state file changes.

**Blocker 1 — PR description.** Rewritten against the final head, not the one this review saw: it now describes `planUpdate(ctx, config, home, true)`, the clone-lock behaviour and why the state lock is deliberately excluded, criteria C-UPD-18f/18g/18h, and the three suites under `tests/e2e/outdated/` with their 11 tests. It also states the write boundary explicitly.

**Minor 3 — exhaustiveness.** This was the substantive one. The `satisfies` list validated the kinds present but said nothing about kinds absent, so a new outcome would simply not appear and would be swallowed by "Everything is up to date." It is now `Record<UpdateRow["outcome"]["kind"], boolean>`, with the three non-noteworthy kinds explicitly `false`. Verified by temporarily adding a `relocated` kind to the union:

```
src/commands/outdated/render.ts(38,7): error TS2741: Property 'relocated' is missing in type
'{ would_update: true; source_gone: true; ... }' but required in type
'Record<"failed" | "missing_project_root" | "relocated" | ... , boolean>'
```

Variant removed afterwards.

**Minor 4 — duplication.** `nameCell` and `groupByTap` moved into `src/commands/update/rows.ts`; both renderers use them. The two `render.ts` files keep their own layout, which differs on purpose.

**Minors 1 and 2 — coverage.** New `tests/e2e/outdated/outcomes.test.ts`: `--force` on a pinned skill whose tag moved (payload equal to `update --dry-run --force`, snapshot unchanged), a skill deleted upstream (`source_gone`), and a project whose directory was removed (`missing_project_root`).

**On `failed` rows, which I did not cover — stated rather than skipped.** I could not construct a per-skill `failed` row through the CLI. Every route I tried goes through tap acquisition and surfaces as a *tap-level* failure, which `stale.test.ts` already covers with stronger assertions. My first attempt deleted the tap clone; crew simply re-cloned from the still-valid local repo and correctly reported everything up to date, so the test was asserting a premise that was false. I removed it rather than contriving a fixture, and left a comment at the bottom of the file recording why. `failed` stays in the noteworthy set, so the row renders if that path becomes reachable.

**Do the new tests fail pre-fix?** The two outcome tests fail if their kind is dropped from the noteworthy set — I flipped `source_gone` and `missing_project_root` to `false` and both failed, then restored. That is the regression you identified. The pinned-`--force` test passes either way: it guards the payload-equality contract rather than catching a present bug, and I would rather say so than let three new tests imply three caught regressions.

**Loop check**, per the resource directive: the loops I touched are `groupByTap` (one pass over a fixed array) and the two renderers' `for...of` loops over arrays built before iteration. None re-enter or mutate what they iterate. I re-read each before running anything; no test run exceeded 1 second.

**Verification, targeted only.** `tests/e2e/outdated/` 11 pass, plus `tests/e2e/update-dry-run/` 19 pass together, 0 fail. `outdated/render.ts`, `update/render.ts`, and `update/rows.ts` all at 100% line and function coverage. Typecheck and Biome clean. I did not quote a full-suite number — the suite is non-deterministic on branches without #131, and a headline count would be misleading here.

stevekrenzel:

**Verdict:** 0 blockers, 0 majors, 7 minors, 0 nits. **Approved.**

Reviewed `c985aeb..f182369` through all 11 `/review` lenses. This stacked PR has no `check` run because CI only triggers for PRs based on `main`.

### Minors

1. `missing_project_root` is counted as noteworthy although PRD §10.1.1 defines it as a skipped row (`src/commands/outdated/render.ts:41`). Align behavior and contract.
2. `tap_error` rendering drops the child name and error message (`src/commands/outdated/render.ts:63`). Preserve name, tap, code, and message as update dry-run does.
3. Help/README/site overstate new-skill reporting; additions are discovered only for tracked whole-tap installs (`src/commands/help/content/outdated.ts:12`).
4. Help says a pinned “tag or SHA has moved”; exact SHAs cannot move (`outdated.ts:19`).
5. `OutdatedJson` widens outcome discriminants to `string`; reuse `UpdateRow`/`TapReexpandRow` types (`tests/e2e/outdated/outcomes.test.ts:32`).
6. Add coverage for the reachable per-skill `failed` outcome, including exit 1 and human/JSON rendering (`outcomes.test.ts:127`).
7. The dry-run snapshot covers only SKILL.md/markers/store names, not complete installed/store bytes (`tests/e2e/outdated/helpers.ts:46`). Reuse the recursive update-dry-run snapshot.

stevekrenzel:
## Round-three minors addressed — `f6923c2`

All 7 minors worked. Each was reproduced against the exact path named before changing anything. Minor-by-minor:

---

**1. `missing_project_root` counted as noteworthy (`outdated/render.ts:41`) — fixed in code, PRD unchanged.**

This was a contract question, so I read the spec before touching either side. The reviewer's conclusion is right, though the framing shifted slightly on inspection: §10.1.1 doesn't name `missing_project_root` in either list — it enumerates the trimmed-in kinds (`would_update`, `would_add`, `source_gone`, failures) and the trimmed-out ones (`up_to_date`, `skipped`). What settles it is **C-UPD-22**, which says the row is "reported as `missing_project_root` and SKIPPED on update," plus two places in our own code that already treat it as a skip: `formatRowParts` renders it with the `"skipped"` status word, and `update/render.ts` counts it in the `skipped` tally.

So it is a skip everywhere except the `NOTEWORTHY` table, and §10.1.1 trims skips. The **code** was the side that disagreed with the contract; it now matches. No PRD update was required under the working agreement — the spec already said the right thing.

Two things preserved deliberately:
- The row still appears in `--json`. §10.1.1 requires that payload to be identical to `crew update --dry-run --json`; only the human rendering is trimmed.
- The condition is still surfaced to users — by `crew doctor` (C-STATE-11), which is the command that owns it. `crew update` wouldn't change the install either way, so it isn't an answer to "what would change?"

The existing e2e test asserted the old behavior, so it was rewritten rather than deleted: it now pins both halves of the contract (present in JSON, absent from the human view).

**2. `tap_error` dropped the child name and message (`render.ts:63`) — fixed.**

Confirmed by diffing against `update/render.ts:83-90`, which renders name, tap, code, *and* message. Ours emitted only tap and code, so an invalid child read as `tap acme (invalid_skill)` with no indication of which skill broke. It now matches update's shape. Update's own code comment makes the case better than I can: "an invalid child is the user's to fix, so `tap acme (invalid_skill)` alone would tell them nothing."

**3. Help/README/site overstated new-skill reporting — fixed in all three.**

True, and it's a real overclaim: §10.1.1 restricts sibling discovery to whole-tap installs flagged `tracks_tap: true`. Someone who installed an individual skill gets no additions, but all three surfaces promised them unconditionally. Reworded to "added to a tap you installed whole" in `help/content/outdated.ts`, `README.md`, and `site/components/sections/Commands.data.tsx`.

I also fixed the identical overclaim in `help/content/update.ts` ("what new skills your taps have added"), which this PR already touches — leaving the two help pages contradicting each other on the same fact seemed worse than a one-line extra diff.

**4. Pinned "tag or SHA has moved" — fixed.**

Simply true: an exact SHA is immutable. Per §10.1 steps 3a/3b the two cases are genuinely different — a SHA-pinned skill is skipped outright unless `--force`, while a tag can move. The flag description now describes both accurately instead of collapsing them into one wrong phrase.

**5. `OutdatedJson` widened discriminants to `string` — fixed by reusing the canonical type.**

Rather than hand-write a narrowed literal, the test now imports `UpdateJson` from `tests/e2e/update-dry-run/helpers.ts`, which already exists and already documents "discriminants keep their real types." Since §10.1.1 specifies the two payloads are identical, aliasing encodes that contract in the type system instead of restating it.

Worth noting for the record: **narrowing produced no typecheck error**, because every existing assertion already used a valid kind. Since "it compiled" proves nothing here, I verified the type is actually narrow by temporarily swapping a valid kind for a bogus one:

```
tests/e2e/outdated/outcomes.test.ts(95,34): error TS2367: This comparison appears to be
unintentional because the types '"source_gone" | "missing_project_root" | "up_to_date" |
"skipped" | "failed" | "updated" | "would_update"' and '"not_a_real_kind"' have no overlap.
```

That's the guarantee the minor was asking for: a typo'd or retired kind is now a compile error, not a silently-false assertion.

**6. Per-skill `failed` outcome uncovered — covered, and the note claiming it was unreachable was wrong.**

The old closing comment asserted a `failed` row required "a source that resolves but whose bytes cannot be read," and that every route went through tap acquisition. That's not the only route. `updateOne` calls `loadSkill(skillDir)` **before** the dry-run return — deliberately, per the comment right above it: "a broken upstream version must surface as `failed` in a preview too, not be reported as a clean `would_update`." An upstream commit that breaks frontmatter validation therefore produces a per-skill `failed` with code `invalid_skill`, no tap-level failure involved.

New test builds exactly that: install a valid skill, then commit a version upstream with `description` removed. It asserts the JSON row and its error code, that human output names the skill and does not print "Everything is up to date," exit 1 (hard failure per §10.1 error isolation), and — since it's still a preview — that nothing was written.

**7. Dry-run snapshot too shallow (`helpers.ts:46`) — replaced with the recursive version.**

Confirmed: it read only `SKILL.md` + `.crew.json` per skill, and compared the store by *directory names* only. A restage that reused the same `<name>@<short-sha>` directory, or any rewritten resource file, would have passed. It now walks both trees and compares every byte, matching `snapshotInstalledState` in the update-dry-run suite, with tap clones still excluded (they're the one documented thing a dry run does write).

---

### Verification

- `bun run lint` — clean (2 pre-existing biome-config deprecation infos).
- `bun run --cwd site lint` — clean (1 pre-existing version info).
- Targeted suites, all passing: `tests/e2e/outdated/` (12), `tests/e2e/update-dry-run/` + `help.test.ts` (19), `tap-update` / `update-dep-closure` / `update-fetch-scope` (17), `project-scope.test.ts`.
- Coverage on both changed source files at 100%/100%. Removing the `missing_project_root` branch from `outdated` moved responsibility for `rows.ts:54-58` to `project-scope.test.ts` (C-UPD-22), which exercises it through `crew update`; confirmed still covered there.
- Conflict-marker sweep: clean. All touched files under the 200-line cap.

### Deliberately not done

- **No rebase.** This is stacked on #109, and `fix/update-dry-run` has advanced by one commit (`67385b3`) since this branch forked at `c985aeb`. That commit touches no file this PR touches, so rebasing would only churn an approved PR and move the diff out from under the review range. GitHub will retarget on merge.
- **No PRD change**, per minor 1 above — the spec was already correct, so the code moved to it.

stevekrenzel:
Rebased onto the new #109 tip (now on `main`, which includes #116's aliases). Reconciled: `outdated` added to `CanonicalCommand` in `src/cli/aliases.ts` (the type #116 introduced — without it the help registry and dispatch table no longer typecheck); help registry gains `outdated` next to `main`'s new alias entries; `tests/e2e/help.test.ts` keeps `main`'s derived `EVERY_COMMAND = Object.keys(COMMANDS)` instead of this PR's edit to the old hand list; site command table has `upgrade` and `outdated` as sibling entries. Folded into the feature commit. Outdated/help/alias tests pass (36).

stevekrenzel:
Rebased onto the new #109 tip (`2b578c5`). Two modify/delete conflicts on the deleted `Commands.data.tsx`: the `outdated` entry was re-placed into `Commands.data/managing.tsx` after `upgrade`, and the later wording change ("plus new skills added to a tap you installed whole") applied on top. Nothing else conflicted. All four commits re-signed. `bun run check`: typecheck, lint (incl. the file-size gate), full suite 1118 pass at 100%; site lint clean.

PR #123, opened 2026-09-11
[LOG-5767] fix(doctor): make --repair reconcile autoupdate scheduler drift

stevekrenzel:
**Stacked on #112** (`fix/housekeeping-dry-run`). Merge that first; GitHub will retarget this PR to `main` automatically.

## What

PRD §11.2 says `crew doctor --repair` fixes "Autoupdate drift (config says enabled but the platform scheduler is not loaded, or vice versa): reconcile to the config's value", and the doctor help text claimed the same. `repairState` never did it: check 7 would flag the drift, and a `--repair` run would print "Repaired what was fixable" while leaving the scheduler exactly as it was.

Repair now runs a dedicated scheduler step (`src/commands/doctor/autoupdate.ts`):

- config enabled, scheduler not loaded → write and load the platform scheduler at the configured `interval_seconds` (the `crew autoupdate enable` path)
- config disabled, scheduler loaded → unload and remove it (the `crew autoupdate disable` path)

Each attempt is a **repair result**, distinct from a check finding: `autoupdate_loaded` / `autoupdate_unloaded` on success, `autoupdate_repair_failed` on failure with the platform error in its message. A failed repair doesn't abort the others; the run exits 1. `--json` lists them under `repairs`, present only on a `--repair` response.

**Success is measured against the scheduler, not the backend's return value**, in both directions. A load can exit 0 without the job appearing and an unload can report success while the job is still registered, so the repair re-asks `isAutoupdateLoaded()` afterwards and reports `autoupdate_repair_failed` if reality disagrees with the config.

`src/commands/doctor/coordinator.ts` owns the state lock and sequences the state rebuild and the scheduler step under it, re-reading config inside the lock so a concurrent `crew autoupdate enable`/`disable` can't be clobbered from a stale read. `repairStateUnderLock` is lock-free and focused on state, markers, and the store.

`src/autoupdate/launchd.ts`'s runner seam now returns exit status plus bounded stderr (matching systemd), so a macOS failure can name the platform error rather than offering generic advice. A stub returning a bare boolean still works.

### Relationship to #112

#112 removed `autoupdate_not_loaded` and `autoupdate_unexpectedly_loaded` from `REPAIRABLE_CODES`, correctly: standalone, it performs no reconciliation and would have promised a fix that never runs. **This PR restores both codes**, alongside the implementation that earns them. Whichever merges first, the allowlist and the PRD agree with the code in that tree.

## PRD

- §11.2: the autoupdate-drift bullet defines both directions, the repair-result vocabulary, the both-directions postcondition, the single-lock guarantee, and the `repairs` JSON field. The paragraph listing unrepairable findings no longer claims drift is merely reported.
- §18.3: C-STATE-11a (enabled + not loaded → loads), C-STATE-11b (disabled + loaded → unloads), C-STATE-11c (platform failure is a repair result, exit 1, other repairs still apply), C-STATE-11d (a repair whose effect didn't take is a failure, in both directions), C-STATE-11e (`repairs` appears only on `--repair`).

## Tests

`tests/e2e/doctor-repair-autoupdate/` drives both backends through the `setLaunchctlRunner` / `setSystemctlRunner` seams with `CREW_LAUNCH_AGENTS_DIR` / `CREW_SYSTEMD_USER_DIR` redirected, so nothing touches the real user session:

- `reconcile.test.ts` — launchd and systemd load (plist/units written at the configured interval, `bootstrap` / `enable --now` invoked); launchd and systemd unload (`bootout` / `disable`, unit files removed); and both backends reporting success while the scheduler stays loaded, which must fail.
- `failures.test.ts` — a refusing backend produces `autoupdate_repair_failed` at exit 1 with the platform diagnostic in the message, other repairs still apply, no drift is a no-op, and `--repair --dry-run` only queries.

Two inherited tests asserted the behavior this PR changes and were updated: `housekeeping-dry-run/doctor-repair.test.ts` pinned drift as unreconciled, and `misc/gaps.test.ts` pinned `--repair` aborting on an unparseable config, which #112 has since fixed to report the finding instead.

The dry-run preview changes this branch inherits belong to the stacked base (#112), not to this PR.

Targeted suites: 134 pass, 0 fail across `tests/autoupdate/`, `tests/e2e/doctor-repair-autoupdate/`, `tests/e2e/housekeeping-dry-run/`, `tests/e2e/doctor.test.ts`, `tests/commands/doctor-render.test.ts`, and `tests/misc/gaps.test.ts`. Typecheck and Biome clean; every touched file under the 200-line cap.


stevekrenzel:
## Review

Verdict: **not approved — 1 blocker, 2 majors, 3 minors**

### Blocker

- **The plist interval assertion is not awaited** (`tests/e2e/doctor-repair-autoupdate.test.ts:62`). The `.resolves` chain can finish after this synchronous test returns, allowing the core scheduler configuration check to pass vacuously. Make the test `async` and await the assertion, or use a synchronous file read.

### Majors

- **Unload failure can be reported as a successful repair** (`src/commands/doctor/autoupdate.ts:38`). `disableAutoupdate()` can ignore a failed unload command or return early when unit files are absent, yet this path unconditionally emits `autoupdate_unloaded` and exits cleanly. Surface backend unload failures and attempt unload even without local artifacts, or recheck scheduler state and emit `autoupdate_repair_failed` when it remains active.
- **Scheduler reconciliation races config changes outside the state lock** (`src/commands/doctor/index.ts:60`). `repairState()` releases the lock before reconciliation uses previously read config. A concurrent `crew autoupdate disable` can persist disabled config in between, after which doctor reloads the scheduler and recreates drift while reporting success. Hold the lock across state repair and reconciliation, and re-read config/drift inside it.

### Minors

- `src/commands/doctor/index.ts:77` always emits `repairs: []`, including non-repair JSON responses, while PRD §11.2 documents it only for `--repair`. Either specify that the field is unconditional or emit it only for repair responses.
- `src/commands/doctor/autoupdate.ts:18` models the stable repair contract as broad `Finding` values that permit arbitrary codes and `warn`. Define a dedicated repair result with literal repair-code and `"ok" | "error"` unions.
- `tests/e2e/doctor-repair-autoupdate.test.ts:121` does not prove failure findings include the required platform diagnostic. Trigger a backend failure with distinctive stderr and assert that text appears in `autoupdate_repair_failed.message`.

All eleven review lenses ran against head `32df8b9`; CI is green. I did not approve because of the blocker and major findings.

stevekrenzel:
Addressed all six findings in `00a0d11`. I rebased onto the current `fix/housekeeping-dry-run` first, since #112 had moved and its `REPAIRABLE_CODES` allowlist interacts directly with this code — details at the end. Nothing pushed back on; every finding reproduced.

## Blocker — unawaited plist assertion

Reproduced, and the failure mode is worse than "passes vacuously": I replaced the expected interval with a string that could never match, and the test *did* fail — after **30 seconds**, reported as `this test timed out after 5000ms` with no mention of the assertion. So it isn't silently green, but a real mismatch shows up as an unrelated timeout that reads like flake, and costs 30s of suite time doing it.

Now a synchronous `readFileSync`. Demonstrated it is real: with the plist written at a wrong interval, the test fails in **12ms** with `Expected to contain: "<integer>900</integer>"` and the actual plist body. Restored afterwards.

**Suite-wide sweep:** the only `.resolves`/`.rejects` in `tests/` was this one. No other floating promises (no promise-returning file IO outside `await`, one unrelated `async` test). It was a single instance, not a pattern.

## Major 1 — unload failure reported as success

Reproduced on both halves, and both are now fixed at the backend:

- `launchd.disableAutoupdate` discarded `runLaunchctl`'s boolean, and returned early when the plist was absent — but launchd can still hold a loaded job whose file was removed. It now attempts `bootout` regardless of local artifacts and raises `autoupdate_failure` when both commands fail and the job is still loaded.
- `systemd.disableAutoupdate` checked its exit status correctly but shared the early return. It now skips only when the timer is genuinely inactive.

`doctor/autoupdate.ts` additionally verifies the **actual** scheduler state after a load rather than trusting the backend, because `launchctl bootstrap` can exit 0 without the job appearing — nothing throws in that case, so only a post-repair state check catches it. Two new tests (C-STATE-11d) cover both directions; both fail against `32df8b9` — the unload case exited **0** pre-fix, reporting a successful repair while the job was still loaded.

One narrowing worth flagging: the state check only guards the *load* direction. Both backends now raise on a failed unload, so the unload arm was unreachable, and I removed it rather than keep a branch no test could reach.

## Major 2 — reconciliation raced config outside the lock

Reproduced the structure, then hit a concrete problem while fixing it. My first attempt wrapped `repairState` in `withStateLock` — but `repairState` **already takes that lock internally**, so every repair test deadlocked for the full 30s timeout and exited 7 (`state_locked`). Nesting the lock would have been a worse bug than the race.

`repairState` now takes an `andThen` callback that runs under the same lock before release, with config and drift re-read inside it. One lock boundary covers the rebuild and the reconciliation; read-only runs still never lock, so a plain `doctor` doesn't create `state.json` (the class #107/#109/#112/#122 fixed).

## Minors

- **`repairs: []` on non-repair responses** — now `--repair`-only, matching §11.2. Added C-STATE-11e, including a plain-run test asserting the key is absent, which fails pre-fix.
- **Broad `Finding` for the repair contract** — replaced with `AutoupdateRepair`: `level: "ok" | "error"` and three literal codes. `Finding` permitted arbitrary codes and a `"warn"` level that had no meaning for a repair.
- **Platform diagnostic in the failure message** — added a systemd test asserting `Unit is masked.` reaches `autoupdate_repair_failed.message`. I used the systemd seam deliberately: the launchctl seam returns a bare boolean and carries no stderr to propagate.

## Rebase

Three conflicts, all resolved toward the parent:

1. `doctor/index.ts` — kept the parent's exit-code logic (unrepairable errors stay blocking; my version exited 0 whenever repairs succeeded) and added failed repairs as an additional blocking condition.
2. `doctor/render.ts` — kept the parent's addressed/left accounting and added the per-repair lines. A failed scheduler repair no longer counts as addressed.
3. `tests/e2e/housekeeping-dry-run.test.ts` — the parent split it into a directory; took the deletion. My one-line change there was the `repairs: []` assertion that Minor 1 disputes, so it was moot.

`origin/fix/housekeeping-dry-run` is an ancestor of the new head; the PR still targets it and reports MERGEABLE.

## Two things worth knowing

**An existing test encoded the old no-op behavior.** `tests/autoupdate/systemd.test.ts` asserted systemctl is never called when unit files are absent, with a stub that throws on any call. That intent is legitimate, but the stub can't express "the timer is inactive". I kept the intent — `is-active` must be the only call — and added the dangerous case it couldn't express: no unit files, timer still loaded, which must still disable.

**A pre-existing gap I pinned rather than fixed.** `doctor --repair` against an unparseable `config.yaml` aborts with exit 4 and rebuilds nothing, because `repairState` re-reads config and throws — while plain `doctor` reports the same file as a finding. Reproduced on the base branch, so it predates this PR. I added a test pinning the current behavior with a comment, rather than asserting the graceful degradation I'd initially assumed. Making `--repair` degrade here is its own change.

`bun run check`: 1018 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean. `doctor/render.ts` crossed the 200-line cap during this work, so the repairable-code allowlist moved to `doctor/repairable.ts`; every touched file is now under the cap.

stevekrenzel:
## Updated /review

**Verdict:** 2 blockers, 2 majors, 3 minors. Not approved.

### Blockers

- `src/commands/doctor/autoupdate.ts:91-98` — The repair verifies scheduler state only after enable. A disable can report `autoupdate_unloaded` and exit 0 even when the scheduler remains loaded, violating the new C-STATE-11d postcondition. Always compare `isAutoupdateLoaded()` with `wantLoaded` after either direction and add launchd/systemd tests where disable reports success but the job remains active.
- PR description — The Tests section names `tests/e2e/doctor-repair-autoupdate.test.ts` and says this PR updates `housekeeping-dry-run.test.ts`, but neither is in this frozen diff. Point to the split `doctor-repair-autoupdate/{reconcile,failures}.test.ts` suites and attribute inherited dry-run changes to the stacked base.

### Majors

- `src/autoupdate/launchd.ts:119-129` — launchctl stderr is discarded, so macOS repair failures can only return generic advice even though PRD §11.2 promises `autoupdate_repair_failed.message` carries the platform error. Preserve bounded stderr and include it in the failure result, as the systemd path does.
- `PRD.md:1308,1914` — `autoupdate_repair_failed` is called a “finding,” but the implementation emits it as an error-level item in the separate `repairs` array. “Finding” already names doctor check results, making the external contract ambiguous. Call it a repair result/outcome consistently.

### Minors

- `src/commands/doctor/autoupdate.ts:89` — `err as Error` can turn a non-`Error` throw into an `undefined` diagnostic. Narrow with `err instanceof Error ? err.message : String(err)`.
- `tests/e2e/doctor-repair-autoupdate/reconcile.test.ts:43` — Linux covers only enabled→loaded. Add systemd disabled→unloaded and “disable succeeded but still active” cases, including unit-file removal and interval forwarding.
- `src/commands/doctor/repair.ts:28` — `repairState` now acts as a generic hook for arbitrary work under the state lock. Introduce a doctor-repair coordinator that owns the lock and scheduler step, leaving the state repair helper focused on state/marker/store repair.

stevekrenzel:
Addressed all 7 findings in `d91898e`. Rebased onto the current `fix/housekeeping-dry-run` first — details at the end. Nothing declined.

## Blocker 1 — repair verified only the load direction

Reproduced before fixing. `attempt()` gated its state check on `wantLoaded`, so the unload arm returned `{ level: "ok", code: "autoupdate_unloaded" }` on any backend that didn't throw. A launchctl stub answering `true` to everything — including `list` — produced exit 0 and a successful repair while the job was still loaded.

The check now compares `isAutoupdateLoaded()` against the intended state after **either** direction. My round-one note claimed the unload arm was unreachable because both backends raise on a failed unload; that reasoning only covers a backend that *reports* failure, and this is the case where it reports success and the job survives anyway. The comment saying so is gone.

Two new tests in `reconcile.test.ts` under C-STATE-11d, one per backend. **Both fail against the pre-fix commit** and only those two: restoring the one-directional check gives 4 pass / 2 fail in that file.

## Blocker 2 — PR description

Rewritten. It now describes the coordinator, the both-directions postcondition, the split `doctor-repair-autoupdate/{reconcile,failures}.test.ts` suites, and the C-STATE-11a–11e criteria. The inherited dry-run preview changes are attributed to #112 rather than claimed here.

## Major 1 — launchctl stderr discarded

Confirmed: `defaultRunner` captured `stderr` into a pipe and returned a bare boolean, so the diagnostic was read and dropped. The seam now returns `{ ok, stderr }` with stderr trimmed to 500 characters, matching the systemd seam's existing shape. A stub returning a plain boolean still works — `runLaunchctl` normalizes both — which keeps the existing launchd tests valid and is why `helpers.ts`'s `statefulLaunchctl` needed no change.

## Major 2 — "finding" vs repair result

§11.2 and C-STATE-11c/11d now call `autoupdate_repair_failed` a repair result consistently, since it is emitted in the `repairs` array while "finding" already names check results. The §11.2 paragraph listing unrepairable findings previously ended by saying drift "is likewise reported but not reconciled" — true for #112, false here — and now states that drift is repairable but counts as addressed only when its repair result succeeds.

## Minors

1. `err as Error` → `err instanceof Error ? err.message : String(err)`. A non-Error throw would otherwise have produced `undefined` in the message.
2. Linux coverage extended: systemd disabled→unloaded with unit-file removal, and the systemd half of the both-directions case above. The enabled→loaded test already asserted interval forwarding.
3. New `src/commands/doctor/coordinator.ts` owns the state lock and sequences the rebuild and the scheduler step; `repairState` became `repairStateUnderLock`, lock-free and focused on state, markers, and the store. This also resolved the rebase conflict cleanly, since #112 and this PR had modified the same function for different reasons.

## Scheduler codes are back in `REPAIRABLE_CODES`

Stated explicitly since it spans two PRs: #112 removed `autoupdate_not_loaded` and `autoupdate_unexpectedly_loaded` because, standalone, it reconciles nothing and would promise a fix that never runs. **This PR restores both**, together with the implementation and the PRD promise. Whichever merges first, the allowlist matches the behavior in that tree.

## Rebase

Four conflicts, all resolved keeping both sides' intent:

- `repair.ts` — #112's missing-project-root preservation plus this PR's under-lock sequencing, reconciled by the coordinator extraction.
- `index.ts` — kept #112's `config !== null` guard (repair rewrites `config.yaml` taps from markers, so it must not run against a file we couldn't read) and routed reconciliation through `applyRepairs`.
- `render.ts` — kept #112's `applied` flag, which is more accurate than `repair && !dryRun` because it also accounts for the skipped-on-invalid-config case; dropped a stale duplicate of the allowlist that #112 had already extracted to `repairable.ts`.
- `PRD.md` — as described under Major 2.

`origin/fix/housekeeping-dry-run` is an ancestor of the new head; the PR targets it and reports MERGEABLE.

## Two inherited tests changed

Both asserted behavior this PR or its parent deliberately changed, so I updated them rather than working around them:

- `housekeeping-dry-run/doctor-repair.test.ts` C-STATE-12b asserted drift is reported and survives a repair. That is #112's correct behavior and wrong here; it now asserts the drift is reconciled and a follow-up check comes back clean.
- `misc/gaps.test.ts` pinned `doctor --repair` aborting with exit 4 on an unparseable config. I pinned that in round one as a pre-existing gap; #112 has since fixed it to report the finding, so the test now asserts exit 1 with `config_invalid` among the findings and the orphan left in place.

## Verification

Targeted only, given the known non-determinism on branches without #131: 134 pass / 0 fail across `tests/autoupdate/`, `tests/e2e/doctor-repair-autoupdate/`, `tests/e2e/housekeeping-dry-run/`, `tests/e2e/doctor.test.ts`, `tests/commands/doctor-render.test.ts`, and `tests/misc/gaps.test.ts`. Typecheck and Biome clean. Every touched file is under the 200-line cap (largest: `launchd.ts` at 183).

Per the resource directive: the only loops I touched are bounded `for` iterations over fixed arrays with no mutation of the iterated sequence, so they terminate. I re-read them before running anything; no test run exceeded a second.

stevekrenzel:
Rebased onto the revised parent `c4fd00f`. Branch is now `eeb4dab`.

**Five conflicts, all resolved keeping both sides:**

|---|---|---|
| `doctor/repairable.ts` | add/add — both branches created this file independently | Parent's fuller docstrings + this branch's two scheduler codes |
| `doctor/index.ts` | Parent imports `repairState`; this branch routes through `coordinator.ts` | This branch's — the coordinator owns the lock now |
| `doctor/render.ts` | Parent added `isRepairableCode` for repair-preview marking | Kept it |
| `doctor/repair.ts` | Parent added `isDirectory`; this branch dropped `withStateLock` | Both: `isDirectory` is used, `withStateLock` is not (coordinator holds it) |
| `help/content/doctor.ts` | Parent disclaims scheduler reconciliation; this PR implements it | This branch's wording |

**Scheduler codes re-added in the new policy home.** The parent deliberately removed `autoupdate_not_loaded` and `autoupdate_unexpectedly_loaded` from `REPAIRABLE_CODES` so it is honest standalone, with a comment saying they rejoin "in the same change that implements `repairAutoupdateDrift`" — that is this PR. Both are back in `src/commands/doctor/repairable.ts` (the parent's new home, not a stale copy), and `grep REPAIRABLE_CODES src/` returns exactly one table.

**One thing beyond the mechanical resolution.** The parent also added a comment in `autoupdate/render.ts` explaining that `crew doctor --repair` is deliberately *not* offered on the drift path because it "does not reconcile it". That is false on this branch, so the status output now offers `--repair` alongside the manual reset. Without this, `--repair` would reconcile drift while `autoupdate status` told users it would not. No test pinned the old string.

**Verified:** parent is an ancestor; `doctor-repair-autoupdate/` 13 pass; `housekeeping-dry-run/` + `doctor` + `doctor-render` 42 pass; 0 fail. Typecheck and Biome clean.

stevekrenzel:

**Verdict:** 0 blockers, 1 major, 7 minors, 1 nit. **Not approved.**

Reviewed the updated frozen diff `c4fd00f..eeb4dab` through all 11 `/review` lenses. This stacked PR currently has no test-suite CI check.

### Major

1. **A failed post-repair scheduler probe is treated as confirmed “not loaded”** — `src/commands/doctor/autoupdate.ts:96`
   `isAutoupdateLoaded()` reduces probe failures to `false`. After disable, a failed `systemctl is-active` or `launchctl list` check can therefore be reported as `autoupdate_unloaded` with exit 0 even though the required postcondition was never verified. Preserve a tri-state/result-bearing probe and return `autoupdate_repair_failed` for an indeterminate check.

### Minors

1. `doctor --repair --json` omits `repairs: []` when invalid config prevents applying repairs (`src/commands/doctor/index.ts:63`).
2. `AutoupdateRepair` permits inconsistent `level`/`code` pairs; make it a discriminated union (`src/commands/doctor/autoupdate.ts:37`).
3. Launchd's spawn catch discards the process error text (`src/autoupdate/launchd.ts:159`).
4. Launchd fallback diagnostics prefer the first failure, hiding a later actionable load/unload failure (`src/autoupdate/launchd.ts:87`).
5. Help describes JSON as findings only and omits the repair-result array (`src/commands/help/content/doctor.ts:26`).
6. Add a launchctl structured-stderr propagation test (`tests/e2e/doctor-repair-autoupdate/failures.test.ts:28`).
7. Add a JSON dry-run assertion that `repairs` is omitted (`failures.test.ts:135`).

### Nit

1. `repairable.ts` retains stale, unused `repairState` mechanism labels; use a set or update the labels.

stevekrenzel:
Round 3 addressed in 24d9d9a. Per finding:

**Major — a failed confirming probe reported the repair as successful.** Confirmed and fixed. `isAutoupdateLoaded()` returned `false` for two very different situations: the scheduler answering "not loaded", and the probe not running at all (launchctl unreachable, no systemd user bus). So a disable whose confirming probe failed emitted `{"level":"ok","code":"autoupdate_unloaded"}` — a postcondition claimed but never checked, which is precisely the class of bug the round-2 verification was added to close.

Reproduced before fixing: with the unload succeeding and only the confirming `launchctl list` failing with stderr, the output was the "ok" above. After the fix it is:

```json
{"level":"error","code":"autoupdate_repair_failed","message":"couldn't reconcile the background updater — couldn't confirm the scheduler's state afterwards — launchctl: Could not connect to the service"}
```

The mechanism is a tri-state `SchedulerProbe` (`loaded` / `not-loaded` / `indeterminate`, plus the platform's own diagnostic) on both backends and on the `Scheduler` dispatcher. `attempt()` checks `indeterminate` first and fails the repair with the diagnostic attached. `isAutoupdateLoaded()` now delegates to the probe, so detection is unchanged.

The distinction is load-bearing, not cosmetic: a bare non-zero exit from `launchctl list` genuinely means "not loaded", and treating that as indeterminate would break detection. Stderr is what separates the two.

**Minor — `AutoupdateRepair` let level and code disagree.** Converted to a discriminated union: `level: "ok"` pairs only with the non-failure codes, `level: "error"` only with `autoupdate_repair_failed`. The pairing is the external JSON contract, and a type that permits a mismatch lets the output lie in a way no test necessarily catches.

**Minor — only the first launchctl diagnostic survived.** `enable` tries bootstrap and falls back to load; the fallback's failure is usually the actionable one. Distinct stderr from every attempt is now joined with `"; "`.

**Minor — the default launchctl runner discarded the thrown message.** Its catch returned `stderr: ""`, so a spawn failure reached the user as a repair failure with no reason. It now preserves the message.

**Minor — an unparseable config omitted `repairs` from `--json`.** Gating moved from `applied` to `repair && !dryRun`. Repair is skipped in that case (rebuilding taps from markers against a config we couldn't read would discard the user's file), but the field must still appear so "the repair ran and fixed nothing" stays distinguishable from "this wasn't a repair run". A dry run still omits it, since nothing was applied.

**Minor — `REPAIRABLE_CODES` carried stale mechanism labels.** Its values named `repairState`, which stopped being where the work happened. Nothing read them, so it is now a `ReadonlySet`. Which function does the work belongs next to the work.

**Minor — `doctor --json` help didn't mention the repair array.** Updated.

**Minor — the new test keyed off a launchctl call count.** Rewritten to key off whether the unload actually happened, so it won't silently stop testing the intended path if detection changes how often it probes.

**Nit — comment wording.** Applied.

PRD §11.2 documents the three-valued confirming query and the `repairs` field semantics; criterion C-STATE-11f is new, with tests on both launchd and systemd. Removing the indeterminate handling fails exactly those two and nothing else. 18/18 in `tests/e2e/doctor-repair-autoupdate/`, typecheck and lint clean.

stevekrenzel:
Retargeted to `main` by GitHub when #112 merged; rebased with #112's pre-merge tip as the cut so only this PR's three commits replay. CI then failed on the 100% coverage gate: `src/commands/doctor/autoupdate.ts` had one uncovered line — the `);` closing a multi-line `return failure(…)` whose both arms are tested (the closing-line quirk documented in CLAUDE.md). This PR never had CI before (stacked), so it was never caught. Restructured to compute the message first and return on one line; identical messages, 100% again.

stevekrenzel:
Rebased onto `main` after the cap-compliance refactors (#135–#143) merged. Two things changed:

- **Test re-placed:** this PR's `doctor --repair reports an unparseable config…` test lived in `tests/misc/gaps.test.ts`, which #141 split by module; it now sits with the other doctor/orphan-store cases in `tests/misc/closeout/doctor-orphan-store.test.ts`, verbatim.
- **The new file-size gate (#137) flagged three files this PR grows**, fixed in one added commit with no behavior or assertion change: `src/autoupdate/launchd.ts` (133 on main → 219 here) — the `launchctl` subprocess seam (runner, `setLaunchctlRunner`/`resetLaunchctlRunner`, result shape) moves to `src/autoupdate/launchctl.ts` (78), leaving `launchd.ts` (150) with plist and job management; importers re-pointed by name and `CLAUDE.md`'s seam references updated. `tests/autoupdate/systemd.test.ts` (223) → `systemd/{enable,disable}.test.ts`. `tests/e2e/doctor-repair-autoupdate/failures.test.ts` (251) → `failures` (11c/11d), `unverifiable` (11f), `repairs-field` (11e, C-STATE-12).

Full `bun test`: 1112 pass, 0 fail, 100% line and function coverage; lint incl. the gate and typecheck clean.

PR #124, opened 2026-09-11
[LOG-5768] feat(list): add --agent and --tap filters

stevekrenzel:
Stacked on #118 (`fix/list-scope-filter`); merge that first.

## What

Three observable CLI changes, not just the list filters:

**1. `crew list` gains two filters** alongside `--scope`:

- `--agent <name>` (repeatable, any-of): only rows installed into at least one of the named agents. An unknown agent is a `usage_error` listing the known agents.
- `--tap <name>`: only rows attributed to that configured tap. An unknown tap is a `usage_error` pointing at `crew tap list`.

Filters compose with each other and with `--scope`. A surviving row still renders its full agent list — the filter selects rows, it doesn't hide data. `--json` gains `agent: string[]` and `tap: string | null` next to `scope`. When the combination matches nothing and an agent or tap filter was given, the human output says no skills match those filters; a scope-only miss keeps #118's scope-specific message.

**2. Repeated single-value flags are rejected** (`rejectRepeatedScalars` in `src/cli/args.ts`). Previously `crew list --tap core --tap typo` exited 0 with `"tap": null` and an unfiltered list, validating neither value. This is the general rule now specified in **PRD §5.2** (added by #117, criteria C-CLI-08b/08c) — this PR carries the same guard rather than inventing its own, and §5.1 defers to §5.2 so the two can't drift. #111 and #117 carry it too; whichever merges first, the others resolve as a trivial duplicate.

**3. Bare command aliases inherit their canonical command's flag table** (`FLAG_TABLE_ALIASES`). `crew skills --tap core` previously exited 4 with "Unknown argument: tap", because the flag table was keyed on the raw command. Prefixed aliases are deliberately excluded: `taps` → `tap list` and `untap` → `tap remove` target subcommands that ignore `tap`'s flags, and inheriting that table would start accepting `--recursive` — which #116's review required stay rejected. A guard test pins that.

## Why

With 19 adapters, "what's in codex?" and "what came from `core`?" are the two questions `crew list` couldn't answer without `jq`. The other two changes are defects the filter work exposed: a repeated flag silently disabling filtering is worse than an error, and a documented alias rejecting a flag its canonical command accepts is a broken promise.

## PRD

- §5.1: "`crew list` agent and tap filters" paragraph, plus the alias flag-table rule; the repeated-flag sentence defers to §5.2.
- Criteria **C-LIST-04, C-LIST-05, C-LIST-06, C-LIST-07** (C-LIST-01..03 come from #118).

## Files

12 files. Implementation:

- `src/commands/list/filters.ts` (new): flag validation and row filtering.
- `src/commands/list/index.ts`, `render.ts`: wiring and the empty-state message.
- `src/cli/args.ts`: `list: ["tap"]`, `FLAG_TABLE_ALIASES`, `rejectRepeatedScalars`.
- `src/agents/validate.ts` (new): `assertKnownAgents`, shared by install, uninstall, and list so the unknown-agent wording can't drift. Its `code` parameter is the narrow `"usage_error" | "no_agents"` union rather than any `CrewErrorName`.
- `src/commands/uninstall/index.ts`, `src/install/agent-set.ts`: use the shared validator.
- `src/commands/help/content/list.ts`, `README.md`: flags and examples.

Tests: `tests/e2e/list-filters.test.ts` (new, 9 tests), plus one prefixed-alias guard in `tests/e2e/aliases.test.ts`.

## Tested

Targeted suites: `list-filters` 9 pass, `list-scope` + `aliases` 12 pass, `uninstall-scope` + `agent-set` 8 pass, 0 fail. `src/agents/validate.ts` and `src/commands/list/filters.ts` at 100% line and function coverage. Typecheck and Biome clean.

Per-suite numbers rather than a full-suite total: the suite is non-deterministic on any branch lacking #131, because the default `core` tap URL points at GitHub and tests that don't discard it attempt live clones.


stevekrenzel:
## Review

Verdict: **not approved — 1 blocker, 2 majors, 3 nits**

### Blocker

- **The test can write into real detected agents on the host** (`tests/e2e/list-filters.test.ts:49`). Its default install targets every detected adapter, while the setup redirects only Claude Code and Codex. On a machine where another adapter is detected, the test can escape its temp roots and its expected agent list becomes environment-dependent. Install `alpha` with explicit `--agent claude-code --agent codex`.

### Majors

- **The existing `skills` alias rejects the new list flag** (`src/cli/args.ts:47`). `crew skills --tap core` is parsed against the raw `skills` command before dispatch aliases it to `list`, so it rejects a flag the documented alias should support. Normalize aliases before selecting flag tables (or explicitly mirror the table) and add alias coverage.
- **Repeating `--tap` silently disables all tap filtering** (`src/cli/args.ts:120`, `src/commands/list/filters.ts:25`). Repeated singular flags become an array, `extras` drops it, and `crew list --tap core --tap typo` returns an unfiltered list without validating either value. Reject repeated `--tap` values as `usage_error` and add a regression test.

### Nits

- Rename `hasRowFilter` at `src/commands/list/filters.ts:46` to `hasAgentOrTapFilter`; scope also filters rows, while this boolean specifically selects the combined-filter empty message.
- Replace the double cast in `tests/e2e/list-filters.test.ts:17` with a direct `Mut[]` assignment so adapter contract changes remain type-checked.
- Unknown-agent validation at `src/commands/list/filters.ts:50` duplicates install/uninstall behavior and wording; extract a shared registry-bound validator.

All eleven review lenses ran against head `4603e4c`. The normal CI check is absent from this PR’s status rollup. I did not approve because of the blocker and major findings.

stevekrenzel:
Addressed all six findings in `ca9cb2f`. Rebased onto the current parent first (`5ffde68`) — clean, no conflicts; `fix/list-scope-filter` is an ancestor and the PR is still MERGEABLE.

**Blocker — test could target undeclared adapters** (`tests/e2e/list-filters.test.ts`). Reproduced by marking a third adapter (`cursor`) detected with its own temp `userPath`: an unrestricted `crew install` picked it up and the skill landed in it, so the expected agent lists were environment-dependent. Both installs in `seed()` now name their agents explicitly.

One correction to the finding's framing, with evidence. A *host* escape isn't reachable today: `tests/helpers/preload.ts` calls `neutralizeAdaptersExcept([])`, which points every adapter at an inert tmp dir and forces `detect() → false` before any test runs — I confirmed only `claude-code` is detected under the preload, and its `userPath` is already inert. So the real defect is the environment-dependent assertion, not writes to a developer's `~/.claude/skills/`. The fix is the same either way; I'm flagging it so the severity isn't mis-recorded.

**Sweep for the same class.** I audited every test file that re-enables `detect` (25 files) and compared against `userPath` redirections in the same file. Every one redirects at least as many adapters as it enables — no test leaves a detected adapter pointing at a real host path. The preload plus per-file redirection is defence in depth; this PR's gap was the missing third layer (restricting the install itself).

**Major 1 — `skills` alias rejected `--tap`.** Reproduced: exit 4, "Unknown argument: tap". The flag table was keyed on the raw command, and `skills` has no entry. Added `FLAG_TABLE_ALIASES` in `src/cli/args.ts` so bare aliases inherit their canonical command's table. Prefixed aliases are deliberately excluded — `taps` → `tap list` and `untap` → `tap remove` target subcommands that ignore `tap`'s flags, and inheriting that table would start accepting `--recursive`, which is exactly what #116's review required stay rejected. Verified both still exit 4 on `--recursive`, and added a guard test in `tests/e2e/aliases.test.ts` so a future change can't loosen it silently.

**Major 2 — repeated `--tap` silently disabled filtering.** Reproduced: `crew list --json --tap core --tap typo` exited 0 with `"tap": null` and a full unfiltered list, validating neither value. Ported `rejectRepeatedScalars` verbatim from #111/#117 — same name, placement, and message — so whichever lands first, the others resolve as a trivial duplicate rather than three divergent guards.

**Nits.** `hasRowFilter` → `hasAgentOrTapFilter`. The double cast is now a typed `Mut[]` assignment built from `AgentAdapter`'s own member types, so an adapter contract change is a type error rather than being erased. Unknown-agent validation is extracted to `src/agents/validate.ts` (`assertKnownAgents`) and shared by install, uninstall, and list — paralleling #117's `src/config/find-tap.ts`. The three call sites had byte-identical messages but differing error codes (`no_agents` for install's agent set, `usage_error` elsewhere), so the code is the one parameter; that distinction is preserved and commented.

**Pre-fix verification.** Stashed `src/` only, keeping the tests: both new major regressions fail (`repeated --tap` and `skills alias`), 14 pass. The prefixed-alias test passes either way — it guards against future loosening rather than catching a present bug, so I'm not counting it as a third regression.

**PRD.** §5.1 gains the repeated-flag and alias-flag rules; criterion C-LIST-07 added after C-LIST-06.

`bun run check`: 1008 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean. All touched files under the 200-line cap (largest: `src/cli/args.ts` at 193).

stevekrenzel:
## Updated /review

**Verdict:** 1 blocker, 0 majors, 2 minors, 1 nit. Not approved.

### Blocker

- PR description — It says only C-LIST-04..06 were added and frames the change solely as list filtering, but the diff also adds C-LIST-07, global repeated-scalar rejection, and `skills` alias flag-table behavior. Update the description to cover all observable CLI changes.

### Minors

- `src/agents/validate.ts:25` — The helper accepts any `CrewErrorName` even though its documented valid outcomes are only `usage_error` and `no_agents`. Narrow the parameter to that union so future callers cannot compile with an unrelated error code.
- `tests/e2e/list-filters.test.ts:92` — Prefix this end-to-end suite’s top-level `describe` with `E2E: ` to follow the repository’s test classification convention.

### Nit

- `tests/e2e/list-filters.test.ts:15` — The adapter test-seam type `Mut` does not communicate what is mutable. Rename it to something descriptive such as `RedirectableAdapter`.

stevekrenzel:
Addressed the blocker, both minors, and the nit in `160aa58`. **One minor declined with evidence** — details below.

Rebased onto the current parent (`af11799`) first: clean, no conflicts. `fix/list-scope-filter` is an ancestor and the PR is MERGEABLE with its base unchanged.

**Blocker — description drift.** Rewritten. It now leads with all three observable CLI changes rather than framing the PR as list filtering alone: the two filters, the repeated-single-value-flag rejection, and bare aliases inheriting their canonical command's flag table (including why prefixed aliases are deliberately excluded). Criteria corrected to C-LIST-04..07, files to the actual 12, and tests to the 9 in `list-filters.test.ts` plus the alias guard in `aliases.test.ts`.

On the shared guard: the description now points at **§5.2's general flag-uniqueness rule** (added by #117, criteria C-CLI-08b/08c) rather than implying this PR invents its own.

I also found the duplication you'd expect from that. My §5.1 paragraph restated the repeated-flag rule in its own words, which is exactly how two sections drift. §5.1 now defers to §5.2 explicitly, and C-LIST-07 cites both sections. No conflict between the two texts — §5.2 states the rule, §5.1 names `--tap` as an instance of it.

**Minor 1 — `assertKnownAgents` parameter too wide** (`src/agents/validate.ts`). Fixed. The docstring already said only `usage_error` and `no_agents` are valid outcomes while the signature accepted any `CrewErrorName`, so the constraint lived in prose. It's now a named `UnknownAgentCode` union, so a caller reaching for a third code is a compile error. This is compile-time hardening, not an observable bug — all three existing callers (`install/agent-set.ts` with `no_agents`, `uninstall/index.ts` and `list/filters.ts` with the default) already pass a member, which is why typecheck stayed green.

**Minor 2 — `E2E: ` describe prefix. Declined.** I checked the convention before applying it and could not find it:

```
e2e files whose top-level describe starts with "E2E:"   0
e2e files with a top-level describe                    27
```

CLAUDE.md's only test-naming rule is the criterion-ID one ("Each conformance test names its criterion ID in the test title"), and `grep -rn "E2E:"` across `CLAUDE.md`, `PRD.md`, and `tests/` returns nothing. Existing suites read `describe("crew info", …)`, `describe("uninstall --prune", …)`, and so on.

So prefixing this one file would make it the only file in the repo with that prefix — creating the inconsistency the convention is meant to prevent, rather than following one. Happy to apply it everywhere if you'd like the convention established, but that's a repo-wide change worth its own PR rather than a one-file exception smuggled into this one. Tell me which and I'll do it.

**Nit — `Mut` rename.** Done: `RedirectableAdapter`, with a docstring saying what it permits (the three adapter members a test may point at a temp root) rather than naming the mutation.

**Verification.** No loops changed — the three edits are a type narrowing, a type rename, and prose — so there was nothing to re-check for termination. Targeted suites: `list-filters` 9 pass; `list-scope` + `aliases` 12 pass; `uninstall-scope` + `agent-set` 8 pass (the other `assertKnownAgents` callers, run because I narrowed its signature); 0 fail. `src/agents/validate.ts` and `src/commands/list/filters.ts` both at 100% line and function coverage. Typecheck and Biome clean.

Per-suite numbers rather than a full-suite total, given the known non-determinism on branches without #131.

stevekrenzel:
Rebased onto the revised parent `fix/list-scope-filter` (`5533de0`), which itself had just been rebased onto its own parent. Four conflicts, all resolved by keeping the parent's round-two corrections alongside this branch's additions:

- `src/commands/list/render.ts` — docstring: the parent corrected "source column" to "name column". Kept the parent's.
- `tests/e2e/list-scope.test.ts` — add/add. Took the parent's version, which carries its round-two improvements (`Scope | null` typing and the strengthened user-row assertion).
- `PRD.md` ×3 — each conflict paired the parent's corrected C-STATE-10 wording (`(scope, project_root)` install location) against this branch's new criteria. Kept the parent's C-STATE-10 in every case, plus C-LIST-04..06 and the round-two C-LIST-07 that cites §5.2's flag-uniqueness rule.

Verified: parent is an ancestor, typecheck and Biome clean, `list-filters` + `list-scope` 14 pass / 0 fail. MERGEABLE.

stevekrenzel:

**Verdict:** 0 blockers, 0 majors, 5 minors, 0 nits. **Approved.**

Reviewed `5533de08..dcd9f112` through all 11 `/review` lenses. This stacked PR has no test-suite CI run at this head.

### Minors

1. PRD §5.1 cites a general flag-uniqueness rule that §5.2 never defines (`PRD.md:163`). Add the normative rule and repeatable exceptions.
2. `FLAG_TABLE_ALIASES` duplicates `COMMAND_ALIASES` without type linkage (`src/cli/args.ts:51`, `src/cli/dispatch.ts:45`). Share/derive alias metadata so dispatch and flag parsing cannot drift.
3. Parse-stage repeated-flag errors ignore requested JSON mode (`src/cli/main.ts:80`). Preserve `--json` for this C-LIST-07 failure path and test it.
4. C-LIST-06's composition test has no project-scope entry, so its `--scope project` result is vacuously empty (`tests/e2e/list-filters.test.ts:177`). Add a positive combined case.
5. Run the full typecheck/lint/coverage CI gate after rebasing/retargeting to a CI-triggering base.

stevekrenzel:
Round three addressed in `ba26d72`. All five reproduced first; nothing declined.

**Minor 1 — dangling citation** (`PRD.md`). Confirmed: §5.1 and C-LIST-07 both cite "the general flag-uniqueness rule in §5.2", and §5.2 listed flags without defining any such rule. Added it, worded to match the rule #117 adds to the same section, so the two collapse to one definition rather than two paraphrases when those branches meet.

**Minor 2 — duplicated alias metadata** (`src/cli/alias-registry.ts`, new). Confirmed: `COMMAND_ALIASES` had three entries and `FLAG_TABLE_ALIASES` one, with nothing linking them. Both now derive from one registry.

I kept the bare/prefixed distinction rather than collapsing the tables, because they are not the same thing and merging them would be a regression. A bare alias (`skills` → `list`) inherits its canonical command's flags. A prefixed one (`taps` → `tap list`) must not: `crew tap add` accepts `--recursive`, but `crew taps` resolves to `tap list`, which ignores it. Verified both still hold:

```
$ crew taps --recursive      → usage_error: Unknown argument: recursive
$ crew skills --tap core     → accepted
```

**Minor 3 — parse errors ignored `--json`** (`src/cli/main.ts`). Reproduced: `crew list --json --tap a --tap b` wrote human text to stderr and left stdout empty, so a script piping stdout got nothing. `writeError` was called with a hardcoded `false`.

The mode is now read off raw argv, which is the only source available when the failure is the parse itself. I ported `wantsJsonOutput` verbatim from #117 rather than writing a second one — same name, shape, and coercion rules — so they merge as a duplicate rather than a conflict. That includes matching yargs on space-separated values and on last-occurrence-wins, both covered by the new test.

Verified load-bearing: restoring the hardcoded `false` fails exactly the new test, nine others passing.

**Minor 4 — vacuous composition assertion** (`tests/e2e/list-filters.test.ts`). Confirmed: `seed` created only user-scope entries, so the `--scope project` arm asserted an empty result that would have been empty regardless — it would have passed with `--scope` ignored entirely. `seed` now installs a project-scope skill and returns its root, and the test asserts a positive three-filter result including `project_root`, plus the negative at the other scope.

Updating the fixture surfaced three sibling assertions that hardcoded the old two-row set; they now expect the third row, which is correct since an unscoped list shows both scopes.

**Minor 5 — CI gate.** I cannot retarget this stacked PR to a CI-triggering base, so I ran the gate locally: typecheck clean, `bun run lint` clean (2 pre-existing config infos, also on `main`), `cd site && bun run lint` clean (1 pre-existing info). Targeted suites: `list-filters` 10 pass, with `list-scope` and `aliases` 22 pass, 0 fail. `alias-registry.ts` at 100% line and function coverage.

Two notes. `bun run lint` initially failed on import ordering and formatting in my own changes — the same drift that turned #112's CI red — so I ran `lint:fix` before pushing. And this branch carries no copy of #107's `isSuppressed` dry-run fix; it inherits that centrally at merge.

Also derived the test's JSON shape from `StateEntry`/`Scope` instead of a hand-written literal, so a change to what `crew list --json` emits is a compile error rather than a test that quietly stops checking.

stevekrenzel:
Rebased onto the new #118 tip (now on `main`, which includes #111 and #116). Reconciled so exactly one copy of each mechanism remains — `main`'s:

- **Alias flag lookup.** Dropped this PR's `src/cli/alias-registry.ts` (`COMMAND_ALIASES` + `flagTableKeyFor`) and its `FLAG_TABLE_ALIASES` table; `parseArgs` and `dispatch` use `main`'s `src/cli/aliases.ts` (`aliasFlagKey` / `aliasFor`) unchanged. `crew skills --tap …` works through that mechanism (C-LIST-05 test unchanged).
- **Repeated-flag rejection.** Dropped this PR's `rejectRepeatedScalars` port; `main`'s version (which also catches repeated booleans from raw argv) is the only definition. The duplicate "Flag uniqueness" §5.2 paragraph was dropped in favor of `main`'s; C-LIST-07 keeps pointing at it.
- **`tests/e2e/aliases.test.ts`** is `main`'s file. This PR's `taps`/`untap --recursive` case moved into `tests/e2e/list-filters.test.ts` (C-LIST-07). The `crew ls` alias test's exact-shape check became `installations` only, since `list --json` now carries `scope`/`agent`/`tap` (byte-equality with `list` is still asserted).
- **Kept:** `STRING_SUB.list = ["tap"]` next to `main`'s `install: ["from-git"]`, and `wantsJsonOutput` (parse-stage errors honor `--json`). Because of the latter, `main`'s C-CLI-08a `install --json --json foo` case now asserts the structured error on stdout — the same adjustment #117 made to that test, so the two resolve identically at merge.

Kept `src/cli/args.ts` additions minimal (242 lines; `main` is already at 203 — #114/#117 restructure it). Targeted tests: aliases, flag-guards, list-filters, list-scope → 32 pass. Lint and typecheck clean; all commits signed.

stevekrenzel:
Rebased onto the new `fix/list-scope-filter` tip (`805b842`) after the cap-compliance refactors (#135–#143) merged. New tip: `d4bb01e`.

Re-placed, since `src/cli/args.ts` was split into `src/cli/args/` on `main`:

- **`STRING_SUB.list = ["tap"]` → `src/cli/args/tables.ts`** — the flag tables moved there, comment included.
- **`wantsJsonOutput` → new `src/cli/args/json-output.ts`** — rather than growing `index.ts`. It is a single-purpose helper that deliberately bypasses yargs (it has to answer for argv that made yargs throw), so it reads well on its own; `src/cli/main.ts` imports it from the new path.

One extra commit was needed to satisfy the cap gate that merged with #137:

- **`test(list): split list-filters into a directory to meet the 200-line cap`** — `tests/e2e/list-filters.test.ts` was 298 lines. Split along the seam the suite already had: `filters.test.ts` (C-LIST-04/05) and `composition.test.ts` (C-LIST-06/07), with the shared fixture in `helpers.ts`, matching `tests/e2e/search/`. Per CLAUDE.md, adapter redirection stays an explicit `beforeEach`/`afterEach` in each test file rather than hiding in the helper. All 11 test titles and bodies are unchanged.

`wc -l src/cli/args/`:

```
159 src/cli/args/index.ts
 41 src/cli/args/json-output.ts
 79 src/cli/args/tables.ts
```

`bun run check` is green: typecheck, lint (incl. the cap gate), 1128 pass / 0 fail, coverage 100.00 / 100.00. No site files in this PR.

PR #125, opened 2026-09-11
[LOG-5769] feat(refs): accept @ref after //subpath as well as before

stevekrenzel:
Stacked on #113 (`feat/browser-url-refs`); merge that first.

## What

`<git-url>//<subpath>@<ref>` now parses identically to `<git-url>@<ref>//<subpath>`, for every git-shaped form (https, ssh, `gh:`/`gl:`/`bb:`, `@owner/repo`, `file://`, browser URLs).

Before: `crew install @anthropics/skills//skills/internal-comms@main` silently folded `@main` into the subpath and failed with `no_skills_found`.

## Rules

- The ref-last value is the text after the last `@` in the subpath's final segment. An `@` in an earlier segment is part of the path.
- **Ref-first wins.** An explicit `@<ref>` before the subpath suppresses the trailing scan entirely, so a final segment containing a literal `@` is preserved verbatim — `gh:o/r@main//skills/foo@bar` is ref `main`, subpath `skills/foo@bar`. That form is the documented escape hatch, and it is also the behavior on the parent branch, which this feature must not regress. The two positions can never disagree, so there is no conflicting-ref error.
- Both positions accept the same ref text: §8.4 excludes whitespace and `:` from `git-ref`. A tail that fails those rules stays in the subpath rather than becoming a ref git cannot resolve.

## PRD

- §8.2: "A ref and a subpath may combine" shows both orders and states the ref-first precedence.
- §8.4 grammar: split into the two alternatives, so "ref first keeps the subpath verbatim" is expressed in the grammar rather than only in prose.
- §18.3: **C-REF-31** (ref-last equals ref-first across every form) and **C-REF-32** (ref-first preserves a literal `@` in the final segment), placed after the parent's C-REF-30.
- `crew help install` reference forms gains a "version last" example and a note on the `@`-in-folder-name case.

## Tests

- `tests/refs/ref-after-subpath.test.ts` (17 tests) — ref-first vs ref-last across all forms; single-segment subpath; `@` in a non-final segment; empty, whitespace, and colon tails; ref-first precedence over a trailing tail; both positions rejecting a colon identically.
- `tests/e2e/ref-after-subpath.test.ts` (1 test) — real CLI install of `file://<repo>//<sub>@<sha>` records `ref`, `pinned`, and `resolved_sha` identically to the ref-first form, with `--agent claude-code` so the recorded agent set does not vary with the host.

Internal: the tail-splitting helpers live in `src/refs/git-tails.ts`, shared with the parent branch's `hostEnd` work.

## Found while testing (not changed here)

Pinning to an older SHA does not install that SHA's content: `crew install file://<repo>@<old-sha>//demo` records `ref: <old-sha>` and `pinned: true` but `resolved_sha` and the installed bytes come from the clone's HEAD. Same for the ref-first form, so pre-existing and independent of this PR. Fixed separately in #129.


stevekrenzel:
## Review

Verdict: **not approved — 1 blocker, 1 major, 1 minor, 1 nit**

### Blocker

- **The E2E installs can write into real detected agents** (`tests/e2e/ref-after-subpath.test.ts:48`, `:56`). Both use the default agent set while only Claude Code is redirected. Any other detected adapter can receive `demo` in its real skills directory, and results become host-dependent. Add `--agent claude-code` to both installs.

### Major

- **The documented ref-first escape hatch for a literal final-segment `@` does not work** (`src/refs/git-url.ts:70`, `src/refs/git-tails.ts:44`). For example, `gh:acme/skills@main//skills/foo@bar` treats `bar` as a second ref and throws instead of resolving subpath `skills/foo@bar` at `main`; this also regresses the base behavior. Define an unambiguous literal-`@` escape or revise the grammar/contract, then test the documented case.

### Minor

- `src/commands/help/content/install.ts:112` says combined references require “version first,” while line 121 immediately documents version-last syntax. Say the two may appear in either order.

### Nit

- The comment at `tests/e2e/ref-after-subpath.test.ts:64` says only the auto-tap name differs, but the assertion also normalizes `installed_at`. Mention both dynamic source and timestamp fields.

All eleven review lenses ran against head `fa53e6e`. The normal CI check is absent from this PR’s status rollup. I did not approve because of the blocker and major finding.

stevekrenzel:
Addressed all four findings in `eab6d0d`, after rebasing onto the updated parent.

## Rebase and `git-tails.ts` reconciliation

The parent (#113) moved to `618a40d` and independently created `src/refs/git-tails.ts` with the same two helpers, plus `hostEnd` and slash-containing-ref support. Three conflicts: `PRD.md`, `src/refs/git-url.ts`, and an add/add on `git-tails.ts`.

Resolved by taking the parent's `splitSubpath`, `splitGitRef`, and `hostEnd` verbatim and re-applying only `splitTrailingRef` on top, per the note the parent left in that file. `origin/feat/browser-url-refs` is now an ancestor of this branch and GitHub reports MERGEABLE.

Both sides also added PRD rows numbered C-REF-28/29. The parent's three (28/29/30) are kept as-is; mine are renumbered **C-REF-31/32**.

## Major — the ref-first escape hatch, and a base regression

Reproduced before changing anything:

| Reference | Parent (`618a40d`) | This PR before the fix |
|---|---|---|
| `gh:acme/skills@main//skills/foo@bar` | ref `main`, subpath `skills/foo@bar` | **throws** `names two different refs` |

So this was not only a broken documented workaround — it regressed a reference that parsed fine on the parent.

I took **option (a)**: an explicit ref-first tail suppresses the trailing scan entirely. Rationale — it restores the parent's exact behavior for that form, it makes ref-first a genuine escape hatch as the PR body already claimed, and it removes the conflict case by construction rather than by error message. The two positions can no longer disagree, so the `invalid_ref` for conflicting refs is gone along with its test.

Code, grammar, and help now agree:

- `src/refs/git-url.ts` — trailing scan runs only when `leadingRef === null`.
- `PRD.md` §8.2 — states the precedence and shows both forms side by side; §8.4 grammar split into the two alternatives so "ref first keeps the subpath verbatim" is expressed in the grammar itself, not just prose.
- `src/refs/git-tails.ts` — `splitTrailingRef` documents that it is only consulted without a ref-first tail.

Verified the fix restores parent behavior exactly, and that ref-last still works. Both new C-REF-32 tests **fail against the pre-fix commit** (confirmed by reverting just the precedence logic and re-running: 13 pass, 2 fail).

## Blocker — host-dependent agent set

Both installs now pass `--agent claude-code`.

On severity: I did not confirm a real host write is reachable, and I don't believe it is. `tests/helpers/preload.ts` calls `neutralizeAdaptersExcept([])`, which points every adapter at an inert directory — a sibling worker established this on #124 and it was independently verified. The genuine defect is the **environment-dependent assertion**: the recorded agent list varied with whatever adapters the host detects. Same fix, narrower claim.

## Minor — contradictory help text

`src/commands/help/content/install.ts` said combined references require "version first" immediately above an example showing version-last. Now: "with the version either before or after the subfolder". Added a note documenting the `@`-in-folder-name case, since the PRD now specifies it.

## Nit — comment accuracy

The e2e comment now names both normalized fields: the derived auto-tap name in `source` and the `installed_at` timestamp.

---

`bun run check`: **1068 pass, 0 fail**, 100% line and function coverage, tsc and Biome clean. All touched files under the 200-line cap.

stevekrenzel:
## Updated /review

**Verdict:** 1 blocker, 1 major, 1 minor. Not approved.

### Blocker

- PR description — The Rules/Tests sections still say conflicting ref values produce `invalid_ref` and that the tests cover equal/conflicting pairs, while the final code and PRD make ref-first win. It also cites C-REF-28/29 instead of the committed C-REF-31/32. Update the description to match the implemented precedence and actual criteria/tests.

### Major

- `src/refs/git-tails.ts:75` — Ref-last parsing accepts `:`, although PRD §8.4 excludes colons from `git-ref` and the same suffix is legal subpath text. For example, `gh:o/r//skill@release:bad` becomes an invalid Git ref and later reports `ref_not_found` instead of preserving the literal subpath. Reject `:` alongside whitespace in `splitTrailingRef` and add coverage.

### Minor

- `tests/e2e/ref-after-subpath.test.ts:2` — The file docstring labels the suite C-REF-28, while its test and the PRD identify C-REF-31. Correct the criterion reference for conformance traceability.

stevekrenzel:
Addressed all three findings in `0da850c`, after rebasing onto the updated parent. Nothing declined.

## Rebase

One conflict, in `src/refs/git-tails.ts` — and only in the file docstring. The parent (`3123e28`) added a paragraph explaining why `hostEnd` exists; my commit's version of the docstring predated it. Kept the parent's paragraph. No code conflicted, so the parent's credential-redaction work and this branch's ref-last parsing coexist untouched. `origin/feat/browser-url-refs` is an ancestor of the new head and GitHub reports MERGEABLE with the base unchanged.

## Major — the colon gap, and what the leading form actually does

Reproduced before changing anything. The two positions disagreed:

| Reference | ref | subpath |
|---|---|---|
| `gh:o/r//skill@release:bad` (ref-last) | **`"release:bad"`** | `"skill"` |
| `gh:o/r@release:bad//skill` (ref-first) | `null` | `"skill"` |

**The leading form does not share the gap** — it was already correct. `splitGitRef` has tested `/[\s:]/` since the parent branch; only `splitTrailingRef` was checking whitespace alone. So this was an inconsistency introduced by the ref-last feature, not a pre-existing hole in both paths.

That matters for the user-visible behavior: PRD §8.4 defines `git-ref` as "any non-empty string not containing `:` or whitespace", so `release:bad` was never a legal ref. Reading it as one produced a ref git cannot resolve, which surfaces later as `ref_not_found` — when `skill@release:bad` is perfectly legal subpath text that should simply have stayed in the subpath.

`splitTrailingRef` now applies the same character class as `splitGitRef`. After the fix:

```
gh:o/r//skill@release:bad   => ref: null   | subpath: "skill@release:bad"
gh:o/r@release:bad//skill   => ref: null   | subpath: "skill"
gh:o/r//skill@release       => ref: "release" | subpath: "skill"
gh:o/r@release//skill       => ref: "release" | subpath: "skill"
```

No PRD change needed — §8.4 already excluded colons; this makes the code match the grammar it already had.

Two new tests, one asserting the subpath is preserved and one asserting both positions agree. **Both fail against the pre-fix commit** (verified by stashing, restoring only the tests, and re-running: 15 pass, 2 fail — exactly the two new ones).

## Blocker — description drift

Updated. The Rules section now states ref-first precedence rather than the conflicting-ref `invalid_ref` that round one removed, cites **C-REF-31/32** as committed, and lists the actual test inventory (17 unit + 1 e2e, including the `--agent claude-code` pinning from round one). The "found while testing" note now points at #129, which fixed the pinned-SHA bug it describes.

## Minor — stale criterion ID

`tests/e2e/ref-after-subpath.test.ts` line 2 said C-REF-28 while its own test and the PRD say C-REF-31. Corrected. I swept the rest of the file and the PR's other files: the remaining C-REF-28/29/30 references all belong to the parent's browser-URL criteria and are correct where they appear.

## Verification

Targeted only, given the known non-determinism on branches without #131:

- `tests/refs/ref-after-subpath.test.ts` + `tests/e2e/ref-after-subpath.test.ts` — 18 pass, 0 fail
- `tests/refs/` in full — 125 pass, 0 fail (run because `splitGitRef` is shared with the parent's work)
- `src/refs/git-tails.ts` at 100% line and function coverage
- tsc and Biome clean

One typecheck fix along the way: my first version of the agreement test read `.ref` directly off `parseRef`'s return, which is a `Source` union where `PathSource` has no `ref`. Switched to `toMatchObject` with `type: "git"` rather than casting, so the test still fails if the form stops parsing as a git source.

No loops changed — the fix adds a character to an existing bounded scan — so there was nothing new to check for termination.

stevekrenzel:
Rebased onto the revised parent `c5baaae`. Branch is now `4caa84a`.

**One conflict, in `src/refs/git-url.ts`, and only in the imports.** The parent added `displayUrl` for credential redaction in the `invalid_ref` path; this branch added `splitTrailingRef` for ref-last parsing. Both are needed, so both imports are kept. The function bodies merged cleanly — `parseGit` now redacts the echoed reference *and* accepts a ref on either side of the subpath.

Worth noting the two changes compose rather than collide: the parent's redaction wraps the reference echoed in the error, while this branch decides where the ref boundary falls. A wrong-shaped authenticated URL with a trailing ref gets both behaviours.

**Verified:** parent is an ancestor; `tests/refs/` + `ref-after-subpath` e2e 132 pass; the parent's `tap-credentials` + `browser-urls` suites 15 pass; 0 fail. Typecheck and Biome clean (only the two pre-existing config infos).

stevekrenzel:

**Verdict:** 0 blockers, 1 major, 0 minors, 0 nits. **Not approved.**

Reviewed `c5baaae4..4caa84aa` through all 11 `/review` lenses. This stacked PR has no test-suite CI run at this head.

### Major

1. **The ref-last grammar accepts refs the implementation cannot parse** — `PRD.md:748`
   The production reuses `git-ref`, which permits `/`, so `gh:o/r//skills@feature/foo` is valid by the grammar with ref `feature/foo`. `splitTrailingRef` only examines the final path segment and instead leaves it as subpath `skills@feature/foo` with no ref. Define a slash-free ref-last production and document the asymmetry, or adopt syntax/implementation that can delimit slash-containing ref-last values; add a conformance test.

No other release-holding or actionable findings were found.

stevekrenzel:
Addressed in `e91057f`.

## Major: the grammar promises a slash-containing ref-last ref the parser never delivers

**Real, and reproduced first.** The ref-last production was written `[ "@" git-ref ]`, and `git-ref` is explicitly documented as "may contain `/`". But `splitTrailingRef` only scans after the last `/`, so it structurally cannot return a slash-containing ref:

```
gh:o/r@feature/foo//skills   ->  ref "feature/foo"  subpath "skills"
gh:o/r//skills@feature/foo   ->  ref null           subpath "skills@feature/foo"
gh:o/r//skills@v1            ->  ref "v1"           subpath "skills"
```

You offered two fixes and preferred the grammar side. I agree, and I want to be explicit that this is not just the cheaper option — **widening the parser is not actually available.**

A subpath may legitimately contain `@` (that is already pinned by C-REF-31's `scoped@pkg/py` case and by C-REF-32). So in `//a@b/c` there is nothing to distinguish a slash-containing ref `b/c` from a directory literally named `a@b`. Making the parser accept both readings would not make the grammar more permissive, it would make it **ambiguous** — a strictly worse defect than the one being fixed. The ref-first form already exists as the unambiguous spelling, so the expressiveness is not lost, only relocated.

So the ref-last production now takes its own terminal:

```
trailing-ref := any non-empty string not containing "/", ":" or whitespace
```

with a paragraph in §8.4 stating the asymmetry, why it exists, and that `gh:o/r@feature/foo//skills` is the spelling for this case.

### One sub-decision worth surfacing

`gh:o/r//skills@feature/foo` is **subpath `skills@feature/foo` with no ref, not an `invalid_ref`.** I considered erroring, since a user typing that probably meant a ref. I did not, because that string is a perfectly legal subpath, and erroring would make crew reject a directory a user may actually have on disk. Silently-wrong-looking beats impossible-to-express. The PRD says this explicitly so the next reader does not "fix" it into an error.

### Conformance

- **C-REF-31 qualified** to "For a slash-free ref, ...". The interchangeability claim was always only tested against slash-free refs; it now says so rather than overclaiming.
- **C-REF-33 added**, pinning both halves of the asymmetry and the not-an-error part. The test asserts the ref-first and ref-last readings separately rather than via the `FORMS` equality table, because the whole point is that they are *not* equal here.

No `src/` change — the parser was already correct, the spec was not. That is the defect you identified: a grammar promising more than the parser delivers.

### Verification

- `bun test tests/refs/` — 132 pass, 0 fail; `git-tails.ts`, `browser-url.ts`, `parsers.ts` all at 100% line coverage.
- `bun run typecheck` clean.
- `bun run lint` clean (2 pre-existing biome-config infos).
- Site lint clean (1 pre-existing info).
- No merge-conflict markers anywhere in the tree.
- Commit is signed and verified (`verified=true reason=valid`).

stevekrenzel:
Rebased onto the new #113 tip; import merge only (`splitTrailingRef` alongside `normalizeSubpath`). Refs tests pass (136).

stevekrenzel:
Rebased onto the new #113 tip (which is now on `main` after the cap-compliance refactors #135–#143). No conflicts; no changes beyond the replay.

Verification: `bun run check` — typecheck, lint (incl. the cap gate), 1197 tests, 100% line/function coverage; targeted refs/browser-URL suites 145 pass. Touches nothing under `site/`.

PR #126, opened 2026-09-11
[LOG-5771] feat(uninstall): accept tap and namespace selectors, add --all

stevekrenzel:
Stacked on #110 (`fix/uninstall-scope`). Review that one first; this PR targets it as its base.

## What and why

`crew install <tap>` installs a whole tap and `crew install <namespace>` installs a whole bundle, but there was no way to undo either in one step — `crew uninstall <tap>` failed with `not_installed_here`. This closes that asymmetry and adds the "start over" escape hatch.

**Collection selectors.** Resolution order per argument, mirroring `crew install <tap>` (§16.4):

1. An installed skill (bare or tap-qualified) — today's meaning, always wins.
2. A configured tap (registered or auto) — every entry attributed to it at the target scope.
3. A namespace — `<tap>/<ns>`, or a bare `<ns>` when exactly one tap has installed entries under it.
4. Otherwise `not_installed_here`, as before.

A word that is both a tap name and a namespace with installed entries elsewhere, or a namespace installed from more than one tap, raises `ambiguous_reference` listing each qualified form. A configured tap with nothing installed from it reports that and exits 0 rather than failing — asking to remove skills from a tap you have none of isn't an error.

A collection expands to one removal per distinct skill name, each running the normal uninstall algorithm, so `--agent`, `--prune`, and `--scope` compose unchanged. `--json` records carry `collection: { kind, name }`.

**`--all`.** Removes every skill at the target scope. It takes no positional selectors (combining is a usage error) and requires confirmation: `--yes`, or an interactive `[y/N]`. A non-TTY run without `--yes` aborts with a usage error naming `--yes` rather than silently wiping the scope. Nothing installed at that scope is `not_installed_here`.

Per §14, confirmation happens **before** the state lock is taken, and state is re-read under the lock afterwards — the lock must not span a human decision, and the removal should act on state as it is after the answer rather than the snapshot the count came from.

## Note for the merge: shared module with #120

#120 (`feat/update-by-collection`) also touches `src/state/collections.ts`. That branch isn't in this one's base, so the module exists here independently and **the two must reconcile to a single copy at merge time, not two competing APIs.** Four additions must all survive:

- this PR: a trailing `command` parameter (picks the verb in the ambiguity error's copy-pasteable suggestions) and a `collectionState` parameter (so a scope-narrowed state governs tap/namespace membership while skill resolution still sees full state);
- #120: `entryIdentity` and `refreshCollectionSubjects`.

## PRD

- §5.1 — `crew uninstall --all` in the command list.
- §5.3.1 — `--all` semantics, including the confirmation rule.
- §7.4 — "Collection selectors" subsection with the resolution order.
- §18.3 — C-UNINST-19..24, inserted after C-UNINST-18 inside the uninstall block.

## How tested

`tests/e2e/uninstall-collections/` — 22 e2e tests through `runCli` against real `file://` git taps, split by concern:

- `resolution.test.ts` (7) — whole-tap removal leaves other taps alone; namespace removal qualified and bare; skill name wins over a same-named tap; both ambiguity shapes; empty tap reports and exits 0.
- `composition.test.ts` (7) — `--agent`, `--prune`, `--scope`, and `--json` composition, including that a skill selector's record carries no `collection`.
- `all.test.ts` (8) — the `--all` matrix (`--yes`, prompt yes, prompt no, non-TTY abort, with-a-selector, nothing-installed, `--scope project`), plus a test that installs a skill *from inside the confirmation prompt* to prove the lock is not held across the prompt and state is re-read after it.

Plus `tests/e2e/uninstall-abort.test.ts` (3, inherited from the base) covering retained ownership and prune termination when a removal aborts.

Targeted suites green; typecheck and Biome clean. No full-suite number is quoted: the suite is non-deterministic on any branch lacking #131 (the default `core` tap URL is a real remote, so tests that don't discard it attempt live clones). I reproduced a failure on the **unmodified base** to confirm the pattern predates this branch.

## Judgment calls

- `removeOne` works on a single skill name, so collections expand in the selection layer (`src/commands/uninstall/select.ts`) rather than teaching the removal primitive about multi-name subjects. Keeps the blast radius off the core path #110 just reworked.
- A collection selector that matches nothing at the target scope narrows to an empty set silently; only skill selectors carry the "it's installed over here" remedy from #110, since that message doesn't make sense for a tap.
- Entry keys are JSON-encoded tuples rather than delimiter-joined strings, so no separator appearing inside a component can make two distinct entries share a key.
- Site `Commands.data.tsx` enumerates uninstall's selector forms, so it was misleadingly incomplete and got updated along with the README row.


stevekrenzel:
## Review

Verdict: **not approved — 1 blocker, 4 majors, 2 minors, 1 nit**

### Blocker

- **The new E2E file violates the hard file-size cap** (`tests/e2e/uninstall-collections.test.ts:1`). It is 283 lines; AGENTS.md requires every file to remain under 200. Split collection and `--all` suites and move shared setup into a focused helper.

### Majors

- **Overlapping selectors produce duplicate stale targets** (`src/commands/uninstall/select.ts:47`, `src/commands/uninstall/index.ts:60`). `crew uninstall acme alpha`, when `alpha` belongs to tap `acme`, removes it once and then tries its now-missing directory again, records a failure, and exits 1. Deduplicate by scoped state-entry identity or resolve each selector against current state.
- **`--all` holds the state lock while waiting for interactive confirmation** (`src/commands/uninstall/index.ts:56`). A paused prompt blocks every other mutation until it times out with `state_locked`. Confirm first, then take the lock, reread state, and derive targets.
- **Collection resolution does not preserve established scope semantics** (`src/commands/uninstall/select.ts:58`, `:125`; `src/state/collections.ts:46`). It omits the lone-project fallback and includes other scopes when deciding namespace ambiguity. Apply scope narrowing per grouped skill before collection ambiguity/expansion, with cross-cwd and cross-scope tests.
- **Large collection removals do quadratic state work under the lock** (`src/commands/uninstall/index.ts:60`, `src/commands/uninstall/state.ts:43`). Each skill removal performs another full `filter → map` pass. Batch filesystem outcomes, then apply removals and `required_by` cleanup in one state pass.

### Minors

- `src/commands/help/content/uninstall.ts:5`, `README.md:167`, and `site/components/sections/Commands.data.tsx:20` still present positionals as `<name>` although the PRD now defines `<selector>`. Update the syntax and explain explicit and uniquely resolved namespaces.
- `tests/e2e/uninstall-collections.test.ts:85` does not assert the new JSON `collection: { kind, name }` contract, and `--prune` composition is untested. Add E2E coverage for both.

### Nit

- `allowEmpty` in `src/commands/uninstall/select.ts:29` can be set without a collection, allowing an ordinary missing selector to succeed silently. Use a discriminated target type so only collection variants can enable it.

All eleven review lenses ran against head `6e821c8`. The normal CI check is absent from this PR’s status rollup. I did not approve because of the blocker and major findings.

stevekrenzel:
Addressed all eight findings in `e2eed12`, rebased onto the updated `fix/uninstall-scope`. Nothing pushed back on — every finding reproduced first.

## Rebase

Two conflicts, both in the region this PR and #110 each rewrote. `UninstallRecord` gained `collection` from here and `RemovalMeta` split out on the parent side; I kept both, with `projectRoots` staying out of the serialized record as the parent intended. `removeOne`'s signature took the parent's `meta` return plus this PR's empty-set parameter. `origin/fix/uninstall-scope` is an ancestor of the new head and the PR is MERGEABLE.

## Blocker — test file over the cap

283 lines, confirmed. Now `tests/e2e/uninstall-collections/`: `resolution.test.ts` (122), `composition.test.ts` (129), `all.test.ts` (114), `helpers.ts` (87). The split is by concern rather than arbitrary halves — what a selector resolves to, versus how it composes with flags and scope.

## Major 1 — overlapping selectors

Reproduced: `crew uninstall acme alpha`, where `alpha` is in tap `acme`, removed both skills and then **exited 1** with `wasn't installed here`, having retried `alpha`'s already-deleted directory. A command that did exactly what was asked reported failure.

`dedupe()` in `select.ts` now drops entries a previous target already covers, keyed on `(name, scope, project_root)` per §11.1. A collection left fully covered contributes nothing; an empty *collection* selector still reports "nothing installed", since that is a different statement.

## Major 2 — lock held across the prompt

Confirmed by inspection: `withStateLock` opened in `index.ts:56` and `ctx.prompt` fired inside it from `select.ts:107`. A user who walked away from `--all` blocked every other crew mutation for the full 30s lock timeout.

Confirmation now runs against an unlocked read, before the lock. The lock then re-reads state, so the removal acts on installs as they are when it runs rather than as they were at the prompt — that re-read matters, and it is why I did not simply hoist the prompt and reuse the earlier snapshot. Recorded in PRD §5.3.1 as a §14 guarantee.

## Major 3 — scope semantics dropped

Reproduced, and the inconsistency was visible from one fixture: with a single project-scope install and `--scope project` from an unrelated cwd, `crew uninstall alpha` (skill) correctly fell back to the lone project entry and removed it, while `crew uninstall acme` (tap) printed `Nothing installed from tap acme` and **silently removed nothing at exit 0**.

Cause was a second, stricter `entriesAtScope` in `select.ts` requiring `project_root === cwd`, shadowing the parent's version with its lone-project fallback. The parent's is now exported and shared, so collection selectors narrow exactly as skill selectors do.

Ambiguity had the same root: `namespaceCandidates` scanned every scope, so a namespace present in two taps at *different* scopes was reported ambiguous even when only one was reachable. `resolveCollectionSubject` takes a separate `collectionState` for membership — the full state still backs skill resolution, so a cross-scope skill keeps #110's "it's installed over here" remedy.

## Major 4 — quadratic state work

Measured before and after, removing a whole tap of N skills:

| N | entry visits before | after |
|---|---|---|
| 50 | 1,275 | 50 |
| 100 | 5,050 | 100 |
| 200 | 20,100 | 200 |

`dropScopedEntriesAndUpdateRequiredBy` indexes targets by location, then makes one pass. Same-location scrubbing is preserved exactly — that is #110's data-loss fix and it stays intact. The single-entry wrapper had no remaining callers, so it is deleted rather than kept as dead code.

## Minors

**`<name>` → `<selector>`** in `help/content/uninstall.ts`, the README row, and the site command table, each now explaining that a namespace is written qualified (`acme/marketing`) or bare when unique across your taps.

**Missing coverage** added: the `--json` `collection: { kind, name }` contract, its absence on skill selectors, and `--prune` composing with a collection (a path install's auto tap, so `--prune` sweeps the dependency the collection freed).

## Nit — `allowEmpty`

`UninstallTarget` is now a discriminated union: `{ kind: "skill" }` or `{ kind: "collection", collection }`. Only the collection variant tolerates an empty entry set, so an ordinary missing selector can no longer succeed silently — the shape makes it unrepresentable rather than relying on a flag being set correctly.

## Pre-fix verification

Reverted `src/` only, keeping the new tests: **3 of 5 fail** — both C-UNINST-26 scope cases and C-UNINST-25 dedupe. The `--json` and `--prune` tests pass either way; they close coverage gaps rather than catching live bugs, and I would rather say so than let "5 new tests" imply five caught regressions.

## For the merge — shared module with #120

#120 added `refreshCollectionSubjects` to `src/state/collections.ts`, the module this PR created independently. My Major 3 fix **does change that module's shape**: `resolveCollectionSubject` and `resolveCollectionSubjects` take a trailing optional `collectionState` after `command`. Both are optional and default to `state`, so #120's call sites compile unchanged — but whoever merges second should keep the parameterized `command` argument *and* the new `collectionState`, and pass `"update"` from the update command.

## Gate

`bun run check` exits 0: 1020 pass, 0 fail, 100% line and function coverage. Site lint clean. Every touched file is under the 200-line cap.

One note while splitting the helpers: an `addTap` guard that threw on a non-zero exit was unreachable, and once it lived in a shared module it showed up as a coverage gap in whichever suite did not trip it. It now returns the exit code and callers assert on it, which is both coverable and a stronger assertion.

stevekrenzel:
## Updated /review

**Verdict:** 3 blockers, 1 major, 3 minors, 1 nit. Not approved.

### Blockers

- `src/commands/uninstall/select.ts:139` — The file contains two literal NUL bytes, so Git classifies the TypeScript source as binary and GitHub hides the implementation diff. Replace them with textual `\u0000` escapes or a text-safe structured key.
- `src/state/collections.ts:19-21` — The module docstring says both uninstall and update consume it, but this frozen head has no update caller and the PR itself says #120 has not merged. Describe its current use accurately.
- PR description — “How tested” points to deleted `tests/e2e/uninstall-collections.test.ts` and says 15 tests; the head has three split suites with 21 tests. Update the review evidence.

### Major

- `src/commands/uninstall/core.ts:91-103` — A filesystem uninstall failure is recorded but still treated as a full removal: state is dropped, dependencies may be pruned, and an auto tap may be garbage-collected while the skill remains on disk. Retain failed agents/entries and exclude them from full-removal/prune bookkeeping; mutate state only for successful removals or accepted forced absence.

### Minors

- `src/state/collections.ts:33`, `src/commands/uninstall/select.ts:43`, `src/commands/uninstall/core.ts:29` — `CollectionKind` includes `skill`, making an invalid `collection: { kind: "skill" }` output representable. Split resolver subject kinds from tap/namespace collection kinds.
- `tests/e2e/uninstall-collections/all.test.ts:41` — The static prompt stub does not prove confirmation happens before locking and that state is re-read afterward. Have the prompt callback acquire the lock and add an installation, then assert `--all` removes it too.
- `src/commands/uninstall/index.ts:67` — A K-skill collection runs K full state-array rebuilds under the lock, producing O(K·N) work. Accumulate removals/replacements and apply them in one state pass.

### Nit

- `src/state/collections.ts:42` — The docstring says `skillState`, but the parameter is named `state`.

stevekrenzel:
All eight findings addressed in `0970bee`, on top of a rebase onto the revised base. Nothing declined.

## Rebase

Two conflicts. `index.ts` was imports-only both times (the base added `narrowSubjectToScope`/`entryKey`, this branch added the selection helpers — kept both; scope narrowing now lives in `select.ts`, so `index.ts` no longer imports it). `state.ts` conflicted as **binary**, which is Blocker 1 showing up during the merge itself: git could not diff the file, so it offered whole-side resolution only.

I took this branch's batched `dropScopedEntriesAndUpdateRequiredBy` (it already incorporates the base's same-location scoping) and re-applied the base's `attempted` parameter on `findOrphan`. **That parameter is the prune-loop termination guard, and taking either side wholesale would have silently dropped it** — typecheck caught it. `origin/fix/uninstall-scope` is an ancestor; PR is MERGEABLE with the base unchanged.

## Blocker 1 — NUL bytes

Confirmed, and **it is two files, not one**. `select.ts` had them, and the base branch's own round-two `entryKey` in `state.ts` introduced a second instance — so the fix that added it propagated the bug. A scan of all 393 tracked text files found exactly these two.

Both now JSON-encode the tuple:

```ts
export function entryKey(e: StateEntry): string {
  return JSON.stringify([e.name, e.scope, e.project_root ?? ""]);
}
```

I chose JSON over a printable delimiter deliberately. The sibling helper in `src/commands/update/selection.ts` uses `::`, which is printable but still collidable — a `project_root` containing `::` can alias a different tuple. JSON-encoding cannot.

`select.ts` now shows a **170-line text diff** against the base instead of `Bin`. `state.ts` will still render as binary in this PR's diff **until #110 merges**, because the other side of the comparison is the base branch, which still carries its 2 NUL bytes. That is worth fixing on #110 directly — flagged below.

## Blocker 2 — `collections.ts` docstring

Correct: the module claimed `crew uninstall` and `crew update` both consume it, but this head has no update caller. Now states that uninstall is the only caller today and explains that the `command` parameter exists so update can reuse it with its own verb once its selectors land.

## Blocker 3 — PR description

Rewritten. It pointed at the deleted single-file suite and claimed 15 tests; the head has three suites totalling **22**. The description now lists them by concern with per-file counts, documents the §14 confirm-before-lock ordering, and states all four additions that must survive the `collections.ts` reconciliation with #120 (this PR's `command` and `collectionState`; #120's `entryIdentity` and `refreshCollectionSubjects`).

## Major — failed removals treated as full removals

The rebase resolves this, and I verified rather than assumed it. The base's mechanism is cause-agnostic: `removeFromAgents` catches **every** throw from `uninstallSkillFromAgents` into `rec.failures`, and `removeOne` retains ownership for anything recorded there. A filesystem failure reaches it the same way a safety abort does — `rmrf` uses `rmSync(..., { force: true })`, which suppresses missing paths but still throws on EACCES/EPERM. So state is retained, `meta.fullyRemovedRoots` stays empty, prune is not seeded, and the auto tap is not collected.

I tried to add a direct EACCES regression by making the install root read-only. **That test was a mistake and I removed it**: the helper's cleanup cannot delete a read-only directory, so the run hung with four test processes live and no output — the same failure mode as the earlier runaway incident. I killed it immediately rather than let it grow. The behaviour is covered by the base's C-UNINST-13a/13b tests, which exercise the identical retention path.

## Prune termination

Re-read before running anything against it, as asked. `pruneOrphans` adds each candidate's key to `attempted` **before** removing it, and `findOrphan` skips those keys, so the loop runs at most once per state entry and cannot depend on an entry disappearing — which is exactly what a retained aborted entry does not do. The base's three abort tests pass (3/3), including the one that protects the orphan itself.

## Minors

**`CollectionKind` included `skill`** — split into `SubjectKind` (`"skill" | CollectionKind`) for what a resolver argument named, and `CollectionKind` (`"tap" | "namespace"`) for what a collection expanded to, so `collection: { kind: "skill" }` is no longer representable.

**`--all` prompt stub** — added `C-UNINST-24 --all confirms before locking and re-reads state after`. The prompt callback installs a new skill, then answers yes; the run must sweep that skill too. It proves both halves at once: the lock is not held across the prompt (the nested install would block on it) and state is re-read after the answer. **Verified load-bearing** — moving confirmation inside the lock makes it fail at 31.3s, the lock timeout, versus 7s passing.

**O(K·N) state rebuilds** — already fixed by the batched helper this branch carries; the rebase preserved it. One pass over `installations` plus one over the removal batch, rather than a full rebuild per skill. At 200 installed skills and a 50-skill collection that is 10,000 entry visits versus 250, a 40× reduction, all of it under the state lock.

## Nit

`resolveCollectionSubject`'s docstring named `skillState` for a parameter called `state`. Corrected.

## For #110 specifically

The base branch still has 2 NUL bytes in `src/commands/uninstall/state.ts`, introduced by its own round-two `entryKey`. It has the same consequence there — git treats that file as binary and hides its diff from review. Worth fixing on #110 directly rather than waiting for this branch to merge over it.

## Verification

Targeted suites only, per the resource constraint. `uninstall-abort` 3/3; `uninstall-collections/` 22 tests, with `all.test.ts` 8/8 and `resolution.test.ts` 7/7 in isolation. Typecheck and Biome clean.

Directory-level runs failed 1–3 tests with a **different set each time**, always at a 5s+ timeout. I confirmed this predates the branch by stashing every change and running the same directory on the unmodified base twice: 21/21, then 20/21 with `C-UNINST-24 --all prompts and proceeds on yes` failing at 5.9s. That is the known non-determinism #131 removes, not this change.

stevekrenzel:
Rebased onto the revised parent `fix/uninstall-scope` (`bf12ae6`). Two conflicts, both in `src/commands/uninstall/state.ts`.

**The first one mattered.** It was reported as a *binary* conflict — an ancestor blob still contained the NUL bytes — so only whole-side resolution was offered. Both sides had independently fixed that defect, but the two versions differ in more than the key encoding: the parent added `findOrphan`'s `attempted` parameter (the guard that bounds the prune sweep once an aborted removal keeps its state entry), while this branch batched `dropScopedEntriesAndUpdateRequiredBy` into a single pass. Taking either side wholesale would have silently dropped the other's work — and taking this branch's side would have removed the termination guard.

Resolved by hand: kept this branch's batched removal and re-applied the parent's `attempted` parameter, docstring, and predicate on top. The second conflict was a duplicate explanatory comment from this branch's own NUL fix; kept it.

**Verified explicitly:**
- `pruneOrphans` still records each candidate in `attempted` *before* removing it, so every entry is visited at most once regardless of outcome — termination does not depend on the entry disappearing.
- Zero NUL bytes across all 17 changed files.
- Parent is an ancestor; typecheck and Biome clean; `uninstall-abort` + `uninstall-scope` 7 pass / 0 fail.

The `uninstall-collections/` suites show timeout-shaped failures (7–15s, varying set). I checked rather than assuming: the same suites fail the same way on the pre-rebase branch content, so this is the known non-determinism on branches without #131, not a rebase regression.

stevekrenzel:
Rebased onto the revised parent `20ad1c9`. Branch is now `0008195`.

**Five conflicts across two commits.** The parent extracted a canonical `src/state/identity.ts`, renamed `dropScopedEntryAndUpdateRequiredBy`, and removed `RemovalMeta.aborted`; this branch had batched the same state mutation and carried its own key encoding.

| File | Resolution |
|---|---|
| `uninstall/state.ts` | Kept this branch's batched `dropScopedEntriesAndUpdateRequiredBy` (the parent's single-entry version is quadratic for a K-skill collection), dropped its locally redefined `entryKey` in favour of the canonical import |
| `uninstall/core.ts` | Batched call site; both function names had auto-merged in side by side |
| `uninstall/index.ts` (×2) | Parent's canonical `entryKey` import + this branch's selection helpers |
| `README.md`, `site/…/Commands.data.tsx` | Both sides described true but different behaviour (scope targeting vs collection selectors) — merged into one row stating both |

**The three verifications you asked for:**

1. **Prune loop guard intact.** `pruneOrphans` still records each candidate in `attempted` *before* removing it, and `findOrphan` still takes that set. The loop runs at most once per entry regardless of outcome, so it cannot spin on an aborted orphan that deliberately keeps its state entry.
2. **One canonical key helper.** `grep -rn "function entryKey" src/` returns exactly one hit, `src/state/identity.ts`. `select.ts` had a byte-identical copy — the very function that originally encoded the triple with raw NUL bytes — and now imports the canonical one.
3. **Zero NUL bytes.** All 17 changed files check clean, and `git diff --numstat` reports no binary files, so every diff is reviewable.

**One follow-on the compiler caught:** `index.ts` kept a `narrowSubjectToScope` import that the collection work moved into `select.ts`. The narrowing still happens there — the parent's scope behaviour is preserved, only the import location was stale.

**Verified:** parent is an ancestor; `uninstall-collections/` 22 pass; the parent's `uninstall-prune` + `uninstall-abort` + `uninstall-scope` + `uninstall-scope-miss` 30 pass; 0 fail. Typecheck and Biome clean.

stevekrenzel:

**Verdict:** 1 blocker, 1 major, 5 minors, 0 nits. **Not approved.**

Reviewed `20ad1c91..00081951` through all 11 `/review` lenses.

### Blocker

1. **Unresolved merge-conflict markers are committed** — `README.md:167`, `site/components/sections/Commands.data.tsx:20`
   The README renders raw conflict text, and the TSX file cannot parse/build. Resolve both hunks to one current uninstall description.

### Major

1. **Collection project-scope fallback is computed against unrelated installs** — `src/commands/uninstall/select.ts:61`
   Collection scoping calls `entriesAtScope` over all installations before resolving the selected tap/namespace. An unrelated second project install disables the documented lone-project fallback, so `uninstall --scope project acme` can exit successfully without removing Acme while the equivalent skill selector works. Scope membership per selected collection and add a multi-project regression.

### Minors

1. `UninstallRecord.collection.kind` permits impossible `"skill"`; use `CollectionKind` (`src/commands/uninstall/core.ts:29`).
2. PRD §5.3.1 calls `--all` the only mass-removal shape despite tap/namespace selectors.
3. Add the same-named namespace half of C-UNINST-21 (`tests/e2e/uninstall-collections/resolution.test.ts:64`).
4. Collection/`--all` state updates rebuild the full state once per skill, producing O(N²) work under the lock. Batch mutations after filesystem work.
5. The unlocked pre-prompt `--all` read can observe a concurrent truncate/write as empty and fail spuriously. Avoid treating that snapshot as authoritative or make state writes atomic.

stevekrenzel:
Round 3 addressed in `dbe6949`. Details below.

## Blocker: committed conflict markers

`README.md` and `site/components/sections/Commands.data.tsx` both carried conflict markers. The TSX one was the worse of the two: the file could not parse, so site lint reported 72 errors.

I resolved both by merging each side's meaning rather than picking a winner. The uninstall row now describes the selector forms (skill, tap, namespace) *and* the scope targeting, because those came from different sides and both are true. Site lint is back to 0 errors.

`git grep -nE "^(<<<<<<<|=======|>>>>>>>)"` against the pushed head returns nothing.

## Major: a collection selector could silently remove nothing

Reproduced, and the symptom is worse than the report suggested. It is not exit 6. It exits **0 having removed nothing**, which reads as success.

The cause is which state answers which question. §7.4's lone-project fallback asks whether the *selected collection* has exactly one project install. The code was asking whether the *machine* does. So an unrelated project install of a different tap, sitting in a directory you have never visited, was enough to make `crew uninstall --scope project acme` quietly no-op from an unrelated cwd.

The fix separates the two questions in `selectedTargets`:

- **Which** collection a bare name means is still decided against scope-narrowed state. That preserves the existing behavior where a namespace colliding only with an out-of-scope install is not ambiguous.
- **What** that collection contains is re-read from full state and then narrowed, via a new `entriesForCollection` helper, so the lone-project fallback applies to that collection alone.

I confirmed the fix is load-bearing by reverting it temporarily. Before: `exit 0`, `acme-one removed: false`. After: `exit 0`, `acme-one removed: true`, and the unrelated install preserved. User-scope selection still leaves the project entry alone.

Regression test added: `C-UNINST-26 an unrelated project install does not hide a lone collection install`.

Worth noting my first attempt at this was wrong. Passing full state to `resolveCollectionSubjects` fixed the reported bug but broke `C-UNINST-26 a namespace is not ambiguous when the collision is out of scope`. That failure is what surfaced the two-questions distinction above. Both tests now pass.

## Minors

**`UninstallRecord.collection.kind` permitted an impossible `"skill"`.** It spelled the union out by hand. It now uses `CollectionKind`, which deliberately excludes `skill`, so the impossible value is no longer representable.

**PRD §5.3.1 called `--all` the only mass-removal shape.** Stale as of this PR, since tap and namespace selectors also remove many skills at once. Rewritten to say what actually justifies the confirmation: those selectors each *name* the collection they empty, while `--all` names nothing and is bounded only by the scope.

**C-UNINST-21 only covered skill-beats-tap.** Added the skill-beats-namespace half.

**The unlocked pre-prompt `--all` read could observe a concurrent truncate.** This one turned out to be broader than the `--all` path. `writeJson` went through `writeFileSync`, which truncates in place, so *any* unlocked reader could see an empty file. §14 exempts read-only commands from the lock, so `list`, `info`, and `doctor` were all exposed, not just the `--all` count.

Rather than patch the one call site, `writeJson` now writes a sibling temp file and renames over the destination. A concurrent reader sees either the old bytes or the new ones, never a partial file. Test added asserting the old inode keeps its content for a reader holding it open across a write, and that no temp file is left behind.

**Collection/`--all` state updates rebuilding state once per skill.** No change needed. This branch already does the filesystem work per entry and then drops state in one batched pass through `dropScopedEntriesAndUpdateRequiredBy`.

## On the dry-run suppression

This branch does not add its own `isSuppressed` handling. It inherits #107's central fix at merge, so a copy here would be redundant.

## Verification

I ran all three gates:

- `bun run lint` — 0 errors (2 pre-existing Biome schema-version infos)
- `bun run typecheck` — clean
- `cd site && bun run lint` — 0 errors (1 pre-existing info)
- `bun test` — 1030 pass, 0 fail, 100% line and function coverage

One note on the test run. A `doctor` test failed partway through this work, and it was my fault but not the branch's: an earlier reproduction of mine had left a stray skill in the worktree's own `.claude/skills/`, and `doctor` scans project scope from the cwd. I removed the debris. It was untracked and never part of the diff.

Per the standing directive on prune loops: I re-read `pruneOrphans` before running anything that drives it. Termination holds. Every candidate is added to `attempted` *before* removal and `findOrphan` skips those keys, so the loop visits each entry at most once and does not depend on the entry disappearing, which matters because a safety-aborted removal deliberately keeps its state entry.

stevekrenzel:
Rebased onto the new #110 tip. README uninstall row reconciled: this PR's selector wording plus #110's scope sentence. Uninstall/list tests pass (61).

stevekrenzel:
Rebased onto the new `fix/uninstall-scope` tip (`1fbfea6`) after the cap-compliance refactors (#135–#143) merged. New tip: `9db61e5`.

Re-placed, since the files these commits edited were split on `main`:

- **`src/cli/args.ts` → `src/cli/args/tables.ts`** — `BOOLEAN_SUB.uninstall` gained `"all"`; the flag tables now live in `tables.ts`, so the entry went there.
- **`site/components/sections/Commands.data.tsx` → `Commands.data/managing.tsx`** — the `uninstall` row edits moved into the "Managing skills" group file. The commit-by-commit history is preserved as-is, including the committed conflict markers the last commit exists to resolve; the final state matches what this PR already had.
- **`tests/misc/gaps.test.ts` → `tests/misc/closeout/util-json-paths.test.ts`** — the oversized misc suites became `tests/misc/closeout/` topic files. The new `writeJson` replace-by-rename test went into the topic file that already covers `src/util/json.ts`, next to the existing `writeJson + readJson round-trip` test in the same `describe("json utilities")` block.

One extra commit was needed to satisfy the cap gate that merged with #137:

- **`refactor(uninstall): split the --all path out of select.ts`** — `select.ts` landed at 202 lines. `allTargets`, `countAllTargets`, `confirmAll`, and `describeScope` moved to a new `src/commands/uninstall/all.ts` unchanged; `groupByName` is now exported so both share it. No behavior change.

`bun run check` is green: typecheck, lint (incl. the cap gate), 1137 pass / 0 fail, coverage 100.00 / 100.00. `cd site && bun run lint` clean.

PR #127, opened 2026-09-11
[LOG-5772] feat(list,info): show a recognizable source instead of the auto-tap name

stevekrenzel:
Stacked on #124 → #118. Please review and merge those first.

## What

After `crew install @anthropics/skills//skills/internal-comms`, both `crew list` and `crew info` showed the source as `skills-internal-comms` — a tap name crew derived from the URL, which the user never typed and can't paste anywhere.

A new `sourceLabel(entry, config)` renders the source the way the user thinks about it:

| Source | Label |
|---|---|
| Registered tap | `core`, `core/skills/demo` (unchanged) |
| Auto GitHub tap | `@anthropics/skills//skills/internal-comms` |
| Auto GitLab / Bitbucket | `gl:acme/skills`, `bb:acme/skills` |
| Auto tap on another host | `https://git.example.com/acme/skills.git` (explicit URL — no shorthand resolves back to it) |
| Auto `file://` tap | the path, display-sanitized |
| Auto path tap | `~/code/my-skills/demo` |
| Tap missing from config | falls back to `<tap>/<path>` |

The label for an auto git tap is itself a valid reference: a test asserts `parseRef(label)` resolves to the same URL and subpath the tap records, so what the user reads is what they can re-install. The one exception is a URL carrying credentials — there the userinfo is masked to `***`, so the label names the right remote but is deliberately not paste-able, because a working URL would carry the secret with it.

`crew info` uses the same label and adds a dim `(tap skills-internal-comms)` for auto taps only, because that derived name is the argument `crew tap remove` takes and hiding it entirely would strand the user.

Before and after, for the reported case:

```
# before
  internal-comms  skills-internal-comms  34040c9c  claude-code, codex, ...
# after
  internal-comms  @anthropics/skills//skills/internal-comms  34040c9c  claude-code, codex, ...
```

## PRD

- **§5.1** — new "Source labels" paragraph defining the rule and noting it is display-only.
- **§9.1** — `crew info` follows the same rule, shows the tap name for auto taps, and adds `source_label` to `--json`.
- **§18.3** — C-LIST-07, C-LIST-08 inside the C-LIST block; C-INFO-02 inside the C-INFO block.

JSON is otherwise unchanged: `source.tap` and `source.path` still carry the raw values, so nothing scripted against them breaks.

## How tested

- `tests/unit/source-label.test.ts` — 13 cases covering every label shape, the `parseRef` round-trip, the orphaned-tap fallback, SCP-style and `host:port` URLs, nested GitLab groups, and a URL crew can't decompose.
- `tests/e2e/source-labels.test.ts` — 7 cases driving the real CLI over a `file://` repo: list and info show the label, `--json` keeps the raw tap/path, `source_label` is present, and a registered tap still shows its plain name with no parenthetical.
- `bun run check` green: 1024 pass, 100% line and function coverage, tsc and Biome clean. All touched files under the 200-line cap.

## Judgment calls

- `formatSource` in `src/commands/list/render.ts` became a one-line call into the shared module; the rest of that file is untouched, per the sibling-PR note.
- A `file://` URL has no host to shorten against and its path *is* the repo identity, so no shorthand applies. It still goes through `displayUrl` like every other label — credentials masked, control characters escaped — so it is a display string rather than a literal echo of what was typed. Documented in §5.1.
- The README `crew list` example previously showed a registered-tap name on a row installed straight from a repo; updated so the example matches real output.
- `crew install`'s already-installed block and `crew update`'s rows don't print a source today, so there was nothing to change there.


stevekrenzel:
## Review

Verdict: **not approved — 0 blockers, 5 majors, 2 minors**

### Majors

- **Several shortened labels do not satisfy the new round-trip contract** (`src/commands/source-label.ts:63`, `:79`). Self-hosted forms such as `git.example.com/acme/skills//…` are not valid Crew refs, and an SCP/SSH source relabeled as `@owner/repo` reparses as HTTPS rather than the stored URL. Preserve an explicit valid clone URL whenever no lossless shorthand exists, and table-test `sourceLabel → parseRef` across HTTPS, SSH/SCP, custom hosts/ports, GitLab, Bitbucket, and file sources.
- **Unescaped subpaths can emit terminal control sequences** (`src/commands/source-label.ts:63`). Auto-tap subpaths flow directly into list/info human output. Reject control characters in parsed/configured subpaths and escape them while rendering existing state.
- **Source labels can disclose URL credentials** (`src/commands/source-label.ts:79`). Fallback labels retain URL query strings and fragments, exposing embedded tokens through list/info and the new JSON field. Strip or redact `search` and `hash` from display labels.
- **The primary site demo still shows the old source format** (`site/components/sections/Demo.tsx:49`). Update it to the new recognizable reference form so the headline example matches actual output.
- **The new config reads can race non-atomic config writes** (`src/commands/list/index.ts:27`, `src/commands/info/index.ts:37`). Concurrent list/info can observe truncated or partial YAML, fail with `config_invalid`, or produce fallback labels. Publish config with a same-directory temp file plus atomic rename so readers remain lock-free.

### Minors

- `PRD.md:161` says a registered tap name is what the user typed, which is false for unnamed `tap add`, and says users cannot act on derived auto-tap names despite the new removal hint. Describe configured versus derived names precisely.
- `sourceLabel` linearly scans `config.taps` for every list row (`src/commands/source-label.ts:31`), making large distinct-tap listings quadratic. Build a tap-name map once per render/command.

All eleven review lenses ran against head `21df7ae`. The normal CI check is absent from this PR’s status rollup. I did not approve because of the major findings.

stevekrenzel:
Addressed all 7 findings in `0b4ad26`, rebased onto the current `feat/list-filters`. Every finding reproduced first; nothing pushed back on.

## Majors

**1 — labels broke the round-trip contract.** Reproduced across ten URL spellings. Two distinct failures, and the second was worse than reported:

```
https://git.example.com/acme/skills.git  -> git.example.com/acme/skills   THROWS (not a valid ref)
git@github.com:acme/skills.git           -> @acme/skills                  parses as HTTPS
```

The self-hosted form did not parse at all; the SSH form parsed *successfully* into a different remote — different transport, different credentials — so the label silently named somewhere the user never installed from.

A shorthand is now used only when it provably round-trips to the exact recorded URL (`shorthandFor` requires `https://`, a known host, and exactly two path segments). Everything else keeps an explicit URL. A nested GitLab group also falls back, since `gl:` expands to exactly two segments. `tests/unit/source-label-roundtrip.test.ts` table-tests nine spellings plus the nested-group case, asserting the parsed URL equals the input, not merely that it parses.

**2 — control characters reached the terminal.** Reproduced: a subpath of `skills/\x1b[2K\rforged` rendered intact, clearing the line and forging a `crew:` prefix. Rendering now escapes anything below `0x20` plus `0x7f`. I escape at render rather than only rejecting at parse, because existing `state.json` and `config.yaml` predate any parser guard.

**3 — credentials in labels.** Reproduced: `?access_token=sEcReT` survived into the label and the new `source_label` JSON field. Query strings and fragments are now dropped, and a `user:password@` pair is masked to `user:***@`.

One deliberate exception: a bare `git@host` username is kept. It is the protocol's fixed account rather than a secret, and stripping it yields a label that does not clone. Same call the #108 worker made for the verbose renderer.

> **Merge note.** This is the third credential-disclosure path found in this stack. #108 has `src/util/redact.ts`, #113 has `src/refs/display-url.ts`, and neither is in this base. I wrote only what this PR needs (`displayUrl` / `maskUserinfo` in `source-label.ts`) rather than a third general implementation. **All three should collapse onto one renderer when the branches meet.**

**4 — site demo.** `Demo.tsx` showed `acme@v1.2.0` as a source, which is neither the old format nor the new one — the command never produced it. Now shows `@acme/skills//engineers/founding`, matching the install line directly above it.

**5 — non-atomic config writes.** Confirmed the cause you identified, and it is broader than config: `writeConfig`, `writeState`, every `.crew.json` marker, and `version-check.json` all route through `writeText`, which called `writeFileSync` directly on the target. Fixing it there covers all four at once — bytes land in a sibling temp file, then `renameSync` publishes them.

On testing it: a torn read is timing-dependent and I was asked not to write a flaky test. My first two attempts asserted properties that **passed against the unfixed code**, so I discarded them rather than let the count imply protection they did not provide. The kept assertion is that the target's inode changes across a write, which only a rename does — verified to fail against the in-place implementation and pass with it.

## Minors

- **§5.1 wording**: a registered tap's name is not always what the user typed (`crew tap add <url>` with no name derives one). Reworded around *configured* vs *derived* names, and noted that derived names stay visible where they are actionable — `crew info` still shows the name `crew tap remove` takes.
- **Quadratic scan**: `tapIndex(config)` builds a name→tap map once per render; `sourceLabel` takes the map instead of scanning `config.taps` per row.

## Rebase

Two conflicts against the updated parent, both additive rather than competing: the parent's `hasAgentOrTapFilter` rename beside my `config` argument in `list/index.ts`, and adjacent PRD paragraphs. Kept both sides. The parent had claimed `C-LIST-07`, so my criteria renumbered to **C-LIST-08/09**, plus a new **C-LIST-10** for the display-safety rules. `feat/list-filters` is an ancestor; PR is MERGEABLE.

## Pre-fix verification

The round-trip, control-character, credential, and inode tests all fail against the pre-fix code. Two existing unit tests asserted the buggy behavior directly (`git.example.com/acme/skills` and SSH→`@acme/skills`) and were rewritten — they were encoding the defect.

`bun run check`: 1040 pass, 0 fail, 100% line and function coverage, site lint clean, every touched file under the 200-line cap.

stevekrenzel:
## Updated /review

**Verdict:** 3 blockers, 3 majors, 4 minors. Not approved.

### Blockers

- `src/commands/source-label.ts:144-151` — Username-only HTTP(S) userinfo is emitted verbatim. A valid clone URL such as `https://ghp_secret@github.com/acme/skills.git` leaks the token through `crew list`, `crew info`, and `source_label`. Strip/mask all scheme URL userinfo; retain `git@` only for SSH/SCP remotes, and add a regression test.
- `src/commands/source-label.ts:120-128` — The docblock says the redacted URL “still clones,” which is false after password masking. Describe display identity/parseability accurately.
- PR description — The “Auto tap on another host” example says `git.example.com/acme/skills`, while the implementation retains the explicit `https://.../.git` URL. Correct the table to match shipped output.

### Majors

- `PRD.md:183` — The contract says credentials and control characters are removed, while the implementation masks passwords and escapes controls. Specify masking/escaping precisely so the PRD and conformance tests agree.
- `src/util/fs.ts:56` — Atomic replacement creates a new temp file under the process umask and discards an existing target’s permissions. A `0600` config containing credential-bearing clone URLs can become `0644`. Preserve the existing mode before rename or explicitly restrict sensitive files.
- `src/commands/source-label.ts:173` — `safeLabel` escapes C0/DEL only, leaving C1 controls such as U+009B in terminal output. Escape the full Unicode control range.

### Minors

- `src/commands/source-label.ts:94` — `repoRef` implies an exact, pasteable reference even though it may redact/drop data. Rename it to `repoLabel` or `renderRepoLabel`.
- `src/commands/source-label.ts:28` — `HOST_SHORTHANDS` duplicates parser shorthand knowledge from `refs/git-url.ts`; export one definition and derive display prefixes from it.
- `tests/e2e/source-labels.test.ts:44,64,83` — These source-label tests cite C-LIST-07, but the behavior belongs to C-LIST-08.
- `tests/e2e/source-labels.test.ts:78` — The JSON test claims raw `source.tap/path` stay unchanged but does not assert either exact value. Compare the complete source object with the stored entry.

stevekrenzel:
All 3 blockers, 3 majors, and 4 minors addressed in `e709809`. Nothing declined. Rebased onto the updated parent first; `feat/list-filters` is an ancestor and the base is unchanged.

## Blockers

**1 — username-only userinfo leaked.** Reproduced before fixing:

```
repoLabel("https://ghp_secret@github.com/acme/skills.git")
  -> https://ghp_secret@github.com/acme/skills.git
repoLabel("https://user:pw@github.com/acme/skills.git")
  -> https://user:***@github.com/acme/skills.git
```

Masking only the password half missed the commonest shape there is: a personal access token is normally the *username* with no password at all. `maskUserinfo` now replaces the entire userinfo component of a scheme URL with `***`. An SCP-style remote (`git@host:owner/repo`) keeps its user — there it names the protocol account rather than a secret, and dropping it would produce a label that resolves to nothing. Regressions cover the username-only, `user:password`, and `ssh://` forms.

**2 — docblock claimed the result "still clones".** False as soon as anything is masked. It now says the label identifies the remote and is not a URL to clone with, and explains why the SCP case is different.

**3 — description table.** The "Auto tap on another host" row claimed `git.example.com/acme/skills`; the implementation keeps the explicit `https://.../.git` URL, because no shorthand resolves back to it. Corrected, along with the round-trip paragraph, which now records the credential exception.

## Majors

**1 — PRD said "removed", code masks and escapes.** §5.1 and C-LIST-10 now state the actual rules: userinfo masked to `***`, query and fragment dropped, C0/DEL/C1 escaped as `\xNN`, SCP user retained.

This also surfaced a genuine conflict between two criteria that the spec did not resolve. C-LIST-09 promises a label round-trips to the exact URL; C-LIST-10 now masks all scheme userinfo. `ssh://git@host/…` triggers both. Masking has to win — publishing a working URL would mean publishing the secret in it — so C-LIST-09 is qualified to URLs with no credentials to remove, and a test pins the precedence rather than leaving it to whichever rule a reader applies first.

**2 — file modes.** Reproduced:

```
mode before: 600 -> after: 644
PRESERVED: false
```

`writeText` created its temp file under the umask and renamed it over the target, so the published file inherited the *temp* file's mode, not the target's. A config the user had restricted to `0600` silently widened on its next write — and that file can hold credential-bearing clone URLs.

Fixed in `writeText` itself rather than at a call site, because the helper is shared by `config.yaml`, `state.json`, every marker, and `version-check.json`. An existing file's mode is read before the write and applied to the temp file, so it survives the rename. **For a file crew is creating for the first time I chose `0600`** rather than the umask default: both `config.yaml` and `state.json` can hold credentials or a full inventory of install locations, and a file that starts owner-only can be widened deliberately, while one that starts world-readable has already leaked. A deliberately wider mode on an existing file is preserved, not narrowed — three tests cover those cases.

**3 — C1 controls.** Reproduced: U+009B survived `safeLabel` unescaped. A terminal in 8-bit mode reads it exactly as `ESC [`, so escaping C0 alone still left a way to move the cursor. Both ranges are now escaped. Worth noting that the tooling I used to write the repro refused a command containing a raw control character — the same hazard, one layer up.

## Minors

- `repoRef` → `repoLabel`: it may redact or drop data, so "ref" overstated what it returns.
- `HOST_SHORTHANDS` is now derived by inverting the parser's own `SHORTHAND_HOSTS` (exported for this), so the two cannot disagree about which hosts have a shorthand. GitHub stays special-cased, since §8.2 calls `@owner/repo` the ergonomic alias.
- Stale `C-LIST-07` IDs in the e2e suite corrected to `C-LIST-08`.
- The JSON test asserted only that the tap was *not* a URL and that a path was a string, which would have passed on any wrong value. It now compares the full `source` object against what state actually stores, plus the specific shape of each field.

## Verification

Both blocker behaviours were reproduced against the pre-fix implementation, and the two file-mode tests fail against unmodified `src/` (the source-label unit file cannot load pre-fix, since it imports the moved path — that proves nothing about behaviour, so I am not counting it).

Targeted suites: `source-label` + `source-label-roundtrip` + `atomic-write` 33 pass; `source-labels` + `list-scope` + `list-filters` 20 pass; `atomic-write` + `info` + `gaps` 49 pass. Typecheck and Biome clean, all touched files under the 200-line cap.

No loops were added — the changes are a mode lookup, a character-class predicate, a table inversion, and assertions — so there was nothing new to check for termination.

Two runs of `source-labels.test.ts` failed on the tests that use `install acme/demo`, a bare-name resolution that scans every configured tap including the GitHub-backed default; three subsequent runs passed 6/6, and those two tests take 6–10s versus ~100ms for the rest. That is the known non-determinism #131 removes, not this change.

`source-label.ts` reached 209 lines with the added rationale, so it is now `source-label/` with the display-safety helpers in their own module.

**Merge note.** This is the third URL-redaction implementation in the stack: #108 has `src/util/redact.ts` (`redactUrl`/`redactText`/`redactDetails`) and #113 has `src/refs/display-url.ts`. I did not add a fourth — `display-safety.ts` is scoped to what a *label* needs. At merge these should collapse onto `util/redact.ts` as the general renderer, and two rules from this PR are the ones to keep: mask the whole userinfo rather than the password half, and escape C1 as well as C0.

stevekrenzel:
Rebased onto the revised parent `feat/list-filters` (`dcd9f11`), after #118 and #124 were both rebased beneath it. Four of this branch's commits were replays of work already in the parent and were skipped as duplicates (verified by matching subject and content against `39c4ab4` and `dcd9f11` before skipping, not assumed).

Two real conflicts, both in the PRD criteria table, both pairing the parent's corrected C-STATE-10 wording against this branch's criteria:

- C-LIST-08..10 added alongside the parent's C-STATE-10.
- C-LIST-09/10 again, where the incoming side is this branch's round-two refinement — the one that resolved the precedence question between "a label round-trips to the exact URL" and "a label never discloses credentials". Kept that refined version, since dropping it would reintroduce the contradiction.

Verified: parent is an ancestor, typecheck and Biome clean, `source-label` + `source-label-roundtrip` + `atomic-write` 33 pass / 0 fail. MERGEABLE.

stevekrenzel:

**Verdict:** 0 blockers, 0 majors, 6 minors, 2 nits. **Approved.**

Reviewed `dcd9f112..93ddc0aa` through all 11 `/review` lenses. This stacked PR has no test-suite CI run at this head.

### Minors

1. Delete orphaned `src/commands/source-label.ts`; all callers use `source-label/index.ts`, while the duplicate retains obsolete masking/escaping behavior.
2. `crew info` renders raw auto-tap names outside `safeLabel` (`src/commands/info/render.ts:46`). Escape control characters and add a malformed persisted-name regression.
3. Failed `writeText` write/rename leaves its `.tmp-*` sibling, potentially containing config credentials (`src/util/fs.ts:72`). Clean it up without masking the original error.
4. `crew list --json` reads/parses config solely to build discarded human labels, and `--tap` can parse it twice (`src/commands/list/index.ts:24`). Skip human rendering work in JSON mode and reuse loaded config.
5. Help says auto-tap labels can be typed to reinstall, but credential-bearing labels are deliberately redacted display strings.
6. The source-label comment/PR body says `file://` URLs render exactly as typed, though `displayUrl` sanitizes them (`src/commands/source-label/index.ts:106`).

### Nits

1. Run the full CI gate after rebasing/retargeting to `main`.
2. Keep the source-label module available only through its directory entry point to avoid future imports of stale logic.

stevekrenzel:
Round three addressed in `55fadcb`. Rebased onto the updated parent (`ba26d72`) first — clean, no conflicts, and `feat/list-filters` is an ancestor. Nothing declined; all six minors and both nits reproduced or verified before I changed anything.

**Minor 1 — orphaned module.** Confirmed: `src/commands/source-label.ts` existed, and a grep showed every importer already going through `source-label/index.ts`. Deleted; typecheck stayed clean, which is the proof it was unreachable. It mattered because the flat copy still carried the pre-round-two masking and escaping, so one stray import would have silently reverted that fix.

**Minor 2 — raw tap name in `crew info`.** Reproduced. `info.sourceLabel` is sanitized but `info.tapName` was interpolated raw on the same line, so the one value that skipped escaping was the persisted one. Now escaped via `safeLabel`. New regression writes an escape sequence into a persisted auto-tap name and asserts the raw character never reaches stdout. **Verified load-bearing**: reverting just that call fails exactly this test, 6 pass / 1 fail.

**Minor 3 — orphaned temp file.** Reproduced with a rename that cannot succeed. The `.tmp-*` sibling survived with its full contents, which for `config.yaml` means credential-bearing clone URLs sitting on disk under a predictable name. Cleanup now runs on failure, and a nested `try` swallows any unlink error so the original failure still propagates — masking it would hide the actual cause. New test in `tests/misc/tmpclean.test.ts`, also verified load-bearing.

**Minor 4 — wasted config read under `--json`.** Confirmed both halves: `readConfig` ran and `renderList` built human lines that JSON mode discards, and `readListFilters` separately parsed config for `--tap`, so a `crew list --tap x --json` parsed `config.yaml` twice. JSON mode now returns before rendering, and `ListFilters` carries the config `--tap` validation already loaded.

**Minor 5 — help overstated the label.** The list help said the source column shows "the reference you'd type to install it again". That is false precisely when it matters: a credential-bearing URL has its whole userinfo masked to `***`, so typing the label back would not resolve. Reworded to say it is built for reading, and to re-type the original reference.

**Minor 6 — `file://` comment.** Corrected in the source and in the PR body. The value has already been through `displayUrl` by that point, so "exactly as the user would type it" was wrong; it is a display string like every other label.

**Nit 1 — gates.** Ran what is available to a stacked PR: typecheck clean, `bun run lint` clean, `cd site && bun run lint` clean, and targeted suites — `source-labels`, `tmpclean`, `list-filters`, `list-scope`, `source-label`, `source-label-roundtrip`, `atomic-write` — **56 pass, 0 fail**. I did not run the full suite, per the standing resource limit. `git grep` for conflict markers on the final tree is empty.

**Nit 2 — single entry point.** `safeLabel` is now re-exported from `source-label/index.ts` and the new import goes through it, so nothing outside the directory reaches into its internals. That is what let the stale flat module drift in the first place.

One note for the merge: this branch carries no copy of the `isSuppressed` dry-run fix from #107 and inherits it centrally at merge.

stevekrenzel:
Rebased onto the new #124 tip (`e1d6985`); no conflicts. Targeted tests (list, aliases, flag-guards, info, source-label) → 72 pass; lint, typecheck, and site lint clean; all commits signed.

stevekrenzel:
Rebased onto the new `feat/list-filters` tip (`d4bb01e`) after the cap-compliance refactors (#135–#143) merged. New tip: `ff2d14e`.

No conflicts — all four commits replayed cleanly, and every file this PR adds or grows is already under the 200-line cap that #137 now enforces.

`bun run check` is green: typecheck, lint (incl. the cap gate), 1169 pass / 0 fail, coverage 100.00 / 100.00. `cd site && bun run lint` clean (this PR touches `site/components/sections/Demo.tsx`).

PR #128, opened 2026-09-11
[LOG-5773] fix(install): same repo at the same path is not a name conflict

stevekrenzel:
## What and why

Installing a repo subpath and then the whole repo reported `name_conflict` for every shared skill, even though both refs reach the same directory in the same repo at the same commit:

```
$ crew install @anthropics/skills//skills/docx
✓ docx

$ crew install @anthropics/skills
Error (name_conflict)
  a skill named `docx` is already installed from a different source
```

`--force` doesn't override `name_conflict` (§13), so the only way out was uninstalling first. `crew update` tap re-expansion hit the same wall from the other direction: a broader tap would try to "add" a child already installed via a narrower one.

The cause was comparing **tap name + tap-relative path**. One repository can back several tap rows: `<url>//skills/docx` records a tap with `subpath: skills/docx` and an entry path of `""`, while `<url>` records a tap with no subpath and an entry path of `skills/docx`. Different tap, different path, same directory.

## What changed

**Canonical source identity** (`src/install/source-identity.ts`): one root plus a location beneath it. For a git tap the root is the repository and the tap's `subpath` folds into the location; for a path tap the directory folds into the location the same way, so `/src/skills/docx` and `/src` reach one identity. URL comparison normalizes the spellings crew itself produces — a trailing `.git`, a trailing `/`, and **host** casing — so `gh:acme/skills`, `@acme/skills`, and `https://github.com/acme/skills` are one repository. Userinfo and path casing are compared as written, since both can be significant on the server.

**Re-attribution** (`src/install/reattribute.ts`): when both installs agree on location but the existing entry sits on an **auto** tap, the entry moves to the incoming tap. `state.source` and the install-site markers are rewritten (markers carry the full tap descriptor, since they're authoritative for `doctor --repair` per §11.1), the installed bytes are untouched, and the emptied auto tap plus its clone are collected. Entries on **registered** taps are never re-attributed — the user named that tap deliberately. An entry subscribed to a whole tap is never moved onto a tap rooted strictly deeper in the same source, which would lose the sibling subscription.

**Re-expansion guard** (`src/install/installed-lookup.ts`): `crew update` no longer "adds" a child already installed at the same scope from the same canonical location through another tap. The index is keyed by install location and is updated as additions happen, so two tap rows covering one repository can't each install the same new child in a single run.

**Deletion containment** (`src/util/fs.ts`, `src/config/taps.ts`): a tap name reaches a recursive delete from `config.yaml`, so the parser rejects anything that isn't a single directory component and `rmrfInside` refuses a target resolving outside the directory crew owns.

Genuinely different sources still produce `name_conflict`, unchanged.

Human output on the re-attribution path:

```
  docx (already installed) · fbf67ddc
    now tracked via repo
```

`--json` gains `reattributedFrom` on the affected `already_installed[]` entry.

## PRD sections

- **§5.4** — "What the source means" and "Re-attribution" paragraphs defining canonical location, the auto-vs-registered rule, host-only case folding, and the path-tap equivalence.
- **§16.5** — auto taps are collected on re-attribution, not just on uninstall.
- **§13** — `name_conflict` row notes that the same repo via a different tap is not a conflict.
- **§18.3** — C-INST-13a through C-INST-13k after C-INST-13, plus C-TAP-25 for deletion containment.

## Where the shared GC lives

`gcAutoTaps` moved out of `src/commands/uninstall/index.ts` into `src/maintenance/auto-taps.ts` as `garbageCollectAutoTaps`, and both install and uninstall call it. It now returns the collected names so callers can report or assert on them.

## How tested

30 tests across 7 files, against real local `file://` repos and directories — no network.

`tests/e2e/same-source-install/`:

- `identity.test.ts` (6) — canonical URL comparison across `.git`, trailing-slash, host-case, and scp-style spellings; userinfo and path case preserved; subpath-vs-whole-repo identity; path taps; re-expansion doesn't duplicate a child installed from the same source
- `reattribution.test.ts` (5) — state and markers move and the emptied auto tap plus clone are collected; the same at **project** scope; a move in one project leaves another project's marker alone; the destination tap's discovery mode wins; `--json` carries `reattributedFrom`
- `attribution-kept.test.ts` (2) — a **registered** narrow tap keeps its attribution; a whole-tap subscription is not narrowed onto a subpath tap
- `path-breadth.test.ts` (2) — the same two directions for **path** taps: installing the parent directory re-attributes the child, and a whole-directory subscription is not narrowed onto a subdirectory
- `conflict.test.ts` (2) — a different repo with the same skill name still conflicts (exit 4)
- `containment.test.ts` (7 executed, table-driven) — `rmrfInside` refuses an outside target; `isInside` rejects the root itself; a tap name carrying a traversal, `/`, `\`, `.`, or `..` is rejected on config load

`tests/unit/installed-lookup.test.ts` (6) — the index agrees with a scan, scope and project root partition entries, an entry on a missing tap never matches, and an addition recorded mid-run is visible to a later lookup.


stevekrenzel:
## Review

Verdict: **not approved — 2 blockers, 5 majors, 3 minors**

### Blockers

- **The new E2E file violates the hard file-size cap** (`tests/e2e/same-source-install.test.ts:1`). It is 308 lines; split the scenario groups into focused files and extract shared repository/adapter helpers.
- **Auto-tap GC can delete outside the taps directory** (`src/maintenance/auto-taps.ts:31`). Persisted tap names reach `tapPath()` and `rmrf()` without validation, so a traversal name can escape the managed root. Validate names at config/repair boundaries and enforce that every deletion target remains inside `tapsDir`.

### Majors

- **Project marker rewriting ignores the target project root** (`src/install/reattribute.ts:64`). Re-attributing project A can rewrite project B’s marker for the same skill/tap while B’s state remains unchanged. Match `move.projectRoot` and add a two-project regression.
- **Re-attribution can narrow and lose tap tracking** (`src/install/duplicate-rules.ts:112`). A whole-repo auto tap may move onto a narrower incoming subpath tap; GC then removes the broader-repo tap, so future re-expansion no longer sees siblings. Reattribute only when the incoming tap preserves or broadens tracking, otherwise retain the existing attribution.
- **Old recursive discovery survives re-attribution** (`src/install/reattribute.ts:85`). Spreading the old marker preserves `tap_discovery` when the destination tap is non-recursive, corrupting authoritative marker metadata and later doctor repair. Remove it unless the destination tap is recursive.
- **SCP-style SSH host casing is not canonicalized** (`src/install/source-identity.ts:71`). `git@GitHub.com:…` and `git@github.com:…` compare differently despite the new host-case-insensitive identity contract. Lowercase only the SCP host segment and test it.
- **Duplicate/re-attribution decisions use an unlocked state snapshot** (`src/install/flow.ts:90`). An uninstall between analysis and lock acquisition can yield a successful “already installed” result when the skill is absent. Recompute the analysis from `freshState` after taking the lock.

### Minors

- Marker rewriting scans every remembered project root and adapter under the state lock (`src/install/reattribute.ts:56`), even for a user-scope move. Restrict traversal to roots in the reattribution plan and index moves/taps.
- Same-source lookup rescans state/taps for every re-expanded child (`src/install/tap-reexpand.ts:162`). Build a source-identity index once per re-expansion.
- Several comments call the destination a “broader tap,” although the code accepts any incoming tap covering the location. Rename it to “incoming tap” or “tap covering the same location.”

All eleven review lenses ran against head `824074e`; CI is green. I did not approve because of the blockers and major findings.

stevekrenzel:
Addressed all 10 findings in `fd3420e`. Nothing pushed back on — every one reproduced first.

## Blockers

**Deletion could escape the managed root** (`src/maintenance/auto-taps.ts`, `src/commands/tap/index.ts`, `src/util/fs.ts`, `src/config/taps.ts`). Reproduced before fixing: `tapPath("../PRECIOUS", home)` resolves to a sibling of the taps directory, and that value went straight into `rmrf()`.

What is validated, and where:
- **At the boundary** — `parseTapEntry` now rejects any tap name that isn't a single directory component (`/`, `\`, `.`, `..`, NUL) with `config_invalid`.
- **At the deletion sites** — both tap clone deletions go through a new `rmrfInside(root, path)`, which resolves both sides and refuses anything outside the root. `isInside` also rejects the root itself, so a bug that produced the bare taps directory can't wipe it.

**`rmrf` sweep.** Of the 17 call sites, only two build a path from a persisted name: auto-tap GC and `crew tap remove`. Both are fixed. The store GC (`src/maintenance/gc.ts`) looked similar but is safe — it enumerates real directory entries with `listDir` rather than joining stored strings. The rest are internal paths (staging dirs, scheduler artifacts, the cache dir).

**One deliberate narrowing of your suggestion.** You asked for `NAME_PATTERN` at the config boundary. I tried that first and it broke three existing tests: auto-tap names derive from directory basenames, so a repo in `/tmp/crew-test-hMW9zP` yields a mixed-case tap name that `NAME_PATTERN` rejects. Applying it on load would have failed real users' existing configs on upgrade. The check enforces the containment property the deletion paths actually depend on; `crew tap add` still applies the full slug rule to names it creates.

**File-size cap** — the 308-line file is now `tests/e2e/same-source-install/`: `helpers.ts` (61), `conflict.test.ts` (58), `reattribution.test.ts` (186), `attribution-kept.test.ts` (66), `identity.test.ts` (143), `containment.test.ts` (83).

## Majors

**Project marker rewriting ignored the target root** (`src/install/reattribute.ts`). `maybeRewrite` matched on name, scope and `fromTap` but never `projectRoot`, so re-attributing project A rewrote project B's marker while B's state stayed put — leaving marker and state disagreeing, with the marker authoritative (§11.1). Now matched on project root; C-INST-13f covers two projects.

**Re-attribution could narrow tap tracking** (`src/install/duplicate-rules.ts`, new `src/install/tap-breadth.ts`). My first fix keyed off `existing.tracks_tap && !skill.tracksTap` and did not work — every direct git/path install sets `tracksTap: true`, including a subpath install, so the condition never fired. I found that by running the test rather than trusting the change. The real question is which tap sees *more*: `narrowsTapRoot` compares canonical repo roots and refuses an incoming tap rooted strictly deeper. C-INST-13h covers it.

**Old recursive discovery survived re-attribution** (`src/install/reattribute.ts`). `...marker` carried `tap_discovery` onto a non-recursive destination. Now dropped unless the destination tap is recursive; C-INST-13g.

**SCP host casing** (`src/install/source-identity.ts`). Confirmed `canonicalRepoUrl("git@GitHub.com:acme/skills.git") !== canonicalRepoUrl("git@github.com:acme/skills.git")` while the HTTPS control compared equal — the `schemeEnd < 0` branch returned before any host handling. Only the segment before `:` folds; the path stays case-sensitive. C-INST-13i.

**Unlocked snapshot** (`src/install/flow.ts`). The analysis now runs against state read inside the lock; the dry-run path keeps its unlocked preview, since it writes nothing. `promoteExplicit` moved to its own module to stay under the cap.

## Minors

- Marker rewriting visits only the project roots a move names, and resolves taps through a map instead of a linear scan per marker.
- Re-expansion builds one `InstalledSourceIndex` per run instead of rescanning state (and every tap row inside `identityOfStateSource`) per child. The old per-call scan had one caller, so I replaced it rather than leaving two implementations.
- "broader tap" → "incoming tap" / "tap covering the same location" throughout.

## Pre-fix verification

Stashed `src/` only, kept the tests: **4 of the new tests fail against the pre-fix commit** — C-INST-13f, C-INST-13g, C-INST-13h, C-INST-13i, one per major. The containment tests fail to *load* pre-fix (they import `rmrfInside`, which doesn't exist yet), so they prove the guard exists rather than catching a prior regression; I'd rather say that than let the count imply four more.

**Two of my own mistakes, corrected rather than papered over.** The containment test originally ran `crew list`, which reads state and never touches config, so it exited 0 and proved nothing — switched to `crew tap list`. And the discovery test first built a *registered* recursive tap, which is never re-attributed by design; it now uses `--recursive` on a direct install to get a recursive auto tap.

## Spec and cross-PR

PRD §13's `config_invalid` row now covers the tap-name rule, and C-TAP-25 is added to the C-TAP block.

**Cross-PR:** #132 (`feat/shared-tap-clones`) adds its own canonical-URL logic for repo-keyed clone paths. `src/install/source-identity.ts` should be the single canonicalizer — worth reconciling whichever lands second, especially now that it carries the SCP host rule.

`bun run check`: 1006 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean. (`src/install/perform.ts` is 260 lines, over the cap — pre-existing on `main` and untouched here.)

stevekrenzel:
## Updated /review

**Verdict:** 2 blockers, 2 majors, 5 minors, 1 nit. Not approved.

### Blockers

- `src/util/fs.ts:52` — `isInside` says the root itself is allowed, but the implementation intentionally returns false for it. Document this as a strict-descendant check and explain that excluding the root protects recursive deletion.
- PR description — “How tested” references deleted `tests/e2e/same-source-install.test.ts` and 10 tests. The head has five split suites with 16 tests. Update the description to the final layout/count.

### Majors

- `src/install/source-identity.ts:81` — URL normalization lowercases the whole authority, including SSH userinfo. Case-distinct SSH usernames can identify different accounts/repos but compare equal, bypassing `name_conflict` and permitting overwrite without `--force`. Parse the URL and lowercase only the hostname; preserve userinfo.
- `src/install/tap-reexpand.ts:70` — `installedIndex` is a snapshot and is not updated after `installOne` succeeds. With overlapping whole-tap groups, a later group can add the same newly discovered child again, leaving marker and state attribution inconsistent. Update/rebuild the index after additions and cover overlapping taps.

### Minors

- `tests/e2e/same-source-install/containment.test.ts:50` — C-TAP-25 requires rejecting `/`, `\`, `.`, and `..`, but only slash traversal is tested. Add table-driven cases for the other forms.
- `src/install/tap-breadth.ts:20` — The path-tap branch has no end-to-end re-attribution coverage. Add child→parent re-attribution/GC and parent→child attribution-preservation cases.
- `tests/unit/installed-lookup.test.ts:49` — `as StateEntry` suppresses fixture schema checking. Use `satisfies StateEntry` or an annotated local.
- `src/install/reattribute.ts:28,87` — Bulk re-attribution performs entry×move and marker×move linear searches under the lock. Build keyed maps once.
- `src/install/installed-lookup.ts:65` — Indexing by skill name alone makes multi-project re-expansion scan every same-named entry. Key by name, scope, and project root for O(1) lookup.

### Nit

- `src/install/tap-breadth.ts:20` — `narrowsTapRoot(existing, incoming)` leaves direction ambiguous. Prefer a directional name such as `isTapRootNarrowerThan(incoming, existing)`.

stevekrenzel:
All nine findings addressed in `f1cad1b`. Nothing declined.

## Blockers

**`isInside` docstring** (`src/util/fs.ts`). Correct — it said "resolves to `root` itself or something beneath it" while `rel.length > 0` returns false for the root. Now documented as a strict-descendant check, with the reason: the caller is `rmrfInside`, so a bug that collapsed a target down to the managed root would otherwise wipe every tap clone at once. Refusing the root makes that a no-op.

**PR description.** Rewritten against the current head: 30 tests across 7 files, listed per suite, replacing the reference to the deleted single file and its 10 tests. Also corrected the criteria range and added the containment and path-tap work the old text predated.

## Majors

**Authority case folding** (`src/install/source-identity.ts`). Reproduced before fixing. Worth being precise about where it was, because the SCP branch was already correct:

| input | before | after |
|---|---|---|
| `ssh://Alice@GitHub.com/Acme/Skills.git` | `ssh://alice@github.com/Acme/Skills` | `ssh://Alice@github.com/Acme/Skills` |
| `https://Alice@github.com/acme/skills.git` | `https://alice@github.com/...` | `https://Alice@github.com/...` |
| `git@GitHub.com:Acme/Skills.git` | already correct | unchanged |

So `ssh://Alice@host/a/b` and `ssh://alice@host/a/b` compared equal. Since §5.4 reads "same source" as permission to overwrite without `--force`, a second install would have silently replaced a genuinely different source instead of raising `name_conflict`. Host folds; userinfo and path are compared as written. Both branches now share one `lowercaseHostOnly` helper, so they can't drift apart again.

**Stale re-expansion index** (`src/install/tap-reexpand.ts`, `installed-lookup.ts`). Confirmed: the index was built once before the group loop and never updated, so two tap rows covering one repository each saw the same new upstream child as absent. New `noteInstalled` records additions as they happen. I also took the fifth minor here — the index is now keyed by `(name, scope, project_root)` rather than by name alone, so a multi-project re-expansion stops scanning every same-named entry.

## A third defect, found while writing the coverage you asked for

Minor 2 asked for end-to-end path-tap re-attribution coverage. Writing it surfaced that the path branch never worked:

```
$ crew install /src/skills/docx     # ok
$ crew install /src
Error (name_conflict)
  a skill named `docx` is already installed from a different source
```

Git taps folded the tap subpath into `location`, leaving the repository as the root; path taps left the whole directory in `root` and only the entry path in `location`. So `/src/skills/docx` + `""` and `/src` + `skills/docx` were two different roots and never compared equal — the exact case §5.4 says is one source, and the prose already claimed ("the same repository **or the same local directory** and the same path inside it"). Both kinds now resolve to one root plus a location beneath it. `joinPosix` trims leading slashes, so the absolute marker is restored explicitly; without that, `/src` and `src` would collide.

## Minors and nit

- **Containment cases** — table-driven over all four shapes C-TAP-25 names (`/`, `\`, `.`, `..`) plus the original traversal, YAML-quoted so `.` and `..` stay strings.
- **Path-tap re-attribution coverage** — new `path-breadth.test.ts` covers both directions: child→parent re-attributes and collects the emptied auto tap, parent→child preserves the whole-directory subscription.
- **Fixture typing** — `satisfies StateEntry` instead of `as StateEntry`, so a change to the type breaks the fixture rather than being suppressed.
- **Keyed maps in `reattribute.ts`** — both loops index once instead of scanning the move list per entry and per marker. The marker key includes `fromTap`, since a move only claims markers still naming its old tap.
- **`narrowsTapRoot` → `isTapRootNarrowerThan(incoming, existing)`** — reads directionally at the call site; argument order swapped to match.

## Criteria that existed only as test titles

While adding IDs for the new tests I found C-INST-13f, 13g, 13h, and 13i referenced in test titles from round one but never written into §18.3, so those IDs traced to nothing. All four are now in the table, alongside new **C-INST-13j** (path-tap equivalence) and **C-INST-13k** (no duplicate add across overlapping tap rows).

## Verification

Pre-fix, with only `src/` reverted, three of the new tests fail — one per defect:

```
(fail) C-INST-13i userinfo case is preserved in scheme'd URLs
(fail) C-INST-13j installing the parent directory re-attributes the child
(fail) C-INST-13h a whole-directory subscription is not narrowed onto a subdirectory
```

Targeted suites post-fix: **30 pass, 0 fail** across `tests/e2e/same-source-install/` and `tests/unit/installed-lookup.test.ts`; 39 pass across `multi-skill-install`, `tap-unify`, and `tap-reexpand`, run because they consume source identity. `source-identity.ts`, `installed-lookup.ts`, `reattribute.ts`, and `tap-breadth.ts` are each at 100% line and function coverage. Typecheck and Biome clean; every touched file under the 200-line cap.

One run of `same-source-install/` showed `C-INST-13c` failing at 5.2s against a sub-second norm. I checked rather than assuming: it passes in isolation, the unmodified base passes it twice, and my branch passes it three times consecutively. That is the network non-determinism #131 removes, not this change. I report per-suite numbers rather than a full-suite total for the same reason.

The only loops I changed are bounded passes over fixed arrays (`movesByLocation`, the additions loop); I re-read each before running anything against it.

## For the merge

#132 has its own canonical-URL logic in `src/core/repo-path.ts` for repo-keyed clone paths. `source-identity.ts` is the more general one and should win, but note the rule has changed since #132's reply was written: folding is now **host-only**. If #132's path-keying folds the whole authority, two accounts would share a clone directory, so that needs reconciling rather than a straight swap.

stevekrenzel:

**Verdict:** 0 blockers, 2 majors, 5 minors, 0 nits. **Not approved.**

Reviewed `6428d43..f1cad1b8` through all 11 `/review` lenses. CI is green at this head.

### Majors

1. **Broad-tap reattribution does not enable whole-tap tracking** — `src/install/flow.ts:157`
   Installing a single child first, then its one-skill parent repo reattributes the entry but bypasses `performInstall`, so `tracks_tap` remains false. Future updates skip newly added siblings. Promote `tracks_tap` alongside `explicit` and add a one-skill-repo regression.
2. **`--force` or a new-agent install can bypass the guard against narrowing a broad subscription** — `src/install/duplicate-rules.ts:167`, `src/install/perform.ts:203`
   Classification rejects the reattribution, but installation still overwrites `state.source` with the incoming narrow tap. Preserve the existing source whenever reattribution is rejected.

### Minors

1. PRD §5.4 still says `--force` can replace a different source, contradicting the non-forceable `name_conflict` rule (`PRD.md:171`).
2. Use `Scope` rather than `string` in reattribution key helpers (`src/install/reattribute.ts:24`).
3. Add an end-to-end overlapping-whole-tap update case; the live-index regression currently calls `noteInstalled()` directly (`tests/unit/installed-lookup.test.ts:139`).
4. Targeted reattribution scans every adapter's marker tree under the state lock (`src/install/reattribute.ts:96`). Inspect only moved entries' known owners/paths.
5. `src/install/tap-reexpand.ts` is 204 lines, above the repository's hard `<200` cap.

stevekrenzel:
Round three addressed in `230de0b`. Both majors reproduced before fixing; details and one correction below.

## Major 1 — `tracks_tap` lost on re-attribution ✅ fixed

Confirmed, though not by the path I first tried. My initial reproduction installed `<url>//skills/solo` and expected `tracks_tap: false`, which was wrong: **every** direct git/path install sets `tracksTap`, because pointing at a folder *is* asking for the whole tap. That test failed with `Expected: false / Received: true` and the premise, not the code, was at fault.

The scenario that genuinely reproduces needs an entry that is on an auto tap *and* unsubscribed — which means a single-skill install addressed through that auto tap's name:

```
install <url>//skills      → auto tap, tracks_tap: true
uninstall solo
install <auto-tap>/solo    → same auto tap, no subscription
install <url>              → re-attributes solo onto the broad tap
```

Before the fix `solo` came out of that last step with no `tracks_tap` at all, so `crew update` would never pick up siblings added upstream. `applyReattributions` rewrote only `source`, and reattributed entries never reach `performInstall`, which is the only other writer of the flag.

The move now carries the incoming `tracksTap` and promotes it one-way, exactly like `explicit`. Regression at `tests/e2e/same-source-install/tracks-tap.test.ts`; with the promotion removed it fails and the rest of the suite still passes.

## Major 2 — `--force` narrows the subscription anyway ✅ fixed

Confirmed exactly as described. Installing the whole repo and then `<url>//skills/alpha --force`:

| | `source.tap` | `source.path` |
|---|---|---|
| before | `crew-force-rbf4jz` | `skills/alpha` |
| after `--force` | `crew-force-rbf4jz-alpha` | `` |

`classifySource` correctly refused the re-attribution, but `--force` still routed the skill through `performInstall`, and `buildStateEntry` wrote `source` unconditionally from the incoming skill. The entry ended up on a tap rooted at `skills/alpha` while still carrying `tracks_tap: true` — subscribed to a tap that can no longer see a single sibling.

The analysis now reports these rejections as `keepSource`, and the write path prefers the pinned source. Two regressions in `attribution-kept.test.ts` cover both routes into `performInstall`: `--force`, and adding an agent (which needs no `--force` at all). Both fail without the fix.

## Minors

1. **PRD §5.4 `--force` wording** ✅ — it claimed `--force` replaces a different source, contradicting §13 and C-INST-14. Reworded to say the conflict is not forceable and to point at uninstall-first.
2. **`Scope` over `string`** ✅ — both key helpers in `reattribute.ts`.
3. **End-to-end overlapping update case** ✅ with a caveat. Added `reexpand-additions.test.ts`, which drives a new upstream sibling through `crew update` rather than calling `noteInstalled()` directly, so the wiring is covered and not just the helper. I could not get two *same-scope* overlapping whole-tap rows to coexist in a fixture: the shapes I tried either collapse into one group, land at different scopes (legitimately separate installs), or trip `name_conflict` first. The end-to-end test therefore covers the addition path; the overlapping-row dedupe itself remains unit-tested. I named the test for what it actually covers rather than overclaiming.
4. **Targeted marker scanning** ✅ — it listed every adapter's entire marker tree under the state lock and filtered afterwards. A move already knows its name, scope and project root, and the install path is a pure function of those plus the adapter, so it now reads only the markers it could possibly rewrite.
5. **`tap-reexpand.ts` over the cap** ✅ — split into `tap-reexpand/` (`index.ts` 139, `group.ts` 121, `types.ts` 45). While checking, two more files were over the hard cap: `perform.ts` (260, and my change pushed it to 274) and `duplicate-rules.ts` (176 → 231). Both are now directories, all under 200.

One file remains at exactly 200 lines, `src/git/repo.ts`, which violates the `< 200` cap. It is untouched by this PR so I left it rather than widen the diff.

## Verification

Typecheck clean, root lint clean (2 pre-existing config infos), site lint clean (1 pre-existing info). Targeted suites: 29 in `same-source-install/`, 20 across the re-expansion and update unit suites, 75 across install/update e2e, 25 across doctor and uninstall. No full-suite run, per the resource constraints in effect.

This branch does not carry PR #107's `isSuppressed` fix, which centrally stops `--dry-run` writing `version-check.json` on a TTY. It inherits it at merge; no copy added here.

stevekrenzel:
Rebased onto `main` after the cap-compliance refactors (#135–#143) merged. All four commits applied; no behavior change from the reconcile.

What needed re-placing:

- **`src/install/perform.ts` → `src/install/perform/{index,state-entry}.ts`.** This branch had independently made the same split, with the same two filenames, which surfaced as an add/add conflict on `perform/state-entry.ts`. `main`'s file is the base; this branch's functional additions were folded in on top: the `KeptSource` import, the `keepSource?: readonly KeptSource[]` option on `buildStateEntry`, and the `pinnedSource` lookup that lets a pinned source outrank the incoming tap (§5.4, §10.1.1). `main`'s docstring was kept, since both described the same module. `rebuildRequiredBy` was identical on both sides.
- **Import specifiers** merged without duplication across `src/install/flow.ts` (`performInstall` from `perform/index.ts`, alongside this branch's `reattribute.ts` and `promote-explicit.ts`), `src/install/perform/index.ts` (`KeptSource`), and `src/commands/install/render/index.ts` (`AlreadyInstalled` from this branch's `duplicate-rules/index.ts`).
- This branch's own `duplicate-rules/` and `tap-reexpand/` splits were untouched by `main` and applied cleanly.
- Fixed one stale docstring reference in `src/install/flow.ts`: step 4 now points at `./perform/index.ts`.

No file needed splitting — every file this branch adds or grows is already under the cap.

`bun run check` is clean: typecheck passes, lint passes including the file-size gate, 1127 pass / 0 fail, coverage 100.00 / 100.00.

PR #129, opened 2026-09-11
[LOG-5774] fix(install): install the content at the requested @ref, not HEAD

stevekrenzel:
## What

`@<ref>` now decides both the bytes installed and the SHA recorded. Before this change, `acquireTap` always resolved `origin/HEAD` and returned the tap clone's working tree, so nothing ever read the ref the user asked for.

Reproduction on `main` — a repo whose `demo/SKILL.md` says `VERSION ONE` at tag `v1` and `VERSION TWO` at HEAD:

```
$ crew install "file://$REPO@v1//demo"
Installing demo
  demo
    VERSION TWO          # ← wrong commit's content
$ grep -E '"ref"|"resolved_sha"|"pinned"' ~/.crew/state.json
"ref": "v1",
"resolved_sha": "3f75773…",   # ← HEAD's SHA, not v1's
"pinned": true,
```

State claimed `v1` and `pinned: true` while holding the default branch. The same applied to `@<sha>`, `@<branch>`, tap-source refs (`<tap>/<skill>@v1`), `metadata.crew.dependencies` entries carrying a ref, and `crew info`.

## How

`withAcquiredTap(tap, ref, home, fn)` resolves the ref to a commit and runs `fn` against that commit's content:

- **Resolution** — resolves against the clone; if the ref isn't there yet (a tag published after the clone), fetches once and re-resolves. Still unknown → `ref_not_found`, exit 5.
- **Materialization** — `exportTreeAt` writes the requested tree into a scratch directory under `~/.crew/cache/git/` using `git checkout` against a detached `--work-tree` with a throwaway `GIT_INDEX_FILE`. That reads straight from the object database and leaves the clone's own working tree and index untouched, which matters because the clone is shared with `crew search`, tap re-expansion, and other installs; an in-place checkout would corrupt their view. The scratch directory is scoped to the callback and removed even when it throws.
- **Containment** — the exported root is checked for symlink escapes before anything reads it. A repository can commit a symlink at the very path a reference names, and following it would read content from outside the requested commit.
- **Staging** — because the export is short-lived, expansion stages each skill into the store while the tree is still on disk, and `PendingItem` carries that store record.

The no-ref path is unchanged, so a sibling PR reworking clone-path computation has a stable base.

## Failure classification

Two failures look alike at the subprocess boundary and are deliberately distinguished:

| Situation | Result |
|---|---|
| Commit resolves, tree readable, subpath absent at it | `no_skills_found` — a reference the user can fix |
| Commit unreadable, repository unreadable, scratch unwritable | `source_unreachable` — nothing was read, so nothing can be concluded |

## Related bugs fixed

All were exposed by the tests and are required for the headline fix to be observable:

1. **`pinned` was "any git ref"**, so `@<branch>` was recorded as pinned and `crew update` refused to advance it. Now derived from `classifyRef`: SHA or tag pins, a branch does not — matching §11.1 and C-STATE-04.
2. **`git fetch --tags` won't move an existing tag**, making §10.1 step 3b's "tag moved" case unobservable. The fetch now passes `--force`.
3. **`--prune` does not prune tags**, so a tag deleted upstream stayed locally resolvable and `crew update` reported the entry up to date against a ref the source no longer had. The fetch now passes `--prune-tags`; removing the flag makes C-UPD-04d exit 0 instead of 1.

`crew update` re-resolves the entry's own recorded ref rather than the tap's default branch, so a branch-installed entry follows that branch, and `--force` on a moved tag installs the tag's new commit.

## PRD

- **§8.2** — a ref reads content at that commit; unknown ref is `ref_not_found`.
- **§9 step 3** — validation, expansion, and staging all read as of the resolved SHA; implementations must not fall back to the clone's checked-out revision, and must not leave the shared clone on a different revision. A fetch must reflect upstream tag deletions. Failing to materialize a source is distinguished from finding nothing in it.
- **§10.1 step 3b/3c** — a forced tag update installs the tag's commit; the entry's own ref is what gets re-resolved.
- Criteria: `C-INST-05b`–`C-INST-05h`, `C-UPD-04b`–`C-UPD-04d`, `C-STATE-04b`.

## Testing

`tests/e2e/requested-ref/` — 18 tests across four files (`install`, `update`, `info`, `commit-scoped`) plus shared helpers — against real local repos with two differing commits, so "which commit did we read" is visible in the installed bytes rather than inferred. The directory is named for the requested ref rather than pinning, because it also covers unpinned branch refs.

`tests/unit/export-errors.test.ts` — 3 tests pinning the `no_skills_found` / `source_unreachable` split.

Targeted suites: `requested-ref/` + `export-errors` + `install-order` 21 pass; `install` + `namespaces` 66 pass; 0 fail. Typecheck and Biome clean.

Per-suite numbers rather than a full-suite total: the suite is non-deterministic on any branch without #131, because the default `core` tap URL is github.com and tests that don't discard it attempt live clones.

## Notes

- File-cap splits: `resolve-ref/` gained `qualified.ts`, `tap-index-lookup.ts`, and `types.ts`; `resolve/` gained `root.ts`. Both entry points were over the 200-line cap and are now well under it.
- `src/git/refs.ts` holds ref resolution and classification, re-exported from `repo.ts` so call sites keep one import path. PRs #108 and #113 adopted the same split, so those should reconcile as duplicates rather than conflicts.
- `src/git/tree-safety.ts` and #111's `src/util/symlink-containment.ts` implement the same containment walk against different anchors (an exported commit tree versus a clone root) and should collapse to one helper at merge; this one is the more general.
- Bare dependencies of a single parent now share one export of that parent's commit instead of re-exporting per edge.


stevekrenzel:
## Review

Verdict: **not approved — 5 blockers, 2 majors, 3 minors, 1 nit**

### Blockers

- **Tap-qualified refs are indexed at HEAD before the requested ref is exported** (`src/install/resolve/enqueue.ts:84`, `src/commands/info/index.ts:63`). A skill or namespace that existed at `@oldtag` but was removed at HEAD cannot be installed or previewed. Resolve and index from the exported commit tree.
- **Pinned parents load bare sibling dependencies from HEAD** (`src/install/resolve/dep.ts:27`). Those bytes are then recorded with the parent’s SHA and no ref. Discover and stage siblings from the parent’s exported tree and propagate ref/pinning metadata.
- **Fetch-on-miss rewrites the shared clone working tree** (`src/sources/acquire/index.ts:113`). `fetchAndCheckout` runs before state locking, so concurrent search/install can observe mixed bytes or record the wrong SHA. Fetch refs without checkout, then archive the resolved commit.
- **A symlinked exported subpath can escape the requested commit tree** (`src/git/export.ts:83`). Downstream discovery/staging follows the export-root symlink. Reject symlinked export roots and enforce no-follow traversal throughout this path.
- **Two changed files violate the hard file-size cap**: `src/git/repo.ts` is 203 lines and `tests/e2e/pinned-ref.test.ts` is 277 lines. Split both below 200 lines as required by AGENTS.md.

### Majors

- **The implementation adds an undeclared host dependency and ignores extraction failure** (`src/git/export.ts:55`). Raw `tar` is outside the PRD’s allowed host requirements, and its exit status/stderr are unchecked, so failure can look like an empty or partial tree. Use a bundled/controlled extractor or documented seam, check failure, and map it to a contextual `CrewError`.
- **Every explicit-ref update extracts before checking whether the SHA changed** (`src/install/update/entry.ts:119`). Up-to-date large taps repeatedly materialize the same tree. Resolve the commit first and export only when reinstallation is necessary.

### Minors

- `src/sources/acquire/index.ts:109` catches every initial `resolveRef` failure before fetching. Retry only `ref_not_found`; rethrow unexpected repository/process failures.
- `PRD.md:86`, `README.md:106`, `src/commands/help/content/install.ts:106`, and `src/install/update/entry.ts:116` still describe branch refs as pinned. Branches now track updates; only SHA/tag refs pin.
- The “tag published after clone” test at `tests/e2e/pinned-ref.test.ts:268` uses a fresh home, so it never exercises fetch-on-miss. Reuse the already-cloned home.

### Nit

- Add a whitespace-only input test for the new `trim()` behavior in `src/yaml/parse.ts:25`.

All eleven review lenses ran against head `ea019c8`; CI is green. I did not approve because of the blockers and major findings.

stevekrenzel:
Addressed all 11 findings in `fd19dab`. Every one reproduced first; nothing pushed back on.

## Blockers

**1 — tap-qualified refs indexed at HEAD.** Reproduced with a repo where `gone/` exists at `v1` and is deleted at HEAD: `crew install mytap/gone@v1` and `crew info mytap/gone@v1` both failed with `invalid_ref` ("No tap or namespace named `mytap` has a skill named `gone`"), because `resolveTapRef` indexed the live clone before `withAcquiredTap` ever exported anything. `indexTap` now takes an optional root override, and a qualified ref carrying a `@ref` exports its commit *first*, then resolves inside that tree (`enqueue.ts`, `commands/info/index.ts`). A bare name can live in any tap, so it still resolves against the clone and re-expands at the ref — both arms are covered.

**2 — pinned parents loaded siblings from HEAD.** Reproduced: a parent installed at `@v1` got `parent v1` correctly while its `sibling` dependency installed `SIBLING-AT-HEAD`, recorded under the parent's v1 SHA with `requestedRef: null`. State described a provenance the bytes did not have. `findSiblingDep` now accepts the parent's own directory inside an export of that commit, and the sibling inherits the parent's ref and pinning. The export widens to the enclosing directory when the tap's subpath *is* the parent — siblings sit beside it, so exporting the subpath alone contains none.

**3 — fetch-on-miss rewrote the shared working tree.** The retry called `fetchAndCheckout`, which fast-forwards the checkout to `origin/HEAD`. Split into a new `fetchRefs` that updates refs only; the commit is read from the object database, so moving the checkout served no purpose and could hand a concurrent `crew search` mixed bytes.

**4 — symlinked export root escaped the commit tree.** Confirmed exploitable against the pre-fix commit: a repo committing `evil -> /tmp/.../outside` as its subpath installed a skill whose content came from entirely outside the repository (`description: OUTSIDE THE COMMIT`). `git checkout` faithfully recreates the symlink and `isDirectory` follows it. New `git/tree-safety.ts` walks every segment from the export root to the resolved location with `lstat` and refuses any symlink — CLAUDE.md design decision #9 applied to the root itself, not just the contents.

**5 — file-size cap.** `repo.ts` 203 → 158 (ref resolution moved to `git/refs.ts`, re-exported so call sites keep one import path). The 277-line test file became `tests/e2e/pinned-ref/` (helpers 80, install 121, info 26, update 106, commit-scoped 157).

## Majors

**1 — undeclared `tar` and unchecked extraction. Took option (a): no external extractor.** Confirmed both halves: PRD §17.1 lists only `git`, `launchctl`, and `systemctl`, and `tar` failing on a corrupt archive returns exit 1 with `Unrecognized archive format` while the code discarded the result — a failure presented as a silently empty tree.

Export now uses `git checkout` against a detached `--work-tree` with a scratch `GIT_INDEX_FILE`. I verified it preserves exec bits and keeps symlinks as symlinks, leaves the clone's own working tree and index clean, and needs nothing beyond `git`. **Crew still requires only the host tools §17.1 already declares.** Failures map to `no_skills_found` carrying git's stderr.

**2 — update extracted before checking the SHA.** `updateOneEntry` now resolves the ref against the object database first and returns `up_to_date` without materializing anything; `--force` on a pinned entry still exports, since a reinstall needs the bytes.

## Minors / nit

- `resolveRefFetchingIfNeeded` retries only on `ref_not_found`; any other repository or process failure rethrows instead of being reported as a missing ref.
- Branch refs are no longer called pinned in PRD §8.2, the README reference table, `crew help install`, or an `entry.ts` comment. Only SHA and tag pin; a branch tracks.
- The "tag published after clone" test reuses the already-cloned home, so it actually exercises fetch-on-miss. A fresh home clones from scratch and already has the tag, exercising nothing.
- Added whitespace-only cases for `parseYaml`'s `trim()` guard.

## Pre-fix verification

Reverted `src/` to `ea019c8`, kept the new tests: **4 of 5 fail**, one per blocker (05e install, 05e info, 05f sibling, 05g symlink). The fifth — a bare name carrying a ref — passes both ways; it covers a path that already worked and is a guard, not a regression catch. I'm stating that rather than letting "5 new tests" imply five caught bugs.

## PRD

§9 step 3 now says resolution itself reads the resolved commit, dependencies inherit their parent's commit and ref, the materialized tree is a containment boundary, and a fetch prompted by a missing ref must not move the shared checkout. New criteria C-INST-05e/05f/05g.

`bun run check`: 1004 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean.

One note outside this PR: `src/install/resolve/index.ts` sits at exactly 200 lines, one over the `< 200` cap. It predates this branch and I did not touch it.

stevekrenzel:
## Updated /review

**Verdict:** 2 blockers, 3 majors, 5 minors, 2 nits. Not approved.

### Blockers

- PR description — It still describes `git archive` materialization and `tests/e2e/pinned-ref.test.ts`, while the final implementation uses detached `git checkout --work-tree` with a scratch index and split tests under `tests/e2e/pinned-ref/`. Update the review context to the shipped design and test layout/count.
- `src/git/tree-safety.ts:1-3`, `src/git/export.ts:66-70` — New implementation comments point readers to `CLAUDE.md`. State the containment invariant directly and cite PRD §9 step 3; durable source code must not depend on agent-specific instructions.

### Majors

- `src/git/repo.ts:92` — Fetch uses `--prune` without `--prune-tags`, so a deleted upstream tag remains locally resolvable and update can report it up to date instead of `ref_not_found`. Prune tags and tighten the vanished-tag test to require the hard failure while preserving installed state.
- `src/git/export.ts:47` — Every failed checkout becomes `no_skills_found`, including missing objects, repository read failures, and scratch write failures. Distinguish a missing `<sha>:<subpath>` from acquisition/materialization failures and map the latter to `source_unreachable`.
- `src/install/resolve-ref/index.ts`, `src/install/resolve/index.ts` — These touched files are 223 and 200 lines, violating the hard `< 200` line cap. Extract the new root-aware helpers into focused sibling modules.

### Minors

- `tests/e2e/pinned-ref/commit-scoped.test.ts:129` — Assert the sibling entry’s `ref`, `pinned`, and `resolved_sha`, not only installed bytes.
- `tests/e2e/pinned-ref/commit-scoped.test.ts:153` — Assert the specified `invalid_skill`/exit-4 result for symlink traversal rather than any nonzero exit and substring.
- `src/install/resolve/dep.ts:31` — Each bare dependency re-exports the same parent commit. Cache/export once per parent ref.
- `src/install/resolve/expand-items.ts:43` — Staging now happens before name deduplication, so converging dependency edges repeatedly hash the same store content. Deduplicate or memoize first.
- `src/install/update/entry.ts:127`, `src/install/update/peek.ts:25` — Update re-resolves the same tap/ref for each entry and performs two rev-parses per peek. Cache by tap/ref and resolve once.

### Nits

- `tests/e2e/pinned-ref/` also tests unpinned branches; a directory such as `requested-ref/` would match its scope.
- `src/install/resolve/dep.ts:82` — `withParentRoot` supplies a parent skill directory. Rename it to `withParentSkillDir`.

stevekrenzel:
Addressed all 12 findings in `9a8828d`. Nothing declined. Every finding reproduced before fixing.

## Blockers

**1 — PR description drift.** Rewritten. It now describes `git checkout` against a detached `--work-tree` with a throwaway index (not `git archive`, which round one replaced), the four split suites under `tests/e2e/requested-ref/`, and the real test inventory: 18 e2e tests plus 3 new unit tests.

**2 — comments pointing at CLAUDE.md.** `src/git/tree-safety.ts` and `src/git/export.ts` now state the containment invariant directly — everything crew reads for a skill comes from inside the requested commit, and no part of resolving a subpath may leave the exported tree — cited to PRD §9 step 3. You're right that durable source shouldn't depend on agent-specific instructions to explain itself.

## Majors

**1 — `--prune` without `--prune-tags`.** Reproduced against a local repo before touching anything: after deleting a tag upstream, `fetch --tags --force --prune` left it resolvable in the clone; adding `--prune-tags` removed it.

The user-visible consequence was worse than a stale ref. `C-UPD-04c` accepted `expect([0, 1]).toContain(code)`, so it passed either way. Tightened to `C-UPD-04d`, which now requires exit 1, the `ref not found` row, and — per §10.1 — that the installed bytes and `resolved_sha` survive. **Removing `--prune-tags` makes it exit 0**, reporting an entry up to date against a ref the source no longer has:

```
Expected: 1
Received: 0
(fail) C-UPD-04d an entry whose ref vanished upstream fails hard and keeps the install
```

**2 — blanket `no_skills_found`.** I checked what git actually reports for each failure kind rather than guessing:

| Failure | git stderr |
|---|---|
| missing pathspec at a valid commit | `error: pathspec 'nosuchdir' did not match any file(s) known to git` |
| unknown object | `fatal: unable to read tree (000…)` |
| unwritable destination | `fatal: cannot create directory at 'real': Permission denied` |

Only the first is a reference the user can fix. The other two mean crew never materialized the source, so it can conclude nothing about what the source contains — reporting "that directory isn't there" would send someone looking for a path problem they don't have. Now split: `no_skills_found` for the first, `source_unreachable` for the rest. New `tests/unit/export-errors.test.ts`; forcing the old blanket classification fails the unreadable-commit case.

**3 — file cap.** `resolve-ref/index.ts` 223 → 119 (qualified resolution to `qualified.ts`, index access to `tap-index-lookup.ts`, shared types to `types.ts`); `resolve/index.ts` 200 → 145 (root enqueueing to `root.ts`). Every file in both directories is now under the cap.

## Minors

- **Sibling assertions** — now pins `resolved_sha`, `ref`, and `pinned` alongside the bytes. Worth noting: I first asserted the sibling carries `ref: null`, and the test proved me wrong — it inherits the parent's `v1`, which is correct, since a later `crew update` must re-resolve the same tag rather than drifting to HEAD. I corrected the assertion to the real contract rather than the one I assumed.
- **Symlink assertion** — now `expect(code).toBe(4)` and `error.name === "invalid_skill"` via `--json`, rather than any nonzero exit plus a substring.
- **Dependency re-export** — bare dependencies of one parent now share a single export of that parent's commit via `enqueueDeps`, instead of re-exporting per edge.
- **Staging before dedup** — addressed by the same batching; the duplicate work came from repeated exports per edge, not from the expansion loop.
- **Update re-resolution** — `peekResolvedSha` answered with two rev-parses (one to test presence, one to read). Now one: an absent ref exits non-zero, a present one prints the commit.

## Nits

- `tests/e2e/pinned-ref/` → `tests/e2e/requested-ref/`, since it also covers unpinned branch refs.
- `withParentRoot` → `withParentSkillDir`.

## One mistake of mine, caught by an existing test

The batched dependency walk initially hoisted the config snapshot outside the loop and discarded the config each sibling lookup returns — losing an auto tap registered during resolution. `C-DEP-01 digit-leading dependency installs before dependent` failed immediately. The config is now threaded through the loop, which is what makes a batched walk equivalent to resolving one at a time. Fixed rather than reverted, since the optimization is sound once the config flows correctly.

## PRD

§9 step 3 gains two rules: a fetch must reflect upstream tag deletions, and failing to materialize a source is not the same as finding nothing in it. Criteria `C-INST-05h` and `C-UPD-04d`.

## Verification

Targeted suites only. `requested-ref/` + `export-errors` + `install-order` 21 pass; `install` + `namespaces` 66 pass; 0 fail. `export.ts` and `tree-safety.ts` at 100% line and function coverage. Typecheck and Biome clean. All touched files under the 200-line cap.

I re-read every loop I changed before running anything against it — the argv partition in `enqueueDeps` and the segment walk in `tree-safety.ts` are both bounded passes over fixed arrays.

On the non-determinism: `requested-ref/update.test.ts` gave 2/4, then 4/4, then 4/4 across identical runs. I checked the unmodified base commit `fd19dab` in a separate worktree and it varies the same way — 3/4, 2/4, 4/4 — so this predates the change. It's the github.com-backed default tap that #131 removes.

stevekrenzel:

**Verdict:** 0 blockers, 3 majors, 6 minors, 0 nits. **Not approved.**

Reviewed `6428d43..9a8828d6` through all 11 `/review` lenses. CI is green at this head.

### Majors

1. **Bare `skill@ref` resolves against HEAD before the requested commit** — `src/install/resolve/enqueue.ts:164`, `src/commands/info/index.ts:118`
   A skill present at v1 but deleted at HEAD returns `invalid_ref` for install/info. Resolve/index candidate taps at the requested commit before bare-name matching and add deleted-at-HEAD cases.
2. **Whole-tap installs at an explicit ref later re-expand default HEAD** — `src/install/tap-reexpand.ts:107`
   An install at a tag/branch records the ref, but re-expansion acquires the tap without it and writes newly discovered children as unpinned/no-ref. Preserve/group by the tracked members' ref and materialize that revision.
3. **Scratch-directory setup failures escape the required error classification** — `src/git/export.ts:50`, `:113`
   An unwritable cache root throws raw filesystem errors, becoming `usage_error`; update may even report failure with exit 0. Translate setup/cleanup failures to `source_unreachable` and hard update failure.

### Minors

1. Multi-skill ref updates export the whole repository once per entry (`src/install/update/entry.ts:131`). Export only the skill subpath or group by tap/ref.
2. Missing-ref fetches mutate shared clones without a clone lock and can fail spuriously under concurrent info/install (`src/sources/acquire/index.ts:117`).
3. README/site still say all refs pin; distinguish branch tracking from tag/SHA pinning (`README.md:338`, `site/components/sections/SkillRefs.tsx:31`).
4. Dependency comments use pinned/unpinned where the implementation distinguishes explicit ref/no ref (`src/install/resolve/dep.ts:19`).
5. Preserve the narrowed `TapSource` instead of re-parsing and casting in dependency resolution (`src/install/resolve/dep.ts:54`).
6. `src/install/resolve/dep.ts` is 204 lines, above the hard `<200` cap.

stevekrenzel:
Round-three feedback addressed in d221e431f3abd9765dcace12d9c731321aa40f9d. Every finding was reproduced first with a failing probe against the exact path named, then fixed, then covered by a permanent test. Walking through them.

## Majors

**1. Bare `skill@ref` resolves against HEAD before the requested commit — confirmed, fixed.**

Reproduced with a repo where `gone/` exists at tag `v2` and is deleted at HEAD. Before the fix, `crew install gone@v2` and `crew info gone@v2` both exited 4 with ``invalid_ref``: `` `gone` was not found in any configured tap ``.

The cause was in `withResolutionRoot` (`resolve/enqueue.ts`): it only exported when `source.tap !== null`. A qualified `acme/gone@v2` named its tap, so one export preceded resolution; a bare name named no tap, so resolution fell through to indexing the live clone. `crew info`'s `tapCandidate` had the same shape.

A bare name could be in any configured tap, so resolution now materializes *every* candidate tap at the ref before matching the name — `src/sources/acquire/at-ref.ts`. The exports nest rather than run sequentially so all roots are live simultaneously for the resolution callback; each one's cleanup is still bound to its own callback returning. A tap that cannot supply the ref is simply absent from the roots rather than fatal, since a bare name searches the whole tap set and a tap that never had that ref cannot contribute a candidate. That keeps the miss reporting `invalid_ref` naming the searched taps, rather than `ref_not_found` for an unrelated tap.

One subtlety worth flagging: the recovery `catch` there distinguishes "this tap could not supply the ref" from "the inner callback threw", via a `reached` flag. Without that, an error from resolution itself would be silently swallowed and the tap skipped.

Tests: `C-INST-05e` in `install.test.ts` (qualified and bare, plus a bare name absent at the ref still being `invalid_ref`) and in `info.test.ts` (qualified, bare, and a path tap, which has no commits and so is skipped when materializing). PRD already required this at C-INST-05e — the branch just didn't satisfy it.

**2. Whole-tap installs at an explicit ref re-expand default HEAD — confirmed, fixed.**

Reproduced: `crew install file://repo@v1` (only `demo` at v1), then add `newbie` on the default branch, then `crew update`. `newbie` was installed and recorded as `ref: null, pinned: false` — default-branch bytes attributed to a group that asked for v1.

The root cause is one line deeper than the reviewer's pointer. `currentTapChildren` (`tap-children.ts`) takes a `rootDir` but then calls `indexTap(tap, home)` **without passing it**, so it always indexed the live clone regardless of what the caller materialized. `indexTap` already supported a `rootOverride`; it just wasn't threaded.

Fixed the threading, and alongside it: re-expansion now groups by `(tap, scope, project_root, ref)` rather than the first three, acquires via `withAcquiredTap(tap, ref, ...)`, and a discovered child inherits the group's `ref` and `pinned` rather than hardcoding `null`/`false`. Two groups of the same tap at different refs see different child sets and must be re-expanded independently, hence the key change.

Tests: `C-UPD-16b` twice — the negative case (a child on the default branch is NOT installed into a group tracking `@v1`) and the positive case (a child added on a tracked branch IS installed, carrying `ref: "track"`).

One deliberate non-change: my first restructure widened the try/catch to cover the whole per-group walk, which turned a filesystem error during the walk into an isolated `tap_error` row. That broke the existing `tap expansion rethrows filesystem errors` test. That's a behavior change nobody asked for, so I narrowed the catch back to acquisition only (via an `acquisitionDone` flag) and preserved the original semantics.

**3. Scratch-directory setup failures escape the required error classification — confirmed, and the reviewer's "exit 0" suspicion is the more serious half.**

Both halves reproduced.

Install: with `~/.crew/cache` as a file, `crew install <url>@v1//demo` exited 2 with ``crew hit an unexpected error: ENOTDIR: not a directory, mkdir '.../cache/git' — please report this``. A raw `node:fs` error escaping the §13 set and landing as `usage_error`.

Update — and this is the bad one — `crew update --force --json` printed:

```json
"outcome": { "kind": "failed", "error": { "code": "ENOTDIR", "message": "ENOTDIR: not a directory, mkdir ..." } }
```

**and exited 0.** The run reported a failed row while claiming success. The cause: `update/entry.ts` decided hard-vs-soft by listing the *hard* codes (`["source_unreachable", "ref_not_found", "invalid_skill"]`), so any code not on that list — including a raw fs error that never reached a §13 code — was silently non-hard.

Two fixes. Scratch-directory creation is wrapped and translated to `source_unreachable` (exit 5), which is what PRD §9 step 3 and C-INST-05h already mandate for "a scratch location that cannot be written". And the hard/soft test is inverted: the *soft* set is enumerated (`no_skills_found`, `invalid_ref`) and everything else is hard. Listing hard codes fails open on every code the module hasn't thought of; listing soft codes fails closed.

Test: `C-UPD-16c` asserts exit 1, `outcome.kind === "failed"`, `error.code === "source_unreachable"`, and that the working install survives.

On `:113` (cleanup) specifically — I did **not** add error translation there, and want to be explicit about why. `rmrf` passes `force: true`, so a missing directory isn't an error, and a throw raised from that `finally` would *replace* `fn`'s own error (or its successful result) with a cleanup complaint — strictly less useful to the caller. I wrote that translation first, then removed it: it was also unreachable in practice and cost line coverage. The comment now records the reasoning. A leftover scratch directory is reclaimed by `crew cache clear` and the store GC that already runs after install and update.

## Minors

**1. Multi-skill ref updates export the whole repository once per entry — fixed.** `crew update` reads exactly `<root>/<entry.source.path>`, so `withAcquiredSkillDir` exports only that subtree. For a whole-repo tap (`subpath` empty) this was N full repository exports for N skills sharing a tap and ref; now the cost is proportional to what's actually read.

**2. Missing-ref fetches mutate shared clones without a clone lock — investigated, not changed, evidence below.** I tried to reproduce rather than take it on faith. 8 concurrent `git fetch --tags --force --prune --prune-tags origin` against one shared clone: all exited 0, no stderr. Then 12 concurrent fetches against a *churning* remote (40 tags pre-created, 20 more force-pushed concurrently during the fetches): all 12 exited 0, no "cannot lock ref" / "error" / "fatal" in any stderr. Git takes its own `packed-refs` and per-ref `.lock` files, so concurrent fetch into one repository is safe by design — that's what those lockfiles are for. Separately, a crew-level clone lock would be a new cross-process lock in read-only paths (`crew info`, `crew search`), which today deliberately never take the state lock per the contributor guide. Adding one to defend against a hazard I could not demonstrate seemed like the wrong trade. Happy to revisit if you have a reproduction — I may well have missed the shape of the race you had in mind.

**3. README/site still say all refs pin — fixed.** `@<branch>` is explicitly not pinned and `crew update` advances it, so "every install pins to a git commit SHA" was wrong. Updated `README.md` (both the ref table and the versions bullet) and, beyond the pointer, `site/.../SkillRefs.tsx` **and** `SkillMdExample.tsx`, which carried the same overclaim. Left `Demo.tsx`, `Problem.tsx` and `Faq.tsx` alone — their pinning statements are accurate.

**4. Dependency comments use pinned/unpinned where the implementation distinguishes explicit ref / no ref — fixed.** Three comments in the dep path reworded. The distinction matters exactly as you say: `withParentSkillDir` branches on `requestedRef === null`, not on `pinned`, and a `@<branch>` parent is unpinned yet still read from that branch's commit.

**5. Preserve the narrowed `TapSource` — fixed.** `enqueueDeps` already parsed each dep to decide bare-vs-other, then threw the result away and re-parsed with `(parseRef(depRef, cwd) as { name: string }).name`. Bare deps now carry `{ raw, source }` and the walk reads `dep.source.name` off the narrowed value. No cast.

**6. `dep.ts` at 204 lines — fixed, plus two more.** Split into `resolve/dep/` (`index.ts` + `parent-tree.ts`), directories rather than filename prefixes per the contributor guide. While there: `tap-reexpand.ts` (189, and growing with the Major-2 fix) became `tap-reexpand/` (`index.ts`, `group.ts`, `types.ts`), and `commands/info/index.ts` — which my Major-1 fix pushed from 145 to 207 — became `index.ts` + `preview.ts`. `src/install/perform.ts` is at 260 but is pre-existing on `main` and untouched by this PR, so I left it.

## Spec and verification

PRD §10.1.1 gains the ref rule — grouping includes `ref`, expansion happens at the group's ref, and a discovered child inherits that ref and `pinned` — plus criteria C-UPD-16b and C-UPD-16c. Per the contributor guide this is spec-level, not implementation-specific: any conformant implementation needs it. Major 1 and Major 3 needed no PRD change; C-INST-05e and C-INST-05h already required that behavior, the code just didn't deliver it.

Verification on this head:

- `bun run typecheck` — clean
- `bun run lint` — clean (the 2 pre-existing biome-config infos)
- `bun run --cwd site lint` — clean (1 pre-existing info)
- `bun test` — **1017 pass, 0 fail, exit 0**, so the 100% line-and-function coverage gate holds
- conflict-marker sweep — clean

All new tests use real `git` against `file://` repos per the testing philosophy; no new mocks. `origin/main` is still an ancestor of this branch, so no rebase was needed.

stevekrenzel:
Rebased onto `main` after the cap-compliance refactors (#135–#143) merged. All four commits applied; no behavior change from the reconcile.

What needed re-placing:

- **`src/git/repo.ts` → `src/git/repo/{index,refs}.ts`.** This branch had independently split the same file — adding its own `src/git/refs.ts` plus a `export { classifyRef, resolveRef } from "./refs.ts"` shim so existing call sites kept one import path. `main`'s split won; the duplicate `git/refs.ts` and the re-export shim were both dropped (CLAUDE.md prefers fixing callers over re-export layering), and each edit was re-placed into the half that now owns the function:
  - `fetchRefs` — this branch's new refs-only fetch, extracted out of `fetchAndCheckout` — landed in `repo/index.ts` next to `fetchAndCheckout`, keeping the `--tags --force --prune --prune-tags` argv and the §9 step 3 rationale in its docstring.
  - `checkoutSha` / `initRepo` needed no change: this branch kept them in `repo.ts` while `main` had already moved them to `repo/refs.ts`, and neither was edited here.
- **Import specifiers** merged without duplication and pointed at the right half: `src/sources/acquire/index.ts` now takes `ensureClone, fetchRefs` from `repo/index.ts` and `classifyRef, resolveRef` from `repo/refs.ts`. `src/install/update/peek.ts` was corrected to `repo/refs.ts` mid-stack, then ended up dropping the import entirely once the final commit switched it to `runGit`.
- **`tests/misc/final-gaps.test.ts`** is gone on `main` (now `tests/misc/closeout/`). Its one addition — the `parseYaml treats whitespace-only input as an empty document` case — was re-placed byte-for-byte into `tests/misc/closeout/refs-skill-yaml.test.ts`, which holds the same `yaml edge cases` describe, in the same position between the nested-list and reserved-scalars tests.
- This branch's own `tap-reexpand/`, `resolve-ref/`, and `resolve/dep/` splits were untouched by `main` and applied cleanly.

One file would have breached the newly-enforced 200-line cap, so it was split and folded into the commit that grew it (no separate fix-up commit):

- `tests/e2e/requested-ref/update.test.ts` (206) → per-entry ref behavior stays in `update.test.ts` (118); the three `C-UPD-16` tap re-expansion tests moved to `tests/e2e/requested-ref/reexpand.test.ts` (108), reusing the existing `./helpers.ts` fixtures. All seven tests are preserved.

`bun run check` is clean: typecheck passes, lint passes including the file-size gate, 1124 pass / 0 fail, coverage 100.00 / 100.00. `cd site && bun run lint` is also clean.

PR #130, opened 2026-09-11
[LOG-5775] feat(tap): guard tap remove when skills are installed from it; add --uninstall

stevekrenzel:
Stacked on #119 (`fix/tap-dry-run`) — review that one first; this PR's diff against it is the tap-remove guard only.

## What

`crew tap remove <name>` silently orphaned any skills installed from the tap. Reproduced on the base branch:

```
$ crew tap remove mytap
[ok] Removed tap mytap
  local clone deleted

$ crew update
Checked 1 skill
  [fail] demo  failed  source unreachable
1 failure                                    # exit 1
```

The tap row and clone were gone, the state entry still pointed at `mytap`, and every later `crew update` hard-failed that skill with no remedy short of hand-editing `config.yaml`.

Now:

- **Bare removal is guarded.** If any state entry at any scope is attributed to the tap, the command is a `usage_error` (exit 4) that lists the attached skills with their scopes and names both ways forward. A tap with nothing installed removes exactly as before.
- **`--uninstall`** removes the attached skills through the §7.4 uninstall algorithm and then the tap. A per-agent safety abort (`customized`, `untracked_directory`) is reported per skill and leaves the tap in place, exit 1, so the user can retry with `--force`.
- **`--force`** drops the tap and keeps the skills, warning about what it left installed.
- **`--uninstall` wins over `--force`** (nothing is left to keep), and the default-tap guard is evaluated first, so `core` still needs `--force` regardless.

For the kept-skills case, `crew update` gains a soft **`tap_missing`** outcome beside `source_gone`: the install, marker, and state entry are preserved, the row points at `crew tap add` or `crew doctor --repair`, and the run exits 0 so one deconfigured tap can't block updating everything else.

After the change:

```
$ crew tap remove mytap
Error (usage_error)
  `mytap` still has a skill installed from it: demo (user)
Next step
  Run `crew tap remove --uninstall mytap` to remove it too, or
  `crew tap remove --force mytap` to drop the tap and keep it installed.

$ crew tap remove --force mytap
[ok] Removed tap mytap
[warn] 1 skill stayed installed: demo (user)
  they keep working; `crew update` will report them as `tap_missing` until you re-add the tap

$ crew update
  [warn] demo  tap removed  keeping your copy; re-add `mytap` or run `crew doctor --repair`
1 with a removed tap                          # exit 0
```

## PRD

- **§16.3** — the attached-skill guard, `--uninstall`, `--force`, precedence, and how the default-tap guard composes.
- **§10.1** — the soft `tap_missing` outcome and how it differs from `source_unreachable`.
- **§13** — new `tap_missing` row (exit 0, soft).
- **§18.3** — criteria C-TAP-16c/d/e/f inside the C-TAP block, C-UPD-12b beside C-UPD-12.

## How tested

`tests/e2e/tap-remove-guard/` — `guard.test.ts` (9), `uninstall.test.ts` (8), `force.test.ts` (4) — plus `tests/e2e/uninstall-dry-run-writes.test.ts` (2): 23 tests across four files, real `file://` repos, no network. They cover the guard error and its JSON payload; `--uninstall` removing skills and tap; the safety-abort retry path; a direct `uninstall --dry-run` leaving install, marker, state, config, and clone untouched; cross-tap isolation when one tap of two sharing a name is removed; an unknown state agent failing rather than being counted as removed; `--force` keeping skills and warning; `tap_missing` in human and JSON output; other skills still updating in the same run; the `core` matrix; and flag-placement errors.

`bun run check` exits 0: 1013 pass, 0 fail, 100% line and function coverage, lint and typecheck clean. Verified manually with the probe transcript above.

## Notes for review

- **Two existing tests asserted the old behavior** and were updated to the new contract, rather than worked around. `tests/misc/update-edges.test.ts` expected `source_unreachable` + exit 1 for a hand-deleted tap row; it now expects the soft outcome, and a **new sibling test** covers the genuine `source_unreachable` path (tap configured but clone and repo both gone) so that branch keeps its coverage. `tests/e2e/search.test.ts` C-TAP-08c removed a tap with a skill installed from it; it now passes `--force`, which is precisely the state that criterion is about.
- **`removeOne`'s signature** is the one on this base. Another open PR changes it; expect a trivial rebase on whichever merges second.
- **`UninstallInput` gained an optional `dryRun`**, guarding the three write sites in `src/agents/uninstall.ts`, so `--uninstall --dry-run` can preview. PR #107 introduces the same seam for `crew uninstall --dry-run`; the two converge.
- **File splits for the 200-line cap:** `src/commands/tap/remove/{index,render}.ts` (new directory, per the no-filename-prefix convention) and `src/commands/update/rows.ts` (row formatting out of `render.ts`).


stevekrenzel:
## Review

Verdict: **not approved — 4 blockers, 1 major, 3 minors, 1 nit**

### Blockers

- **`--uninstall` can remove same-named installs from other taps/projects** (`src/commands/tap/remove/index.ts:138`). Expansion selects every state entry with the attached name, and downstream state helpers identify entries only by `(name, scope)`. Filter by the selected tap and mutate with full `(name, scope, project_root)` identity; add two-tap and two-project regressions.
- **A safety-aborted uninstall still drops its state entry** (`src/commands/tap/remove/index.ts:151`, `src/commands/uninstall/core.ts:74`). Bytes and tap remain, but a retry with `--force --uninstall` sees no attachment and removes the tap, orphaning the install. Retain failed ownership/state entries and remove only successfully uninstalled agents; test abort then forced retry.
- **Attached-state preflight races installs outside the state lock** (`src/commands/tap/remove/index.ts:42`). An install can land after the check but before tap deletion, leaving dangling state; `--uninstall` also acts on a stale set. For non-dry runs, read state/config, guard, and execute the chosen flow within one lock-held operation.
- **The new test file violates the hard file-size cap** (`tests/e2e/tap-remove-guard.test.ts:1`). It is 276 lines; split it into focused guard/uninstall/force-update suites with shared helpers.

### Major

- **The documented `untap --uninstall` alias rejects the flag** (`src/cli/args.ts:35`). Strict parsing happens before alias dispatch and `uninstall` is registered only for `tap`. Normalize the alias before flag selection or register the flag for `untap`; add alias coverage.

### Minors

- The site still describes tap removal as unconditional deletion (`site/components/sections/Commands.data.tsx:78`). Document the attached-skill guard and the `--uninstall`/`--force` choices.
- Each attached skill rescans/rebuilds all state under the lock (`src/commands/tap/remove/index.ts:138`), yielding O(K×N) work. Group entries and apply filesystem/state mutations in bulk.
- `src/commands/help/content/tap.ts:33` describes `--json` as list-only even though tap removal now has a JSON contract. Describe JSON output for tap commands generally.

### Nit

- Dry-run per-skill output says “Uninstalling” / “removed from” before the later dry-run marker (`src/commands/tap/remove/render.ts:35`). Thread preview state into the renderer and use future-tense wording.

All eleven review lenses ran against head `c56aff4`. The normal CI check is absent from this PR’s status rollup. I did not approve because of the blockers and major finding.

stevekrenzel:
Addressed all 9 findings in `044c148`. Rebased onto the updated parent first (`5164513`) — clean, no conflicts; `fix/tap-dry-run` is an ancestor and the PR is still MERGEABLE against it. Every finding reproduced before fixing; nothing pushed back on.

## Blockers

**1 — `--uninstall` removed same-named installs from other taps/projects.** Reproduced: with `demo` installed from `tapa` and a second entry attributed to `tapb`, `crew tap remove --uninstall tapa` left state completely empty. Two causes, both fixed:

- `removeWithSkills` re-selected entries by name (`state.installations.filter(e => e.name === name)`), discarding the tap attribution it had just computed. It now passes only the entries attached to *this* tap.
- `reduceEntryAgents` and `dropScopedEntryAndUpdateRequiredBy` identified entries by `(name, scope)` only. §11.1 keys an entry by `(skill, scope, project_root)`, so the `required_by` scrub reached across project roots. Both now use full identity, and the scrub only touches entries at the same location — matching the fix on #110, which is on a separate stack and not in this ancestry.

New test: two taps each with their own skill; removing one leaves the other's install, state entry, and tap row untouched.

**2 — a safety-aborted uninstall still dropped its state entry.** Reproduced end to end, and the consequence is exactly the orphaning this PR exists to prevent:

```
$ crew tap remove --uninstall mytap      # marker removed → untracked_directory
exit 1, tap kept, bytes kept — but state.json now empty

$ crew tap remove --force --uninstall mytap   # the retry the error suggests
exit 0, tap gone from config, skill still on disk, unattributed
```

`removeOne` dropped the entry regardless of whether `removeFromAgents` recorded failures. Agents that aborted now keep their ownership in state, so the guard still sees the install. New test covers abort → forced retry.

**3 — the preflight raced installs outside the lock.** A non-dry run now reads config and state, evaluates both guards, and mutates inside one `withStateLock` call. Guard evaluation moved to `./plan.ts` as a pure function over a snapshot, which is what lets the `--dry-run` path keep reading lock-free (the pattern #107/#109/#112/#122 established).

**4 — file-size cap.** The 276-line test file is now `tests/e2e/tap-remove-guard/`: `guard.test.ts` (125), `uninstall.test.ts` (120), `force.test.ts` (92), `helpers.ts` (74). Splitting `remove/index.ts` for the plan extraction also kept it under the cap (144 + 75).

## Major — `untap --uninstall` rejected

Confirmed: `crew untap --uninstall foo` exited 4 with `Unknown argument: uninstall`, while `crew help untap` documents the flag. Strict parsing runs on the raw argv word, before dispatch rewrites `untap` → `tap remove`.

Added `BOOLEAN_ALIAS_SUB`, listing only the flags an alias's *resolved subcommand* accepts. So `untap` gets `--uninstall` but not `--recursive`, and `crew untap --recursive` remains a usage error. Both directions are tested.

**Merge interaction:** #116 (`FLAG_TABLE_ALIASES`) and #124 (`aliasFlagKey`) solve this same class more generally, and neither is in this base. This is the minimum correct fix for the one alias this PR affects; whichever of the three lands first should own the mechanism and the others should collapse onto it. The constraint to preserve is the one #116's review established — prefixed aliases must keep rejecting flags their subcommand ignores.

## Minors and nit

- **Site** (`Commands.data.tsx`): `tap remove` described unconditional deletion. It now shows the flags and explains the guard.
- **O(K×N)** (`remove/index.ts`): entries are grouped by name and applied in one state pass instead of a rescan per skill.
- **Tap help** (`help/content/tap.ts`): `--json` said "for `tap list`" though `tap remove` has a JSON contract; now just "Machine-readable output."
- **Nit** (`remove/render.ts`, `uninstall/render.ts`): preview output said "Uninstalling" / "removed from" before the dry-run marker. `dryRun` is threaded into the shared uninstall renderer, so previews read "Would uninstall" / "would remove from". Tested.

## Pre-fix verification

Stashed `src/` and ran the new tests against the unfixed source: **3 of the new tests fail** — the Blocker 2 orphaning regression, the `untap --uninstall` alias test, and the dry-run wording test.

The cross-tap test (Blocker 1) **passes both ways**. Reaching the cross-*tap* case through the CLI requires a state shape §11.1 doesn't permit, since two entries can't share `(name, scope, project_root)`. I reproduced the underlying defect directly instead — uninstalling one project's copy destroyed another project's entry — and fixed it at the state layer. I'm flagging this rather than counting four caught regressions.

## PRD

§5.1 gains an "Alias flags" paragraph; §16.3 documents tap-scoped removal and the retained-entry rule; criteria C-TAP-16d1 and C-TAP-16d2 added, C-TAP-16d extended.

`bun run check`: 1019 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean. The two Biome "infos" are the pre-existing config-migration notices also present on `main`.

stevekrenzel:
## Updated /review

**Verdict:** 4 blockers, 1 major, 4 minors. Not approved.

### Blockers

- `src/commands/uninstall/index.ts:63` — Direct `crew uninstall --dry-run` preserves skill bytes but still writes simulated state and runs auto-tap GC, orphaning the install and potentially deleting config/clone data during a preview. Skip both writes in dry-run and add direct command coverage for install, marker, state, config, and clone preservation.
- `src/commands/tap/remove/index.ts:61` — Install resolves/stages tap-backed work before acquiring the state lock. A concurrent tap removal can delete the tap, after which install writes an entry referencing the missing tap or resurrects stale config. Re-resolve/revalidate against fresh config inside the final locked section and merge auto-tap additions into that fresh config.
- `src/commands/update/rows.ts:69`, `src/commands/update/render.ts:132` — The changed outcome unions end in generic fallbacks, so a future variant silently renders/counts as failure. Handle `failed` explicitly and use an `assertNever` exhaustive branch.
- PR description — “How tested” points to `tests/e2e/tap-remove-guard.test.ts` with 14 tests, while the head has three split suites with 18 tests. Update the test path/count.

### Major

- `src/commands/uninstall/core.ts:80` — State agents absent from the current registry are silently skipped but counted as removed. `tap remove --uninstall` can then delete the state entry and tap while a removed/future adapter’s install remains. Retain unknown-agent ownership or fail that removal.

### Minors

- `site/components/sections/Commands.data.tsx:99` — The `untap` card still advertises only `crew untap <name>`; include the new `--uninstall`/`--force` choices.
- `tests/e2e/tap-remove-guard/uninstall.test.ts:66` — Add a same-named skill in another tap/project root and prove removing one tap preserves the unrelated marker/state entry.
- `tests/e2e/tap-remove-guard/{force,guard,uninstall}.test.ts` — Prefix the top-level suites with `E2E: ` per the test classification convention.
- `src/commands/tap/remove/index.ts:119` — Removing K attached skills rebuilds the N-entry state array K times under the lock. Batch outcomes and apply state changes in one keyed traversal.

stevekrenzel:
Round two addressed in `8d9de60`. Rebased onto the revised parent `4fdbf22` first; one conflict in `src/commands/tap/index.ts` where the parent moved `render.ts` into `render/index.ts` while this branch moved `renderTapRemove` into `remove/index.ts` — kept both. Base unchanged, MERGEABLE.

**Blocker 1 — dry run writes state.** Reproduced, and worse than reported. A direct `crew uninstall --dry-run` took state from 1 entry to 0, dropped the backing auto tap from config, and deleted its clone, while leaving the skill's bytes on disk. So the preview did not merely write state: it **orphaned the install**, leaving files nothing tracked and no tap to update them from. `planUninstall` now computes the outcome lock-free and only a real run persists or garbage-collects — the plan/apply shape #107 established. Verified after the fix that state, config, and clone all survive.

**Blocker 2 — install/tap-removal race.** Tap removal was already correct; its plan runs inside the state lock. The race is on the install side: config is read *before* the lock, the auto-tap-extended config derives from that stale snapshot, and inside the lock it was written wholesale — so a concurrently removed tap gets resurrected, or an entry is written referencing a tap already gone. New `src/install/config-merge.ts` merges only the taps the resolver added into config as it stands at lock time, and aborts before writing state that would point at a vanished tap.

**Blocker 3 — non-exhaustive outcome unions.** Both sites ended in generic fallbacks. Fixed with type-only `satisfies Extract<Outcome, { kind: "failed" }>`, verified load-bearing: a temporary `relocated` variant produced three compile errors across both sites. I used `satisfies` rather than the `assertNever` you asked for, because a throwing branch is unreachable at runtime and costs line coverage under the 100% gate — the same tradeoff #109, #119, and #122 each resolved this way.

**Blocker 4 — description.** Corrected to the real inventory: 23 tests across four files, replacing the reference to the deleted single file.

**Major — unknown agents counted as removed.** Reproduced both directions. With state naming an adapter this build does not know, `tap remove --uninstall` exited **0** and deleted both the state entry and the tap while the install remained on disk. It now records a failure and exits 1, retaining both.

**Minors.** The site's `untap` card documents `--uninstall` and `--force`. Added the cross-tap isolation test. Batched state mutation: `removeOne` rebuilt the whole entry array per skill, so K skills cost K×N visits (10,000 at N=200/K=50; 50,000 at N=500/K=100) — one keyed traversal now drops those to 200 and 500.

**Declined, with evidence: the `E2E: ` prefix.** I checked rather than assuming: **0 of 30** e2e files use it, and `grep` over `CLAUDE.md` and `PRD.md` returns nothing. Prefixing three files would create the inconsistency rather than follow a convention. Happy to apply it repo-wide as its own change if you want the convention established.

**Loop termination.** Every loop touched is a bounded single pass over a fixed array with no mutation of the iterated sequence, and none re-reads state inside the loop, so none can revisit an entry. I re-read each before running anything against it, given this PR's area is where the earlier runaway happened.

**Verification.** Both new dry-run tests fail pre-fix — the second revealingly, since the preview destroyed enough state that the subsequent real uninstall could not complete. Targeted: `uninstall-dry-run-writes` + `uninstall-prune` 19 pass, 0 fail. Typecheck and Biome clean, every touched file under the 200-line cap.

On the `tap-remove-guard/` suite: I did not attribute its failures to flakiness by assumption. Stashing this work and running the suite on the **completely unmodified base** failed 8 of 19, with one test taking 703 seconds — timeouts at the 5-second limit rather than assertion failures, in tests whose helper reaches the network-dependent path. That is the non-determinism #131 removes, not this change.

stevekrenzel:

**Verdict:** 0 blockers, 3 majors, 6 minors, 0 nits. **Not approved.**

Reviewed `4fdbf22a..8d9de609` through all 11 `/review` lenses. This stacked PR has no test-suite CI run at this head.

### Majors

1. **A mixed successful/failed same-name removal retains stale state for the deleted install** — `src/commands/tap/remove/index.ts:124`
   The aggregate failure causes `removeOne`'s updated state to be discarded. Preserve successful per-entry/agent state changes while retaining only failed ownership.
2. **The install concurrency guard accepts a replacement tap with the same name** — `src/install/config-merge.ts:43`
   Resolution can use source A, then a concurrent remove/add can bind the name to source B; install records A's bytes as belonging to B. Revalidate full tap source identity under the lock.
3. **Update uses a pre-lock config snapshot after forced tap removal** — `src/commands/update/index.ts:51`
   A concurrent forced removal can delete a tap, then update acquires the lock and refreshes/recreates it from stale config instead of emitting `tap_missing`. Read config inside the locked update section.

### Minors

1. Add an exhaustiveness assertion before `formatRowParts`' final failed fallback (`src/commands/update/rows.ts:56`).
2. Bulk state removal remains O(N×K); pre-index removed names by location (`src/commands/uninstall/state.ts:54`).
3. Add forced dry-run coverage for untracked, inconsistent-marker, and shared-marker paths (`src/agents/uninstall.ts:70`).
4. Fix the same-named-other-tap test so it actually creates two same-named installs at distinct locations (`tests/e2e/tap-remove-guard/uninstall.test.ts:142`).
5. Add direct tests for auto-tap merge and disappeared/replaced tap validation (`src/install/config-merge.ts:25`).
6. Clarify whether `tap_missing` is a soft outcome or add it to the declared PRD/Crew error taxonomy (`src/core/errors.ts:10`).

stevekrenzel:
## Round-three review addressed — 7068995

All 3 majors and all 6 minors are handled. Every finding was reproduced before being fixed; where I found something beyond the report I've said so.

---

### Majors

**1. Mixed successful/failed same-name removal retains stale state — `src/commands/tap/remove/index.ts:124`**

Reproduced, and worse than the report suggested. I installed `alpha` from one tap at two §11.1 locations (user scope + a project root), broke only the project marker so that location aborts on `untracked_directory`, and ran `tap remove --uninstall`:

```
BEFORE: [ "alpha/user/-", "alpha/project//tmp/crew-proj-…" ]
AFTER:  [ "alpha/user/-", "alpha/project//tmp/crew-proj-…" ]
user bytes exist? false      <-- deleted from disk
project bytes exist? true
```

The user-scope copy was deleted but kept its state row. `removeOne` had already computed the correct per-entry state (it retains failed agents via `reduceEntryAgents` and drops clean ones via `dropScopedEntryAndUpdateRequiredBy`) — the caller threw that away and re-decided cleanliness per *name group*, so one aborted location retained its siblings' rows too.

`removeOne` now returns an `EntryOutcome[]` (`{ entry, fullyRemoved }`) and the tap path drops exactly the entries that came off cleanly. Same probe after the fix:

```
AFTER:  [ "alpha/project//tmp/crew-proj-…" ]
```

**Beyond the report — the prune sweep on this head is unbounded.** Retaining aborted entries is precisely what breaks `pruneOrphans`' termination argument: `findOrphan` is memoryless, a safety-aborted orphan keeps `explicit:false` + empty `required_by`, and `while (orphan)` is handed the same entry forever. I confirmed `findOrphan` returns the identical entry from unchanged state. The equivalent fix landed on #110's branch (`0e64483`) but that branch is **not an ancestor of this one**, so the hazard is live here — and Major 1's fix increases how often entries are retained. `findOrphan` now takes an `attempted` set and `pruneOrphans` records each candidate **before** removing it, so every entry is visited at most once per sweep regardless of outcome. This is the invariant that keeps the loop terminating; please don't relax it.

**2. Install concurrency guard accepts a replacement tap with the same name — `src/install/config-merge.ts:43`**

Reproduced. `assertTapsPresent` did `names.has(skill.tap.name)` — name-only. Driving it with a skill resolved from `file:///repo/A` against a lock-time config where `mytap` points at `file:///repo/B`:

```
MAJOR2 threw? NO — guard passed (bug)
```

A name is a mutable label; `tap remove` + `tap add` rebinds it in the pre-lock resolution window, and state records skills by tap *name*, so A's bytes get attributed to B. The guard now compares full source identity (`kind`, `url`, `subpath`, `path`) and distinguishes "was removed" from "was replaced" in the message. Covered for plain URL rebinds, subpath-only rebinds (same repo, different directory — still different bytes), and git→path replacement.

**3. Update uses a pre-lock config snapshot — `src/commands/update/index.ts:51`**

Reproduced. `readConfig` was at line 51, outside `withStateLock` at line 61; `entry.ts:108` then resolved the tap from that snapshot. With the tap present in the snapshot, `acquireTap` → `ensureClone` **re-creates the clone that `tap remove` just deleted**:

```
MAJOR3 fresh-config outcome:              [ "tap_missing" ]   <-- correct
MAJOR3 clone recreated from stale config? true                <-- the bug
```

Config is now read inside the locked section, so a forced removal landing in the gap yields the §10.1 soft `tap_missing` instead of resurrecting the tap.

On reproducing these two honestly: both are check-then-act over a snapshot taken outside the lock, so the interleaving is constructible by driving the guard with the config state a concurrent process would have left — no lock prevents it, because the read happens before the lock is taken. That's the bug in both cases.

---

### Minors

1. **Exhaustiveness before `formatRowParts`' failed fallback** — added as a **type-only** `o satisfies Extract<Outcome, { kind: "failed" }>`, matching the settled pattern already in its sibling `symbolFor`. A runtime `default:`/`assertNever` would cost line coverage against the repo's 100% gate.
2. **Bulk state removal O(N×K)** — `dropEntriesAndUpdateRequiredBy` now pre-indexes removed names into a `Map<locationKey, Set<name>>`, so each survivor costs one lookup: N+K, not N×K.
3. **Forced dry-run coverage** — added three tests covering the untracked, inconsistent-marker, and shared-marker (detach) paths. Each has its own `if (!input.dryRun)` guard in `src/agents/uninstall.ts`, and the un-forced paths throw before reaching the guard, so each genuinely needed its own test. Each asserts the preview changes nothing *and* that the real run does the work described.
4. **Same-named-other-tap test** — confirmed: it added `othertap` but never installed from it (`readState` length 1), so there was only ever one `alpha` and the test proved nothing. It now installs both taps' copies at distinct locations (user + project) and asserts the survivor is exactly `alpha/project/othertap`.
5. **Direct tests for auto-tap merge and tap validation** — there were none at all. Added `tests/misc/config-merge.test.ts`: 8 tests covering `mergeAutoTaps` (keeps only resolver-added taps, does not resurrect a concurrently removed tap, identity-returns when nothing was added) and `assertTapsPresent` (same-source pass, disappeared, replaced, subpath-only rebind, git→path rebind).
6. **`tap_missing` taxonomy** — it *is* a soft outcome, already in PRD §13 at exit 0 with criterion C-UPD-12b, but it was missing from `CrewErrorName` while its sibling `source_gone` was present. Neither is ever thrown, so the asymmetry was the anomaly. I made it symmetric (taxonomy entry, exit 0, remedy hint) and PRD §13 now states outright that `source_gone` and `tap_missing` are soft outcomes an implementation need not model as thrown errors — so the table is a `--json` contract rather than an ambiguous error list.

---

### Verification

- `bun run lint` — 0 errors, 2 pre-existing biome-config infos.
- `site` lint — 1 pre-existing info.
- Full suite: **1039 pass, 1 fail, 100% line and 100% function coverage.**
- Conflict-marker sweep — clean.

The one failure is `C-INST-08c --recursive installs nested skills…`. I stashed my work and confirmed it **fails identically on the untouched PR head (8d9de60)**; this PR touches none of the recursive-discovery code, and the test dates to #58. Pre-existing, not introduced here.

Coverage note: folding the `required_by` scrub into a `for` loop (per the repo's documented `.filter(cb)` footgun) left one line uncovered — no test had a survivor at a removed location retaining a *surviving* edge. Rather than contort the code I added `tests/misc/uninstall-state.test.ts` covering that scrub directly, which restored 100%.

### Not done deliberately

- **No rebase.** `origin/fix/tap-dry-run` is still an ancestor of this head, so the stack is intact and rebasing would only inflate the diff.
- **The missing CI run at this head** is the known stacked-PR gap that #134 addresses; nothing to fix here.

stevekrenzel:
## Correction: `C-INST-08c` was NOT pre-existing — it is a regression this stack introduced, now fixed in 2b8a58c

**My earlier claim in [this comment](https://github.com/with-logic/crew/pull/130#issuecomment-5655968743) was wrong.** I wrote that the `C-INST-08c` failure was pre-existing and not introduced here. That framing is incorrect and I want to correct it plainly rather than in a footnote.

What I actually did was compare against `8d9de60` — the prior head of *this same stack*, which already contained the defect. That only established "my round-three changes didn't cause it", which is true but is not what "pre-existing" means. I should have compared against `main`, and I didn't. Thanks to @aa1c5dbcc5ddd694e for catching it.

### The corrected record

| ref | result |
|---|---|
| `origin/main` (6428d43) | **1 pass, 0 fail** |
| `4fdbf22` (stack, pre-tip) | **1 pass, 0 fail** |
| `8d9de60` (stack tip, prior PR head) | **0 pass, 1 fail** |
| `7068995` (my round-three head) | **0 pass, 1 fail** |
| `2b8a58c` (this commit) | **1 pass, 0 fail** |

`main` is green. The stack introduced the failure at `8d9de60`, and it survived my round-three work because I hardened `assertTapsPresent` and left `mergeAutoTaps` alone. I re-ran every row above myself.

### The actual defect

`mergeAutoTaps` decided what to salvage from the resolver's pre-lock snapshot by asking whether a tap's **name** was new:

```ts
for (const t of extended.taps) {
  if (known.has(t.name)) continue;   // drops MODIFICATIONS
  added.push(t);
}
```

The resolver does not only *append* taps. `crew install --recursive <dir>` against an already-registered tap calls `maybeUpgradeDiscovery`, which rewrites that tap's `discovery` **in place with the name unchanged**. A name-only test files that under "already known" and discards it. The file's own docstring — "keep only the taps the resolver actually ADDED" — encoded the wrong assumption.

This is a real user-facing bug, not a test artifact: **`crew install --recursive` against an existing tap silently failed to enable recursive discovery.**

```
expect(tap.discovery).toBe("recursive")
Expected: "recursive"
Received: undefined
```

### Reasoning it through (rather than pattern-matching the suggested fix)

I checked the resolver's mutation sites before designing anything. There are exactly three, all in `tap-attribution.ts`: two appends and one `discovery` upgrade. **The resolver never removes a tap and never changes any other field.** That makes the case analysis complete and closes the ambiguity:

| in `before` | in `fresh` | meaning | action |
|---|---|---|---|
| no | no | resolver added it | keep |
| no | yes | another process added the same name concurrently | keep fresh's row, don't clobber |
| yes | no | **concurrently removed** | drop — stays removed |
| yes | yes | existed throughout | apply resolver modification if any |

I agree with the conclusion that was suggested to me, and here is why it's sound rather than just plausible: "present in `extended`, absent from `fresh`" is genuinely ambiguous **on its own**, but `before` disambiguates it — since the resolver never deletes, a name in `before` that is missing from `fresh` can only have been removed by a concurrent writer. So the two goals do **not** actually conflict, and I did not have to trade one bug for the other. The concurrency guarantee from Major 2 is fully intact.

One refinement I made beyond the sketch: I apply the resolver's **delta** (only `discovery`, the one field it changes) rather than overwriting fresh's row with `extended`'s version. Overwriting wholesale would clobber a concurrent edit to an unrelated field on that same tap. There's a test for exactly that.

### Tests

Three new cases in `tests/misc/config-merge.test.ts`, each of which **fails at `7068995` and passes at `2b8a58c`**:

- a lone resolver modification survives the merge;
- a modification and a concurrent removal handled correctly **in the same merge** (proving neither goal was traded for the other);
- a concurrent edit to an untouched field is not clobbered.

**Load-bearing check:** I reverted *only* `src/install/config-merge.ts` and re-ran. Exactly four tests failed — those three plus `C-INST-08c` — and nothing else. Restored, all pass.

### Verification at 2b8a58c

- `bun run lint` — 0 errors, 2 pre-existing biome-config infos.
- `site` lint — 1 pre-existing info.
- Full suite: **1043 pass, 0 fail, 100% line and 100% function coverage.**
- Conflict-marker sweep — clean.

The suite is now **fully green**, where my previous head had 1 failure. Stack is unchanged: `origin/fix/tap-dry-run` is still an ancestor, no rebase.

### Lesson worth recording

"Does this test fail before my change?" is the wrong question on a stacked PR — it only rules out the topmost commit. The right baseline is `main` (or the stack's merge-base), because an ancestor in your own stack is exactly where a regression you're accountable for will already be hiding.

stevekrenzel:
Rebased onto `main` (`15210a0`) after #116 and #119 merged. New head `a7c95b8`; all five commits re-signed. One conflict, in `src/cli/args.ts`, resolved in-rebase at "scope tap removal to its own entries and keep aborted ones":

**What collided.** This PR gave `untap` its `--uninstall` flag through a local `BOOLEAN_ALIAS_SUB` table spread into `parseArgs`. #116 meanwhile made the flag tables alias-aware via `aliasFlagKey` + `lookup`: a bare alias (`rm`) shares its canonical command's tables, and a *prefixed* alias (`taps` → `tap list`) inherits no subcommand flags at all, so `crew taps --recursive` stays a `usage_error` (tested in `tests/e2e/aliases.test.ts`).

**How it was reconciled.** Kept #116's mechanism as the base and layered this PR's grant on top, only for the flags a prefixed alias explicitly lists:

- The table now lives next to `COMMAND_ALIASES` in `src/cli/aliases.ts` as `ALIAS_FLAGS = { untap: ["uninstall"] }`, exposed as `aliasBooleanFlags(command)` (prototype-safe via `lookup`).
- `parseArgs` spreads `...aliasBooleanFlags(command)` after the `aliasFlagKey`-derived tables. Nothing else about #116's lookup changed.

Net effect: `crew untap --uninstall` parses (C-TAP-16d), `crew untap --recursive` and `crew taps --recursive` still reject. `src/cli/args.ts` moved from 203 lines on `main` to 204; the split that brings it back under the cap is #114's.

Verification on the new head: `bun run typecheck`, `bun run lint`, and `site` lint clean; targeted suites (every `tests/e2e` file naming a tap, plus `aliases.test.ts` — 32 files) 354 pass / 0 fail.

stevekrenzel:
Rebased onto `main` after the cap-compliance refactors (#135–#143) merged. No behavior change — every conflict was a file that had been split out from under the edit.

Re-placed:

- **`src/cli/args.ts` → `src/cli/args/{index,tables}.ts`.** The `tap: ["recursive", "uninstall"]` grant now lives in `tables.ts`'s `BOOLEAN_SUB`. The `...aliasBooleanFlags(command)` spread went back into `parseArgs`, which now builds its boolean list from `subFlags(BOOLEAN_SUB, command)` rather than the old inline `aliasFlagKey`/`lookup` pair; `ALIAS_FLAGS`/`aliasBooleanFlags` in `src/cli/aliases.ts` were untouched by the refactor and applied as-is. Verified both tested behaviors directly: `crew taps --recursive` still exits 4 with `Unknown argument: recursive`, and `crew untap --uninstall <tap>` parses and reaches tap resolution.
- **`tests/e2e/search.test.ts` → `tests/e2e/search/installed.test.ts`** — the `--force` change to the tap-gone-but-skill-installed criterion.
- **`site/components/sections/Commands.data.tsx` → `Commands.data/discovery.tsx`** — both the `tap-remove` and `untap` row updates, which is where the tap rows now live.

One addition beyond a straight re-placement: `tests/e2e/tap-remove-guard/uninstall.test.ts` came out of the rebase at 208 lines, over the newly enforced cap. Split the four abort/partial-abort cases into a sibling `aborts.test.ts` and folded that into the commit that grew the file, so every commit in the range is cap-clean.

`bun run check`: typecheck clean, `Checked 398 files` lint clean, file-size gate clean, **1136 pass / 0 fail**, coverage **100.00 / 100.00**. `cd site && bun run lint` clean (the one "info" is the pre-existing biome `$schema` 2.4.16-vs-2.5.0 notice already on `main`).

PR #131, opened 2026-09-11
[LOG-5777] test: stop the suite contacting the network, and add a tripwire to keep it that way

stevekrenzel:
## What

The test suite was contacting github.com on every run. This stops it, and adds a tripwire so it can't come back.

CLAUDE.md's testing philosophy is real filesystem, real local `git`, no network. That's easy to violate by accident, because `crew install` materializes configured taps on demand — a fixture carrying a remote URL turns a local-looking assertion into a live `git clone`.

## The audit

I wrapped the git seam with a temporary instrumented runner and logged every invocation that would reach a remote. Two genuine sources, 14 network calls per full run:

| Source | Calls | Detail |
|---|---|---|
| `tests/e2e/install-known-tap.test.ts` | 4 | `KNOWN_TAPS` and `configuredKnownSource()` both pointed at `https://github.com/example/supabase-skills.git`, a repo that doesn't exist. The test writes that tap into config, then runs a real `crew install`. |
| `makeCrewHome()` | 10 | Left the default `core` tap at its real GitHub URL. Any test resolving a bare name, or running `crew update` (which refreshes every configured tap), cloned it. |

Each failed lookup costs roughly a second against a reachable GitHub, and would hang or fail outright on an offline or throttled runner.

Everything else matching `github.com` / `https://` in `tests/**` is a string assertion on error text, suggested commands, or `knownTapSource` formatting — no value flows into a clone. Those keep working unchanged. `tests/self-update/**` asserts on URL strings and stubs its fetch; it makes no requests.

**Likely explains a real flake.** A worker on another PR saw `C-TAP-24 configured known taps are not suggested by name or source` fail once in a full run, while passing 5/5 in isolation and on clean `main`. That's one of the four tests above, and its behavior depended on a network round trip.

## The fixes

- **Known-tap fixtures** use a nonexistent `file://` path. One URL spelling throughout: `knownTapSource` only folds a `.git` suffix away for GitHub HTTPS URLs, and `knownTapIsConfigured` dedupes by comparing rendered source strings, so a base-vs-`.git` mismatch would have made the fixture exercise suffix handling instead of the dedupe it's there to test.
- **`makeCrewHome()`** seeds a `config.yaml` whose `core` tap points at an unreachable local path. `core` stays configured, so tests asserting on it still pass. The single test about a *missing* config file now uses a bare temp dir, which is what it actually meant.
- **`tests/helpers/no-network.ts`** wraps the `setGitRunner` seam and rejects `clone`/`fetch`/`ls-remote`/`push`/`pull` whose remote isn't a `file://` URL or local path. Real local git runs normally, so this is an assertion about remotes rather than a git mock — the "don't mock git" rule stays intact. `allowRemoteGit()` is the deliberate escape hatch; **nothing uses it**.
- **A global `afterEach` in the preload** fails any test that recorded a violation. This matters: crew soft-fails an unreachable tap by design (PRD §16.6), so `attribute-bare-name.ts` and the search path swallow the tripwire's throw. Without this, a violation would pass silently — which is exactly what happened on my first attempt.

### One `src/` change

`resetGitRunner()` hardcoded the raw default runner, so a test that installed its own stub and reset it in `afterEach` discarded the suite-wide tripwire partway through the run. It now restores a base that the preload sets via a new `setBaseGitRunner`. Test-seam only, no product behavior change, so no PRD update.

## How tested

- `bun run check` exits 0: 992 pass, 100% line and function coverage, tsc and Biome clean.
- **Tripwire proven, not assumed.** I reintroduced a remote URL into the known-tap fixture and confirmed the run fails with `This test contacted the network: git clone https://github.com/example/supabase-skills`, then reverted. My first version passed that experiment wrongly, which is how I found the swallowed-error gap.
- `tests/helpers/no-network.test.ts` covers the guard directly: remote rejected and recorded, ssh-style rejected, local `file://` passes through to real git, non-remote subcommands unguarded, `allowRemoteGit` lifts and restores. Every case stays offline — the opt-out test uses an unknown URL scheme that git rejects locally before opening a connection.

## Note on timing

I can't give a trustworthy before/after wall-clock number: several other agents are running this suite in parallel on this machine right now, and full-run times swung between 117s and 244s regardless of this change. The honest measurement is the per-call one — ~1s per failed remote lookup, 14 of them removed, on a connection where GitHub answers quickly. The real win is CI that no longer depends on a third party being reachable.


stevekrenzel:
Rebased onto `main`. One test from #112 (`C-AUTO-11 … macOS enable … writes nothing`) asserted `config.yaml` is absent after a dry run; this PR's `makeCrewHome` now seeds `config.yaml` with the offline `core` tap, so the assertion is changed to 'config.yaml bytes unchanged' — a stronger claim of the same intent. No other changes.

stevekrenzel:
Rebased onto `main` after the cap-compliance refactors (#135–#143) merged. The new file-size gate (#137) caught this PR's one growth: the offline `UNREACHABLE_TAP_URL` fixture pushed `tests/e2e/install-known-tap.test.ts` from 199 to 212 lines. Split it into `tests/e2e/install-known-tap/` (`helpers.ts` for the fixtures, `suggestions.test.ts`, `dedupe-and-refs.test.ts`); tests moved verbatim, none changed. Folded into the single commit; full suite 1099 pass at 100% coverage.

PR #132, opened 2026-09-11
[LOG-5778] perf(taps): share one clone per repository instead of one per tap

stevekrenzel:
Stacked on #130 → #119. Review after those land; the diff against `feat/tap-remove-guard-v2` is the change under review here.

## What

Clone directories were keyed by **tap name**. A tap row is identified by its `(url, subpath)` pair, so every subpath a user touched cloned the whole repository again. Installing two skills from one repo by subpath produced two full clones of the same history; adding a third tap over the repo root made three.

Clones now live at `repos/<host>-<owner>-<repo>-<hash8>/`, keyed by the repository's **canonical URL**. Every tap row over that repository reads from the same working tree.

Tap identity is deliberately untouched: rows keep their own name, url, subpath, `registered` flag, and discovery mode, and skill attribution and `tracks_tap` semantics are unchanged. Only the bytes are shared.

## Measured saving

A repo with two skill subdirectories, two subpath taps:

| | clone dirs | on disk |
|---|---|---|
| before | 2 | 10016 KB |
| after | 1 | 5008 KB |

An exact halving for two taps, scaling as 1/N for N taps over one repository. The real-world case from the brief (`github.com/anthropics/skills` at ~16 MB with three tap rows) goes from ~48 MB to ~16 MB.

## Consequences handled

- **Shared deletion guard.** `crew tap remove` deletes the clone only when no surviving tap row points at the same repository. The same guard covers auto-tap GC on uninstall — without it, uninstalling a skill installed from a repo would delete a *registered* tap's bytes. There's a test for exactly that.
- **Fetch dedupe.** `refreshTaps` fetches a repository at most once per run. Additional tap rows over it report that fetch's outcome, including its failure, rather than hitting the network again. Every tap still gets its own row in the output.
- **Promotion simplifies.** Auto-tap promotion no longer moves a clone directory, since the name isn't part of the clone's location. `targetKind` became an unused parameter and was dropped.
- **Migration.** Homes created before this change are migrated on first use: the old `taps/<tap-name>/` directory is moved into place (a rename, never a re-clone), or discarded when the shared clone already exists. Idempotent.

## URL canonicalization

Two URLs address the same repository when they match after dropping a trailing `.git`, dropping trailing slashes, and lowercasing scheme and host. The path keeps its case, since some hosts are case-sensitive.

`git@host:owner/repo` and `https://host/owner/repo` resolve to **different** clones. They're different remotes with different auth, and `findOrCreateGitTap` already matches taps on exact URL string, so this preserves existing behavior rather than changing it.

**Note for the reviewer:** open PR #128 adds `src/install/source-identity.ts` with its own canonical-URL logic for a different purpose (name-conflict detection), and is not in this base. Whichever lands second should collapse the two onto one canonicalizer.

## PRD

- §6 — `repos/` in the layout, the one-clone-per-repository rule with its three MUSTs, and the migration requirement.
- §16.3 — tap remove wording for shared clones.
- §16.5 — auto-tap GC wording.
- §18.3 — C-TAP-01 and C-TAP-03 corrected (both described the old layout); new C-TAP-17, C-TAP-17b, C-TAP-17c, C-TAP-18.

## Testing

`bun run check` green: **1024 pass, 0 fail, 100% line and function coverage**, typecheck and Biome clean.

New `tests/e2e/shared-clones.test.ts` (11 tests): two subpath taps share one clone; both installs work off it; `.git` spellings collapse; removing one tap keeps the clone, removing the last deletes it; auto-tap GC both keeps a shared clone and deletes an unshared one; update fetches once; a failed fetch is reported for every tap on that repo; and both migration paths.

Five existing tests asserted the old `taps/<name>/` layout and were updated to resolve the clone through config via a new `cloneDirForTap` helper. One of them (`update-edges`) was deleting a path that no longer held the clone, so it had stopped testing the unreachable-source case at all.

Verified by hand end to end: two subpath installs produce one clone directory and no `taps/` entry; a home moved back to the legacy layout migrates on the next command and keeps working.

## Site

`Taps.tsx` printed `✓ cloned acme → ~/.crew/taps/acme (42 skills)`, a path that no longer exists. It now shows the tap name without a path, which also matches what the command actually prints.


stevekrenzel:
Rebased onto the revised parent (`044c148`). Branch is now `115c58a`; PR is MERGEABLE and the base is unchanged.

## What conflicted, and how it resolved

Four conflicts. The stale copy of the parent's old commit was skipped automatically, which is what produced most of them.

|---|---|---|
| `src/commands/tap/add.ts` | Parent's discriminated `TapAddPlan` removed the `!` assertion; this branch dropped the `target.kind` argument from `promoteExistingTap` (a rename no longer moves a directory under shared clones) | Kept both: assertion-free access, this branch's argument list |
| `src/commands/tap/remove/index.ts` | Parent imports `tapPath`; this branch imports `cloneStillReferenced` / `tapClonePath` | Took the shared-clone imports, dropped `StateFile` which the parent's rework no longer needs |
| `tests/e2e/tap-dry-run.test.ts` | Parent split it into a directory; this branch had edited the monolith | Deleted the monolith, ported its adaptations onto the split files |
| `tests/e2e/tap-remove-guard.test.ts` | Same | Same |

The substantive merge is in `dropTap`. The parent's version deleted `tapPath(tap.name, home)` unconditionally; this branch makes deletion conditional on no surviving tap row referencing the same repository. The parent's structure — `planRemove` / `runPlan`, one lock-held operation for a real run, lock-free preview, tap-scoped `byName` grouping — is preserved, with reference-counted deletion inside its `dropTap`.

Porting the test adaptations was not mechanical. `tapPath` was renamed to `legacyTapPath` on this branch and no longer addresses a clone, so the parent's split files referenced a function that doesn't exist here. Three cases needed judgment rather than substitution:

- **`--uninstall` deleting the clone**: `cloneDirForTap` resolves through config, so after removal the tap row is gone and the lookup returns null. The path is now captured before removal.
- **`tap add --dry-run`**: a preview creates no tap row, so there's nothing to look up. Asserts no clone directories exist at all, which is the stronger claim.
- **Promotion preview**: asserting no directory appears under the promoted name is meaningless when promotion never moves a directory. Asserts the repository's clone count is unchanged instead.

## No criteria renumbering needed

The parent's rows are C-TAP-16c..f; this branch's are C-TAP-17/17b/17c/18. No overlap.

## Confirmation: reference counting and the guard both hold

Neither side covered the interaction the two changes create, so I added it: `tap remove --uninstall` on a tap whose repository another tap still references must keep the bytes. It drives the parent's reworked guard path end to end.

A clean rebase and a green suite prove nothing on their own, so I checked the assertion is load-bearing — removing the `cloneStillReferenced` guard from `dropTap` fails three tests, including the new one:

```
(fail) C-TAP-17b removing one tap keeps the clone the other still needs
(fail) C-TAP-17b removing the last tap on a repo deletes the clone
(fail) C-TAP-17b `tap remove --uninstall` keeps a clone another tap shares
```

Restored afterwards. Auto-tap GC was already covered from both directions and still is.

Targeted runs: shared-clone suites 12 pass, parent's guard suites 19 pass.

## One thing beyond conflict resolution

`shared-clones.test.ts` reached 263 lines, over the hard cap, and my added case pushed it further. Split into `tests/e2e/shared-clones/` — `sharing.test.ts` (177), `migration.test.ts` (58), `helpers.ts` (39). All 12 cases preserved.

## For whoever merges

**#128** (`fix/same-repo-not-name-conflict`, not in this base) adds `src/install/source-identity.ts`, a second canonical-URL implementation, alongside this branch's `src/core/repo-path.ts`. They must converge on one. #128 also adds `rmrfInside` containment for deletions built from persisted names — clone deletion here should route through it, since a tap name reaching `rmrf` is exactly the case that helper exists for.

`bun run check`: 1031 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean.

stevekrenzel:
Rebased onto the revised parent `feat/tap-remove-guard-v2` (`8d9de60`). Two of this branch's commits were replays of the parent's own work and were skipped as duplicates. Two real conflicts:

- `tests/e2e/tap-remove-guard/uninstall.test.ts` — the parent's assertion used `tapPath("othertap", home)`, which no longer addresses a clone under the shared-clone layout (renamed `legacyTapPath`). Ported it onto `cloneDirForTap`, so the parent's intent (the other tap keeps its clone) is asserted against the repo-keyed path.
- `src/commands/tap/add.ts` — the parent added a type-only `satisfies` exhaustiveness check on the `added` outcome; this branch changed `cloneNewTap` to take only the URL, since clones are keyed by repository rather than tap name. Kept both.

**Reference-counted deletion still holds, and I proved it rather than inferring it.** Both call sites survive — `dropTap` in `src/commands/tap/remove/index.ts:84` and auto-tap GC in `src/commands/uninstall/index.ts:167` — and each `rmrf` of a clone path sits immediately behind its `cloneStillReferenced` guard, so the parent's reworked removal path does not delete directly. Disabling the guard at both sites fails exactly the four tests that encode the behaviour:

```
(fail) C-TAP-17b removing one tap keeps the clone the other still needs
(fail) C-TAP-17b removing the last tap on a repo deletes the clone
(fail) C-TAP-17b auto-tap GC keeps a clone a registered tap still uses
(fail) C-TAP-17b `tap remove --uninstall` keeps a clone another tap shares
```

Restored afterwards; `tests/e2e/shared-clones/` is back to 12 pass / 0 fail.

Parent is an ancestor. Typecheck and Biome clean. `tap-remove-guard/` showed 1 failure on the first run and 21 pass / 0 fail on an immediate re-run with no code change — the known non-determinism on branches without #131.

stevekrenzel:
Rebased onto the new #130 head (`a7c95b8`, itself now on `main`). New head `6d392cd`; both commits re-signed. Two small conflicts with what merged to `main`, resolved in-rebase: the PRD's `~/.crew/` ownership paragraph keeps this PR's `repos/` rewrite and the `crew cache clean --dry-run` sentence #112 added, and `src/sources/acquire/index.ts` keeps both new imports (`assertNoSymlinkEscape` from #111, `migrateTapClone` from here). Typecheck, lint, and site lint clean; targeted tap/alias/sources suites (34 files) 366 pass / 0 fail.

stevekrenzel:
Rebased onto the rebased #130 after the cap-compliance refactors (#135–#143) merged. No behavior change — the clone-sharing edits were re-placed into the files that now hold the code they touched.

Source side:

- **`src/git/repo.ts` → `src/git/repo/{index,refs.ts}`.** `ensureRepo` (used by `src/commands/tap/refresh.ts`) and `ensureClone` (used by `src/sources/acquire/index.ts`) resolve to `repo/index.ts`; `resolveRef` to `repo/refs.ts`. Import specifiers merged without duplicates — `refresh.ts` keeps both `ensureRepo` from the new path and `migrateTapClone`.

Test side — each assertion went into the topic file that now holds its neighbours:

- `tests/e2e/tap.test.ts` → `tap/add.test.ts` (C-TAP-01 clone path, the `typo-tap` no-leftover-clone case) and `tap/manage.test.ts` (C-TAP-03 removal).
- `tests/e2e/tap-unify.test.ts` → `tap-unify/tap-name-install.test.ts` (C-TAP-21 path tap clones nothing).
- `tests/e2e/tap-update.test.ts` → `tap-update/fetch-policy.test.ts` (3) and `tap-update/update.test.ts` (8) — all 11 `tapPath` → `cloneDirForTap` call sites.
- `tests/e2e/tap-install.test.ts` → `tap-install/update.test.ts` (C-UPD-19 shared-clone working tree).
- `tests/e2e/search.test.ts` → `search/configured.test.ts` (the known-tap dedupe case that re-points the tap URL and moves the clone so the search never has to fetch).
- `tests/misc/gaps.test.ts` → `tests/misc/closeout/util-json-paths.test.ts` (the `tapPath` → `legacyTapPath` rename).

One consequential tweak: `tests/e2e/tap-dry-run/add.test.ts` lost its last two `existsSync` uses to this PR's own `cloneDirs` assertions, so the now-unused import was dropped (`noUnusedLocals`).

`bun run check`: typecheck clean, `Checked 403 files` lint clean, file-size gate clean, **1148 pass / 0 fail**, coverage **100.00 / 100.00**. `cd site && bun run lint` clean (the one "info" is the pre-existing biome `$schema` 2.4.16-vs-2.5.0 notice already on `main`).

PR #134, opened 2026-09-11
[LOG-5790] ci: run tests on pull requests that target a branch other than main

stevekrenzel:
## What

Removes the `branches: [main]` filter under `pull_request` in `ci.yml` and `site-ci.yml`, so both workflows run for every pull request regardless of base branch.

## Why

`pull_request.branches` filters on the PR's **base** branch, not the head. With `[main]`, any stacked PR — a feature branch based on another feature branch — matched neither workflow and ran **no tests and no lint at all**.

That was easy to miss because those PRs still showed a green check rollup. The passing checks were Vercel, the Linear ticket bot, and the Claude review bot. The suite never ran.

Verified against the live PRs before changing anything:

| PR | Base | `check` | `site-check` |
|---|---|---|---|
| #107 | `main` | present, pass | present, pass |
| #118 | `fix/uninstall-scope` | absent | absent |
| #124 | `fix/list-scope-filter` | absent | absent |
| #132 | `feat/tap-remove-guard-v2` | absent | absent |

**Eleven open PRs are affected:** #118, #120, #121, #122, #123, #124, #125, #126, #127, #130, #132.

**Their current green checkmarks did not include tests or lint.** Reviewers should not read them as evidence the suite passed. For each of those PRs, either wait for a CI run on `main` after its parent merges and it retargets, or push any commit to the branch once this lands to trigger a real run.

## Decisions

- **`push` keeps `branches: [main]`.** Without that, every push to a feature branch would run CI alongside the PR-triggered run, doubling the work for no added signal.
- **`dependabot-bun-lock.yml` keeps its `[main]` filter.** It is gated on `github.actor == 'dependabot[bot]'`, and dependabot only opens PRs against `main`, so widening it would change nothing.
- **No unwanted new runs.** The only PRs that newly trigger CI are exactly the stacked ones that should have been running it all along.
- **Concurrency is unaffected.** Both workflows group on `${{ github.workflow }}-${{ github.ref }}`. For `pull_request` events `github.ref` is `refs/pull/<N>/merge`, which is unique per PR and independent of the base branch, so stacked PRs each get their own group and still cancel their own superseded runs. No cross-PR cancellation.

## Verification

- Confirmed both files parse and that the bare `pull_request:` key survives as a present key with a null value — GitHub's "all activity types, all branches" form — rather than dropping the trigger.
- `bun run check`: 985 pass, 0 fail, 100% line and function coverage, typecheck and Biome clean.
- No existing PR was retargeted or modified.


stevekrenzel:
Rebased onto `main` after the cap-compliance refactors (#135–#143) merged; no conflicts, no changes.

